PR #45, opened 2020-07-02
[xlscc] Quick changes to enable combinational IO blocks

spurserh:
These changes support an internal project. 

The gist is that we want to have HLS-generated combinational mux blocks, with no flops. This requires generating logic to select both the forward signals of data and valid, as well as the backward ready signal. 

There are also improvements to the driver, and make some initial steps toward supporting IO channels.


spurserh:
Abandoning this as it will be made obsolete by the clang rewrite.

PR #83, opened 2020-08-10
Add data generation/plotting for study of SMT solver speeds

julianviera99:
Add in scripts for generating add, mul, and shift proofs, scripts to generate and store data in csv files, and a script to plot data from these csv files.


RobSpringer:
Obviated by a more recent commit.

PR #121, opened 2020-08-28
Fixed bug in BDD simplification pass. 

### LINKED ISSUE #115: Fuzz bug: incorrect "one_hot msb" transform
Found by fuzz test while tweaking something unrelated, and I minimized it a bit by hand since the current minimizer only works at the IR level. First time we've seen the fuzzer catch something in a while! As a mitigation for now I'll disable the transform until we properly understand the bug and can re-enable the pass.

It also would be nice if we could separate out the transform from the BDD analysis somehow, right now it's monolithic so we have to internally disable that part within the mega BDD transform pass.

```
// options: {"input_is_dslx": true, "convert_to_ir": true, "optimize_ir": true, "codegen": false, "simulate": false, "simulator": null}
// args: bits[8]:0x5; bits[12]:0xe9c; bits[35]:0x5_5540_0000

fn main(p3: u8, p6: u12, p8: s35) -> u8 {
  let x13: u12 = clz(p6);
  let x14: u12 = clz(x13);
  let x18: u8 = ((p8 as u8)) | (p3);
  let x24: u1 = xor_reduce(x14);
  let x30: u8 = one_hot_sel(x24, [x18]);
  x30
}
```

Assigning to Mark since he's most familiar with the pass.

jbaileyhandle:
The bug occured because we reused BDD analysis generated before OneHot MSB simplification after OneHot MSB simplification. Although the OneHot transform is only performed if the output is unaffected, this transform may change intermediate values. The previous BDD analysis on such intermediate values may therefore no longer be accurate. Fixed by performing a fresh analysis after the OneHot Transfrom.

Fixes #115

### DIFF
--- a/xls/passes/bdd_simplification_pass.cc
+++ b/xls/passes/bdd_simplification_pass.cc
@@ -323,10 +323,6 @@ xabsl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
 }
 
 xabsl::StatusOr<bool> SimplifyOneHotMsb(Function* f) {
-  // TODO(meheff): 2020-08-24 Disabled for debugging of crasher
-  // xls/dslx/fuzzer/crashers/2020-08-24-min.x
-  return false;
-#if 0
   bool changed = false;
   XLS_ASSIGN_OR_RETURN(
       std::unique_ptr<PostDominatorAnalysis> post_dominator_analysis,
@@ -435,7 +431,6 @@ xabsl::StatusOr<bool> SimplifyOneHotMsb(Function* f) {
   }
 
   return changed;
-#endif
 }
 
 }  // namespace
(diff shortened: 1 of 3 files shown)


cdleary:
@jbaileyhandle looks like you need to sign the CLA? LMK when you have and I'll have it re-scan.

jbaileyhandle:
@googlebot I signed it!

jbaileyhandle:
@cdleary cla is signed.

Different question:  I tried to fuzz this, and the fuzzer's memory footprint would monotonically increase until my computer crashed.  I tried limiting the number of workers to 1, but this only slowed down the footrpirnt growth without preventing a crash.  I never observed this internally (granted, I had access to a lot more RAM...).  Have you seen this before?  Could a memory leak have been introduced in the past couple weeks?

cdleary:
@jbaileyhandle Very interesting, the fuzzer OOM could be also why our nightly worker blows up. I'll file an issue and look into that. Maybe something different about google3 multiprocessing vs OSS...

We currently don't have linting or clang-formatting instructions in OSS. I think clang-format with google style should be right on the C++ code. If you happen to come up with instructions for setting that up / running it I'd happily accept them on CONTRIBUTING.md! Also could we add to the GitHub action for Continuous Integration...

PR #147, opened 2020-09-25
Specialize select arms

### LINKED ISSUE #145: Specialize select arms
Some select operations can be optimized by specializing the arms of the select based on the selector value which chooses the arm. For example:

selector = ...
lhs = ...
rhs = ...
out = select(selector, cases=[lhs, rhs])

Here lhs is selected when selector is false. The expression computing lhs can assume the selector is false, and the expression computing the rhs can assume the selector is true. This assumption is safe because if it turns out to be wrong the "miscomputed" lhs or rhs is dead.

A more concrete example:

  a ? (!a ? x : y) : z

in the subexpression (!a ? x : y) we can assume a is true and it simplifies to y, so the whole expression can be reduced to:

  a ? y : z

Mechanically the transformation to perform might be something like the following. IR looks like:

S = ...
A = ...
B = ...
C = select(S, cases=[A, B])

Necessary condition to specialize arm A which is selected when S is false (there are symmetric conditions for B):

  A's only use is C

Then for each use of S which is post-dominated by A, replace the use of S with literal 0. If there are no such uses, then there is nothing to transform.

XLS has a post-dominator analysis which should make this easy.

Later versions of this can be fancier. Like if the selector is some comparison like (foo < 7) then you could assume foo < 7 (or foo >= 7) in an arm of the select. This woudl dove-tail with range analysis. Also we could handle more than just two-way select.

felixzhuologist:
Closes #145 

Adds a step to `SimplifyNode` in the select simplification pass that will try to specialize the arms of the select. The current implementation does not support simplifying cases where the selector is not post-dominated by the arm, and works by doing an exact match on the selector within each arm.

Test by running: `bazel test -c opt --test_output=errors //xls/passes/...`

### DIFF
--- a/xls/passes/select_simplification_pass.cc
+++ b/xls/passes/select_simplification_pass.cc
@@ -31,6 +31,7 @@
 #include "xls/ir/node_iterator.h"
 #include "xls/ir/node_util.h"
 #include "xls/ir/nodes.h"
+#include "xls/passes/post_dominator_analysis.h"
 #include "xls/passes/ternary_query_engine.h"
 
 namespace xls {
@@ -962,6 +963,40 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
               .status());
       return true;
     }
+  }
+
+  // Specialize the arms of the select based on the selector value which chooses
+  // the arm.
+  if (node->Is<Select>()) {
+    XLS_ASSIGN_OR_RETURN(
+        std::unique_ptr<PostDominatorAnalysis> post_dominator_analysis,
+        PostDominatorAnalysis::Run(node->function()));
+    Select* sel = node->As<Select>();
+    int64 selr_width = sel->selector()->BitCountOrDie();
+    bool changed = false;
+    for (int64 selr_value = 0; selr_value < sel->cases().size(); ++selr_value) {
+      Node* arm = sel->cases()[selr_value];
+      // Skip any cases where the select arm is used other than just in the
+      // select statement
+      if (arm->users().size() != 1) {
+        continue;
+      }
+      std::vector<Node*> users(sel->selector()->users().begin(),
+                               sel->selector()->users().end());
+      for (Node* selr_user : users) {
+        if (!post_dominator_analysis->NodeIsPostDominatedBy(selr_user, arm)) {
+          continue;
+        }
+        XLS_ASSIGN_OR_RETURN(
+            Literal * literal,
+            node->function()->MakeNode<Literal>(
+                sel->selector()->loc(), Value(UBits(selr_value, selr_width))));
+        if (selr_user->ReplaceOperand(sel->selector(), literal)) {
+          changed = true;
+        }
+      }
+    }
+    if (changed) return true;
   }
 
   return false;
--- a/xls/ir/node.h
+++ b/xls/ir/node.h
@@ -82,7 +82,8 @@ class Node {
   // SwapOperands() below.
   absl::Span<Node* const> operands() const { return operands_; }
 
-  // Returns whether at least one operand was replaced.
+  // Replaces all instances of 'old_operand' with 'new_operand', and returns
+  // whether at least one operand was replaced.
   bool ReplaceOperand(Node* old_operand, Node* new_operand);
 
   // Replaces the existing operand at position 'operand_no' with 'new_operand'.
--- a/xls/passes/BUILD
+++ b/xls/passes/BUILD
@@ -471,6 +471,7 @@ cc_library(
     hdrs = ["select_simplification_pass.h"],
     deps = [
         ":passes",
+        ":post_dominator_analysis",
         ":ternary_query_engine",
         "@com_google_absl//absl/algorithm:container",
         "@com_google_absl//absl/container:flat_hash_set",
--- a/xls/passes/post_dominator_analysis.h
+++ b/xls/passes/post_dominator_analysis.h
@@ -31,19 +31,20 @@ class PostDominatorAnalysis {
       Function* f);
 
   // Returns the nodes that post-dominate this node.
-  absl::Span<Node* const> GetPostDominatorsOfNode(const Node* node) {
+  absl::Span<Node* const> GetPostDominatorsOfNode(const Node* node) const {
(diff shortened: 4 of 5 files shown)


felixzhuologist:
@meheff this should be ready for a first pass, PTAL :)

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +966,41 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
     }
   }
 
+  // Specialize the arms of the select based on the selector value which chooses
+  // the arm.
+  if (node->Is<Select>()) {
+    Select* sel = node->As<Select>();
+    auto selr_width = sel->selector()->BitCountOrDie();
```
meheff, line 973: Use auto when the type is redundant (like when assigning from make_unique) but generally use the full type to improve readability. Ie, "int64 selr_width =..".

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +966,41 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
...
+      // Skip any cases where the select arm is used other than just in the
+      // select statement
+      if (arm->users().size() != 1) {
+        continue;
+      }
+      for (Node* selr_use : sel->selector()->users()) {
+        if (!post_dominator_analysis.NodeIsPostDominatedBy(selr_use, arm)) {
+          continue;
+        }
+        for (int64 operand_no = 0; operand_no < selr_use->operand_count();
```
meheff, line 987: Rather than this loop, you can create the literal then call:

selr_use->ReplaceOperand(selr, literal);

Node::ReplaceOperand could use a better comment. Mind adding something like "Replace all instances of old_operand with new_operand." in that method comment? Thanks!

meheff (changes requested):
Looks great and nice tests! Very nice first XLS CL!

meheff:
> Looks great and nice tests! Very nice first XLS CL!

Nice *second* XLS code change I should say. I forgot you worked on the DSLX front end.

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +966,41 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
     }
   }
 
+  // Specialize the arms of the select based on the selector value which chooses
+  // the arm.
+  if (node->Is<Select>()) {
+    Select* sel = node->As<Select>();
+    auto selr_width = sel->selector()->BitCountOrDie();
+    bool changed = false;
+    for (uint64 selr_value = 0; selr_value < sel->cases().size();
```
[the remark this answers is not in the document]
felixzhuologist: good to know, thanks!

felixzhuologist:
thanks! learned a lot just reading through the optimization code

felixzhuologist:
ah just realized it would have been easier to see changes if i just pushed a separate commit and then squashed later instead of force pushing, sorry about that

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +966,35 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
...
+      // Skip any cases where the select arm is used other than just in the
+      // select statement
+      if (arm->users().size() != 1) {
+        continue;
+      }
+      for (Node* selr_use : sel->selector()->users()) {
+        if (!post_dominator_analysis.NodeIsPostDominatedBy(selr_use, arm)) {
+          continue;
+        }
+        auto literal = std::make_unique<Literal>(
```
meheff, line 986: Use Function::MakeNode which automatically adds the node to the function and verifies it:

XLS_ASSIGN_OR_RETURN(Literal* literal, node->function()->MakeNode<Literal>(...));

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +966,35 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
...
+            sel->selector()->loc(), Value(UBits(selr_value, selr_width)),
+            node->function());
+        if (selr_use->ReplaceOperand(sel->selector(), literal.get())) {
+          node->function()->AddNode(std::move(literal));
```
meheff, line 990: Not needed if you use MakeNode.

meheff (approved):
Just a couple suggestions. Otherwise LGTM!

xls/passes/select_simplification_pass.cc
felixzhuologist: Sounds good, wasn't sure if it was OK to add a Literal if it never gets referenced within the Function

felixzhuologist:
Thanks for the review! 😁

cdleary:
Hey Felix, looks like CI is still failing, can you reproduce locally?

felixzhuologist:
Ah sorry I missed that, should be fixed now!

Post dominator analysis was failing because the nodes in the function were changed after it was run. I changed the code so that it runs the analysis each time before trying to specialize the select arms, assuming it's not too costly. If that's not the case, I can try running it less often (e.g. rerun only if we get to the specialize select arms step AND the function has changed since the last run)

xls/passes/select_simplification_pass.cc
```diff
@@ -964,6 +965,40 @@ absl::StatusOr<bool> SimplifyNode(Node* node, const QueryEngine& query_engine,
...
+    for (int64 selr_value = 0; selr_value < sel->cases().size(); ++selr_value) {
+      Node* arm = sel->cases()[selr_value];
+      // Skip any cases where the select arm is used other than just in the
+      // select statement
+      if (arm->users().size() != 1) {
+        continue;
+      }
+      std::vector<Node*> users(sel->selector()->users().begin(),
+                               sel->selector()->users().end());
+      for (Node* selr_user : users) {
```
felixzhuologist, line 986: Using `sel->selector()->users()` directly in the range based for loop here is what was causing the container overflow.

I'm not exactly sure why that is, but I noticed that a call to `users()` elsewhere in the code was being collected into a vector first before being iterated through and that seems to fix it.

PR #148, opened 2020-09-25
Use html checkbox checked state on initial draw

felixzhuologist:
Otherwise it can be confusing when nodes are missing in the visualization and the "only view selected" toggle is off (when in actuality the graph is only rendering the selected nodes)


felixzhuologist:
wasn't sure if this behavior was intended or not, feel free to close if that's the case!

meheff (approved):
Thanks! and sorry for missing this PR early.

Issue #141, opened 2020-09-14
Enable LLVM as default registered toolchain in Bazel builds


cdleary:
Makes the build hermetic with respect to that known toolchain -- we've worked through a few issues that have prevented turning it on so far (e.g. warnings that we hadn't seen before, m4 build configuration), but there may be some remaining. This would conceptually make use of earlier versions of Ubuntu more easily possible, but the libc DSO version requirements for the clang binary toolchains are somewhat strong so I think it wasn't working with Ubuntu < 18.04.

cdleary:
Fixed by @ericastor's work. We also decided to go this way so that we could have things like gtest fuzz tests by default in the OSS repo.

PR #151, opened 2020-10-03
Fixes for python3.9 and gcc-10.2.1

cbalint13:
Proposed PR fixes build on more recent systems.

* Increase the default 600s (way too small) timeout for **pip_deps** fetches.
* Impose a newer **cython** compatible with python 3.9 too.
* Update **grpc** to be able build with latest gcc-10.2.1 too.
* Request newer **scipy** buildable with gcc-10.2.1 and python 3.9 headers

These fixes works on a Fedora 34 (rawhide) distro.

Thank You !

----------

PS:
* Preparing more patches towards (optional) distro friendly **unbundled** builds, (like ```systemlibs``` option for tensorflow)
#141

### DIFF
--- a/dependency_support/com_github_grpc_grpc/BUILD
+++ b/dependency_support/com_github_grpc_grpc/BUILD
@@ -1,0 +1,15 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


cdleary:
Cool, thanks for the contribution Balint!

One preliminary question I have as I'm taking a look: do you happen to know if there a way to test this configuration via GitHub actions? I worry we'll accidentally regress it without knowing unless we have some signal, but best effort maybe is ok if we have no way to do something continuous.

cdleary (commented):
Could you add the apache2 license header to your newly introduced BUILD file? I know the file contains nothing but we have a downstream linter that ensures it's there in all BUILD files.

cbalint13:
> Cool, thanks for the contribution Balint!
> 
> One preliminary question I have as I'm taking a look: do you happen to know if there a way to test this configuration via GitHub actions? I worry we'll accidentally regress it without knowing unless we have some signal, but best effort maybe is ok if we have no way to do something continuous.

PR #154, opened 2020-10-06
Added ir_integrator

jbaileyhandle:
...which will be used to combine multiple functions into a single function capable of implmenting all original functions. The initial code copies the source functions into a new package where they can be combined. Functions to be merged should be in the same package to ensure type compatibility. This required extending Function's clone method to enable call site remapping.  This way, we can recursively copy functions and their callees such that the cloned caller function calls the cloned callee function. Additionally, a GetType(Type*) function is added to Package.  This allows us to take a Type from one package and fetch / construct a corresponding Type in another package.

### DIFF
--- a/xls/contrib/integrator/ir_integrator.cc
+++ b/xls/contrib/integrator/ir_integrator.cc
@@ -1,0 +1,88 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+#include "xls/contrib/integrator/ir_integrator.h"
+
+#include "xls/ir/ir_parser.h"
+
+namespace xls {
+
+absl::StatusOr<Function*> IntegrationBuilder::CloneFunctionRecursive(
+    const Function* function,
+    absl::flat_hash_map<const Function*, Function*>* call_remapping) {
+  // Collect callee functions.
+  std::vector<const Function*> callee_funcs;
+  for (const Node* node : function->nodes()) {
+    switch (node->op()) {
+      case Op::kCountedFor:
+        callee_funcs.push_back(node->As<CountedFor>()->body());
+        break;
+      case Op::kMap:
+        callee_funcs.push_back(node->As<Map>()->to_apply());
+        break;
+      case Op::kInvoke:
+        callee_funcs.push_back(node->As<Invoke>()->to_apply());
+        break;
+      default:
+        break;
+    }
+  }
+
+  // Clone and call_remapping callees.
+  for (const Function* callee : callee_funcs) {
+    if (!call_remapping->contains(callee)) {
+      XLS_ASSIGN_OR_RETURN(Function * callee_clone,
+                           CloneFunctionRecursive(callee, call_remapping));
+      (*call_remapping)[callee] = callee_clone;
+    }
+  }
+
+  std::string clone_name =
+      function_name_uniquer_.GetSanitizedUniqueName(function->name());
+  return function->Clone(clone_name, package_.get(), *call_remapping);
+}
+
+absl::Status IntegrationBuilder::CopySourcesToIntegrationPackage() {
+  source_functions_.reserve(original_package_source_functions_.size());
+  for (const Function* source : original_package_source_functions_) {
+    absl::flat_hash_map<const Function*, Function*> call_remapping;
+    XLS_ASSIGN_OR_RETURN(Function * clone_func,
+                         CloneFunctionRecursive(source, &call_remapping));
+    source_functions_.push_back(clone_func);
+  }
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::unique_ptr<IntegrationBuilder>> IntegrationBuilder::Build(
+    absl::Span<const Function* const> input_functions) {
+  auto builder = absl::WrapUnique(new IntegrationBuilder(input_functions));
+
+  // Add sources to common package.
+  XLS_RETURN_IF_ERROR(builder->CopySourcesToIntegrationPackage());
+
+  switch (builder->source_functions_.size()) {
+    case 0:
+      return absl::InternalError(
+          "No source functions provided for integration");
+    case 1:
+      builder->integrated_function_ = builder->source_functions_.front();
+      break;
+    default:
+      return absl::InternalError("Integration not yet implemented.");
+  }
+
+  return std::move(builder);
+}
+
+}  // namespace xls
--- a/xls/contrib/integrator/ir_integrator.h
+++ b/xls/contrib/integrator/ir_integrator.h
@@ -1,0 +1,75 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef XLS_INTEGRATOR_IR_INTEGRATOR_H_
+#define XLS_INTEGRATOR_IR_INTEGRATOR_H_
+
+#include "absl/status/statusor.h"
+#include "xls/ir/function.h"
+#include "xls/ir/package.h"
+
+namespace xls {
+
+// Class used to integrate separate functions into a combined, reprogrammable
+// circuit that can be configured to have the same functionality as the
+// input functions. The builder will attempt to construct the integrated
+// function such that hardware common to the input functions is consolidated.
+// Note that this is distinct from function inlining. With inlining, a function
+// call is replaced by the body of the function that is called.  With function
+// integration, we take separate functions that do not call each other and
+// combine the hardware used to implement the functions.
+class IntegrationBuilder {
+ public:
+  // Creates an IntegrationBuilder and uses it to produce an integrated function
+  // implementing all functions in source_functions_.
+  static absl::StatusOr<std::unique_ptr<IntegrationBuilder>> Build(
+      absl::Span<const Function* const> input_functions);
+
+  Package* package() { return package_.get(); }
+  Function* integrated_function() { return integrated_function_; }
+
+ private:
+  IntegrationBuilder(absl::Span<const Function* const> input_functions) {
+    original_package_source_functions_.insert(
+        original_package_source_functions_.end(), input_functions.begin(),
+        input_functions.end());
+    // TODO(jbaileyhandle): Make package name an optional argument.
+    package_ = std::move(absl::make_unique<Package>("IntegrationPackage"));
+  }
+
+  // Copy the source functions into a common package.
+  absl::Status CopySourcesToIntegrationPackage();
+
+  // Recursively copy a function into the common package_.
+  absl::StatusOr<Function*> CloneFunctionRecursive(
+      const Function* function,
+      absl::flat_hash_map<const Function*, Function*>* call_remapping);
+
+  NameUniquer function_name_uniquer_ = NameUniquer(/*separator=*/"__");
+
+  // Common package for to-be integrated functions
+  // and integrated function.
+  std::unique_ptr<Package> package_;
+
+  Function* integrated_function_;
+
+  // Functions to be integrated, in the integration package.
+  std::vector<Function*> source_functions_;
+  // Functions to be integrated, in their original packages.
+  std::vector<const Function*> original_package_source_functions_;
+};
+
+}  // namespace xls
+
+#endif  // XLS_INTEGRATOR_IR_INTEGRATOR_H_
--- a/xls/contrib/integrator/ir_integrator_test.cc
+++ b/xls/contrib/integrator/ir_integrator_test.cc
@@ -1,0 +1,125 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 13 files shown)


xls/integrator/ir_integrator.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef XLS_INTEGRATOR_IR_INTEGRATOR_H_
+#define XLS_INTEGRATOR_IR_INTEGRATOR_H_
+
+#include "absl/status/statusor.h"
+#include "xls/ir/function.h"
+#include "xls/ir/package.h"
+
+namespace xls {
+
+class IntegrationBuilder {
```
RobSpringer, line 24: Class comment, plz!

RobSpringer: Also, based on the CL description itself, it's not immediately clear how an "integrated" function differs from a function after an inlining pass (that being said, after looking @ the CL, it's a lot clearer WRT maps/fors) - please be sure to add that info to this comment.

RobSpringer (changes requested):
What's the long-term plan w/this utility? It seems potentially useful, but I don't actually know to what end (I suspect Mark, at least, does, but even so, we should probably capture this info in the public repo).

xls/integrator/ir_integrator.h
jbaileyhandle: Comment added.

The long term plan is to take 2 or more separte functions / circuits and create a single, reprogramamble circuit that can be configured to have the same functionality as the input circuits.

So if we input
func AXpB = A * X + b
and 
func AX = A*X

We can combine into something like
tmp = A*X
return config_signal ? tmp : tmp + b

xls/integrator/ir_integrator.cc
```diff
@@ -0,0 +1,82 @@
...
+
+namespace xls {
+
+// Return the name of 'function', prepended with the 'package' name.
+std::string GetParsableQualifiedFunctionName(Package* package,
+                                             const Function* function) {
+  return "PKGzzz" + package->name() + "zzzFNzzz" + function->name();
+}
+
+absl::Status IntegrationBuilder::CopySourcesToIntegrationPackage() {
```
meheff, line 27: Any reason why you need to copy the functions to a new package rather than create new functions in the existing package? And if you do need it, would an xls::Package::Clone method do what you need?

xls/integrator/ir_integrator.h
```diff
@@ -0,0 +1,63 @@
...
+#include "absl/status/statusor.h"
+#include "xls/ir/function.h"
+#include "xls/ir/package.h"
+
+namespace xls {
+
+// Class used to integrate separate functions into a combined, reprogrammable
+// circuit that can be configured to have the same functionality as the
+// input functions. The builder will attempt to construct the integrated
+// funciton such that hardware common to the input functions is consolidated.
```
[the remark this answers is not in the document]
jbaileyhandle: Done

xls/integrator/ir_integrator.cc
jbaileyhandle: Currently, there is a 2-step process:

1.  Copy input packages into duplicate packages, then rename the functions to avoid later name-space collisions.  This is partly a nicety to avoid modifying the input packages.  It also helps to avoid issues that could come up in modifying things in-place in the original package (e.g. orignal has funcA and funcB, where funcA calls funcB.  In the same package, we make copies copy_funcA and copy_funcB.  Unless we traverse and fix the copied functions, copy_funcA calls funcB instead of copy_funcB.  Fixup is more work and will likely require adjustments as other things ie procs, other invocation types etc are added).  This part could be accomplished with a Package::Clone method if it existed.

2.  We copy the renamed functions to a common package.  Here, I am not assuming that the input functions are from the same source package.  The assumption is that functions that will be integrated were developed independently, are not in the same source package, may have name collisions, and were likely scanned in from separate .ir files.  We need to add them to a common package to ensure type compatability.  This part cannot be avoid with a Package::Clone method.

meheff: I'd guess you'll need a Function::Clone method which takes a map<Function*,Function*> to remap call sites.

xls/integrator/ir_integrator.cc
```diff
@@ -0,0 +1,82 @@
+// Copyright 2020 Google LLC
```
meheff, line 1: How about putting this in xls/contrib/integrator? Once its more developed and more generally useful we can move it to the main directory. As is, the use case is pretty specific.

xls/integrator/ir_integrator.cc
```diff
@@ -0,0 +1,82 @@
...
+#include "xls/integrator/ir_integrator.h"
+
+#include "xls/ir/ir_parser.h"
+
+namespace xls {
+
+// Return the name of 'function', prepended with the 'package' name.
+std::string GetParsableQualifiedFunctionName(Package* package,
+                                             const Function* function) {
+  return "PKGzzz" + package->name() + "zzzFNzzz" + function->name();
```
meheff, line 24: PKGzzz? If you need to generate unique_names, check out xls/ir/name_uniquer.

xls/integrator/ir_integrator.cc
```diff
@@ -0,0 +1,82 @@
...
+
+    // Change all function names before dumping any function so that
+    // all function invocations refer to the new function name.
+    for (const auto& function : srctmp_package->functions()) {
+      function->SetName(
+          GetParsableQualifiedFunctionName(source->package(), function.get()));
+    }
+
+    // Dump functions to common IR string.
+    for (const auto& function : srctmp_package->functions()) {
```
meheff, line 47: I'm guessing dumping functions from different packages into the same IR file will fail because of duplicate node ids which must be unique in the package scope. Using Function::Clone avoids this issue.

xls/integrator/ir_integrator.cc
jbaileyhandle: If we wanted to add new funcitons in the original package, yes. but I think it's better to copy the original package than to add stuff to it.

meheff: And it looks like we do have a Function::Clone. It just needs to be modified to take a target package and callsite remapping.

xls/integrator/ir_integrator.h
```diff
@@ -0,0 +1,63 @@
...
+// circuit that can be configured to have the same functionality as the
+// input functions. The builder will attempt to construct the integrated
+// function such that hardware common to the input functions is consolidated.
+// Note that this is distinct from function inlining. With inlining, a function
+// call is replaced by the body of the function that is called.  With function
+// integration, we take separate functions that do not call each other and
+// combine the hardware used to implement the functions.
+class IntegrationBuilder {
+ public:
+  IntegrationBuilder(absl::Span<const Function* const> input_functions) {
```
meheff, line 34: How about making the constructor private and just having a static build method. It could return a std::unique_ptr<Package> and a Function pointer. This avoids potential ownership issues where you require the original functions to be kept around until Build is called.

meheff (changes requested):

xls/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/integrator/ir_integrator.h
jbaileyhandle: Done.  However, absl::make_unique needs access to a public constructor to work.  I implemented a hacky workaround where we call make_unqiue for a private sublcass with a public constructor, similar to the one presented here:
https://seanmiddleditch.com/enabling-make-unique-with-private-constructors/

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -0,0 +1,97 @@
...
+  }
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::unique_ptr<IntegrationBuilder>> IntegrationBuilder::Build(
+    absl::Span<const Function* const> input_functions) {
+  // make_unique does not have access to the private IntegrationBuilder
+  // constructor. This is a workaround that lets us make an IntegrationBuilder
+  // unique_ptr.
+  struct make_unique_compatible_builder : public IntegrationBuilder {
```
meheff, line 72: You can use absl::WrapUnique for this purpose:

auto builder = absl::WrapUnique(new IntegrationBuilder(...));

xls/contrib/integrator/ir_integrator.h
```diff
@@ -0,0 +1,77 @@
...
+  // Creates an IntegrationBuilder and uses it to produce an integrated function
+  // implementing all functions in source_functions_.
+  static absl::StatusOr<std::unique_ptr<IntegrationBuilder>> Build(
+      absl::Span<const Function* const> input_functions);
+
+  Package* package() { return package_.get(); }
+  Function* integrated_function() { return integrated_function_; }
+
+ private:
+  // IntegrationBuilder() {}
```
meheff, line 43: left over comment?

xls/contrib/integrator/ir_integrator_test.cc
```diff
@@ -0,0 +1,154 @@
...
+  XLS_ASSERT_OK_AND_ASSIGN(Node * map_original_generic,
+                           main_func->GetNode("map"));
+  Map* map_original = map_original_generic->As<Map>();
+  EXPECT_EQ(map_original->to_apply(), double_func);
+  EXPECT_EQ(map_original->to_apply()->name(), "double");
+  EXPECT_EQ(map_original->to_apply()->package(), p.get());
+
+  // Original Invoke
+  XLS_ASSERT_OK_AND_ASSIGN(Node * invoke_original_generic,
+                           main_func->GetNode("invoke"));
```
meheff, line 113: IrTestBase has FindNode for this purpose:

```
Invoke* invoke_orignal = FindNode("invoke", main_func)->As<Invoke>();
```
It checkfails if it can't find the node but that's fine because it's a test. Saves you from having to unwrap the StatusOr.

xls/ir/function.h
```diff
@@ -144,8 +150,15 @@ class Function {
...
   // Creates a clone of the function with the new name 'new_name'. Function is
-  // owned by the same package.
-  absl::StatusOr<Function*> Clone(absl::string_view new_name) const;
+  // owned by targt_package.  call_remapping specifies any function
+  // substitutions to be used in the cloned function, e.g. If call_remapping
+  // holds {funcA, funcB}, any references to funcA in the function will be
+  // references to funcB in the cloned function.
+  absl::StatusOr<Function*> Clone(
+      absl::string_view new_name,
+      absl::optional<Package*> targt_package = std::nullopt,
```
meheff, line 159: Make this a plain pointer with an optional value of nullptr. An optional pointer has too many states: nullopt, nullptr, real value.

xls/ir/function.h
```diff
@@ -144,8 +150,15 @@ class Function {
...
+      absl::optional<const absl::flat_hash_map<const Function*, Function*>*>
+          call_remapping = std::nullopt) const;
```
meheff, line 161: Can you make this

```
const absl::flat_hash_map<...>& = {}
```

xls/ir/nodes_source.tmpl
```diff
@@ -105,6 +105,29 @@ SliceData Concat::GetOperandSliceData(int64 operandno) const {
   return SliceData{start, this->operand(operandno)->BitCountOrDie()};
 }
 
+absl::StatusOr<Node*> Param::Clone(absl::Span<Node* const> new_operands,
+                                           Function* new_function) const {
+  // TODO(meheff): Choose an appropriate name for the cloned node.
+  XLS_RET_CHECK_EQ(operand_count(), new_operands.size());
+  XLS_ASSIGN_OR_RETURN(Type* new_type,
```
meheff, line 112: Hmm, ick. I guess there is no way around this. We need something higher scoped to hold the types. Nothing for you to do.

xls/ir/package.h
```diff
@@ -59,6 +59,7 @@ class Package {
            owned_function_types_.end();
   }
 
+  absl::StatusOr<Type*> GetType(const Type* type);
```
meheff, line 62: Just to be very clear what this does, how about calling this MapTypeFromOtherPackage or something like that. And add a method comment.

xls/ir/package.cc
```diff
@@ -181,6 +181,34 @@ std::string Package::SourceLocationToString(const SourceLocation loc) {
   return absl::StrFormat("%s:%d", filename, loc.lineno().value());
 }
 
+absl::StatusOr<Type*> Package::GetType(const Type* type) {
```
meheff, line 184: How about checking that type is not owned by this package?

meheff (changes requested):
Thanks!

xls/ir/package.cc
jbaileyhandle: Do you mean short circuit and return the type if it is already owned or return an error if it is already owened?  I prefer the former.

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/ir/package.cc
jbaileyhandle: To elaborate, the use case for function that I imagine is "I have a type of unknown provenance, give me a version that I know belongs to this package".  This seems more useful than making the caller check which package owns the type and crashing if they messed up.

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Neat + done.

xls/contrib/integrator/ir_integrator.h
jbaileyhandle: Done.

xls/contrib/integrator/ir_integrator_test.cc
jbaileyhandle: Done.

xls/ir/function.h
jbaileyhandle: Done.

xls/ir/function.h
jbaileyhandle: Done.

xls/ir/nodes_source.tmpl
jbaileyhandle: Ack.

xls/ir/package.h
jbaileyhandle: Done.

xls/ir/package.cc
meheff: Sounds fine. The way things are currently implemented, mixing packages exposes a lot of foot guns around types. Probably want to have a singleton type manager so there is no potential for accidentally using a type in the wrong context.

meheff (approved):
Thanks!

PR #155, opened 2020-10-06
Fix typo in README.md install instructions

cjdrake:
The current version tells the user to install python3-dev twice.
I think the intention is to install both python-dev *and* python3-dev,
but there is a typo here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ installed](https://docs.bazel.build/versions/master/install-ubuntu.html).
 $ bazel --version
 bazel 3.2.0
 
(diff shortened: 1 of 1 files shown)


cdleary:
Thanks for the fix Chris! I think we should be python3 only so I think python3-dev is sufficient if you delete python-dev entirely I think it should all work. That looks like what the Dockerfile is doing -- "documentation needing to be kept in sync" problems because we can't have a dynamic README file. :-)

cjdrake:
Okay, SGTM.

PR #171, opened 2020-10-28
Eliminate the need for "the container hack".

### LINKED ISSUE #170: Debug "the container hack" (!)
Something about mixing pybind11 objects, absl::flat_* (e.g. flat_hash_map), and gcc9 causes breakages in OSS builds that don't appear on internal (clang-based) builds. They seem quite reproducible -- if we exchange the references to third_party/xls/ir/container_hack.inc with their absl map counterparts, the errors should be observable. Should be a bit of a debugging challenge, if somebody is feeling adventurous.

cdleary:
Fixes #170.

When objects were being passed from one pybind11 translation unit to another
invariant violations were observed, like absl::flat_hash_map().find() not
finding keys that were present when iterating all the entries and checking for
key presence. "The container hack" used the std::*{map,set} types because the
issue was not observed for those types.

Turns out this is because [absl cleverly uses a symbol's address as a seed for
its hashing algorithm, which is randomized via
ASLR](https://t.co/xKnjSu1dLC?amp=1) on supported operating systems; however,
when the symbol is statically linked into shared objects, instead of
dynamically linked / exposed as undef, each DSO would have a unique symbol
definition, and thus use a different hashing seed, so the resolved hash bucket
would differ between the main (static) portion of the binary and each of the
dynamic python extension modules.

Root cause is that we were creating "statically linked shared libraries", where
the contents of all linked object files were being linked statically into the
.so -- by doing "dynamically linked shared libraries" we get undef symbols
appropriately, the seed symbol resolved against the single (global) ASLR
virtual memory location, and all the modules now probe hash tables the same
way.

Huzzah!


RobSpringer (approved):
Simply beautiful work.

meheff (approved):
Nice!

cdleary:
Thanks. :-) Debugging these sorts of "maze of twisty little passages, all alike" bugs is actually fun so long as it's not all the time!

PR #179, opened 2020-11-03
Functions for merging nodes.

jbaileyhandle:
Added functions to merge two nodes and estimate the cost of merging two nodes. These frontend function call a common backend function. Sharing a common backend will let us implement a given merge optimization (e.g. merge adds of different bitwidths) in a single place, rather than having duplicate logic to estimate the cost of a particular optimization and to actually implement the optimization. To enable merging and cost estimation, added a UnifyNodeOperands function to unify all operands of two nodes and a DeUnifyIntegrationNodes function to undo a previous unification.

### DIFF
--- a/xls/contrib/integrator/ir_integrator.cc
+++ b/xls/contrib/integrator/ir_integrator.cc
@@ -175,6 +175,54 @@ IntegrationFunction::UnifyIntegrationNodes(Node* node_a, Node* node_b) {
   return UnifiedNode({mux, /*new_mux_added=*/true});
 }
 
+absl::StatusOr<IntegrationFunction::UnifiedOperands>
+IntegrationFunction::UnifyNodeOperands(const Node* node_a, const Node* node_b) {
+  XLS_RET_CHECK_EQ(node_a->operands().size(), node_b->operands().size());
+
+  // Get mapped operands.
+  XLS_ASSIGN_OR_RETURN(std::vector<Node*> a_ops, GetIntegratedOperands(node_a));
+  XLS_ASSIGN_OR_RETURN(std::vector<Node*> b_ops, GetIntegratedOperands(node_b));
+
+  // Unify operands.
+  UnifiedOperands unify_result;
+  unify_result.operands.reserve(a_ops.size());
+  for (int64 idx = 0; idx < a_ops.size(); ++idx) {
+    XLS_ASSIGN_OR_RETURN(UnifiedNode uni_op,
+                         UnifyIntegrationNodes(a_ops.at(idx), b_ops.at(idx)));
+    unify_result.operands.push_back(uni_op.node);
+    if (uni_op.new_mux_added) {
+      unify_result.added_muxes.push_back(uni_op.node);
+    }
+  }
+
+  return unify_result;
+}
+
+absl::Status IntegrationFunction::DeUnifyIntegrationNodes(Node* node) {
+  XLS_RET_CHECK_EQ(node->op(), Op::kSel);
+  Select* mux = node->As<Select>();
+  XLS_RET_CHECK(IntegrationFunctionOwnsNode(mux));
+  XLS_RET_CHECK_EQ(mux->cases().size(), 2);
+  Node* a_in = mux->cases().at(0);
+  Node* b_in = mux->cases().at(1);
+
+  // Clean up bookkeeping.
+  std::pair<const Node*, const Node*> key = std::make_pair(a_in, b_in);
+  XLS_RET_CHECK(node_pair_to_mux_.contains(key));
+  XLS_RET_CHECK_EQ(node_pair_to_mux_.at(key), mux);
+  node_pair_to_mux_.erase(key);
+
+  // Clean up nodes.
+  Node* selector = mux->selector();
+  XLS_RET_CHECK(mux->users().empty());
+  XLS_RETURN_IF_ERROR(function()->RemoveNode(mux));
+  XLS_RET_CHECK(selector->users().empty());
+  XLS_RETURN_IF_ERROR(
+      function()->RemoveNode(selector, /*remove_param_ok=*/true));
+
+  return absl::OkStatus();
+}
+
 bool IntegrationFunction::HasMapping(const Node* node) const {
   return original_node_to_integrated_node_map_.contains(node);
 }
@@ -184,6 +232,8 @@ bool IntegrationFunction::IsMappingTarget(const Node* node) const {
 }
 
 absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
+  // TODO(jbaileyhandle): Implement this by calling MergeNodes behind
+  // the scenes.
   XLS_RET_CHECK(!IntegrationFunctionOwnsNode(to_insert));
   XLS_RET_CHECK(!HasMapping(to_insert));
   XLS_ASSIGN_OR_RETURN(std::vector<Node*> mapped_operands,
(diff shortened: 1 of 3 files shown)


xls/contrib/integrator/ir_integrator.cc
```diff
@@ -194,6 +244,168 @@ absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
...
+  tabulate_node_elimination_cost(node_a);
+  tabulate_node_elimination_cost(node_b);
+  for (Node* new_node : merge_result.added_muxes) {
+    cost += GetNodeCost(new_node);
+  }
+  for (Node* new_node : merge_result.other_added_nodes) {
+    cost += GetNodeCost(new_node);
+  }
+
+  // Cleanup.
```
meheff, line 328: Is there a way to compute the cost without mutating the graph? It is a bit surprising that a method called GetMergeNodeCost adds nodes to the graph then removes them.

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -194,6 +244,168 @@ absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
...
+  XLS_ASSIGN_OR_RETURN(MergeNodesBackendResult merge_result,
+                       MergeNodesBackend(node_a, node_b));
+  // Can't merge nodes.
+  XLS_RET_CHECK(merge_result.can_merge);
+  XLS_RET_CHECK(merge_result.target_a != nullptr);
+  XLS_RET_CHECK(merge_result.target_b != nullptr);
+
+  // Commit changes.
+  auto commit_new_node = [this](Node* original, Node* target) {
+    auto map_status = SetNodeMapping(original, target);
```
meheff, line 382: You can use macro here XLS_ASSIGN_OR_RETURN()

I believe you'll also have to declare the return type of the lambda:

auto commit_new_node = [...](...) -> absl::Status {

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -194,6 +244,168 @@ absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
...
+      cost -= GetNodeCost(node);
+    }
+  };
+  tabulate_node_elimination_cost(node_a);
+  tabulate_node_elimination_cost(node_b);
+  for (Node* new_node : merge_result.added_muxes) {
+    cost += GetNodeCost(new_node);
+  }
+  for (Node* new_node : merge_result.other_added_nodes) {
+    cost += GetNodeCost(new_node);
```
meheff, line 325: The iteration order here is not deterministic and floating point addition is not associative so this may produce nondeterministic results. Can the cost be an int64? ints have much nicer behavior.

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -194,6 +244,168 @@ absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
...
+  }
+
+  // Cleanup.
+  int64 nodes_eliminated = 0;
+  int64 num_other_nodes_to_remove = merge_result.other_added_nodes.size();
+  while (nodes_eliminated < num_other_nodes_to_remove) {
+    int64 initial_eliminated = nodes_eliminated;
+    for (Node* new_node : merge_result.other_added_nodes) {
```
meheff, line 333: Avoid iterating over a set when doing something mutating. The order is non-determistic and can have knock on effects on the data structures you're changing.

meheff (changes requested):

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: I think it's the lesser of two evils. I had a couple of things in mind that motivated this decision:

-We'll want to add different ways to merge non-identical nodes (simple example: merge adds of different bitwidths, then use a bitslice to create a node for the narrower of the two merged nodes, implement.  Other non-identical merges may involve adding other nodes that are more costly than bitslice / can't be assumed to be free).  If we have a non-mutating GetMergeNodesCost, we'll have to duplicate most of the logic for merging and for getting the cost of a merge (e.g. we'll have "do this" and "what would I do" version of all merge code).  This is both a waste of effort and is likely to introduce errors, for instance if we implement the merge + cost versions of a techinque but then later update only the merge version.

-Similar to the above, the do vs. estimate split would likely bleed out to code beyond just that directly implementing merging.  For instance, we have functions to mux together nodes that are used on operands when merging nodes.  This, too, would need to be split into two versions.  Further, it could be difficult to do this correctly.  For instance, if we need to mux together the same two nodes for multiple purposes, we reuse the same mux.  If we are trying to estimate the cost of merging two nodes where we have two operands that will mux together the same nodes (e.g merge (Node(a,a) and Node(b,b) -> Node(mux(a,b, mux(a,b)), we'd need to only count the cost of adding one mux.  This gets tricky - currently, GetNodesMergeCost outsources the operand muxing to UnifyNodeOperands, which outsources per operand to UnifyIntegrationNodes. UnifyIntegrationNodes can checks some bookkeping to see if we've made this mux before.  For cost-only functions, we haven't modified this global book-keeping, so we'd have to handle duplicate checking higher up in the call stack.  Said another way, helper functions might have to be collapsed into cost estimation functions.

-Modularity would likely be hurt.  There's the helper function collapsing issue mention above.  Beyond that, we'd have to keep the the merge vs cost split in mind when designing other code.  There may be solutions to the merge vs cost issues we see in the code now, but I'm worried those solutions would break if assumptions or other design components change e.g. merge techniques only add 1 costly node and the others are trivial things like bit slices or oncats at some point in developement, so we decide we only have to id the coslty node, but this breaks if we add more complicated merge techniques. Or similarly, perhaps we make assumptions about how we add muxes or inline functions or track node mappings or any other number of things at some point, but we may change these later.  The unkown uknowns seem to present a lot of risk here.

In summary, while I agree that having a non-mutating cost estimate is conceptually cleaner, I think this would likely lead to a lot of pain and suffering on the implementation end and open up possibilies for bugs.  Another approach to estimating cost without mutating the graph is to clone the function, estimate the cost of the whole graph, apply the merge on the cloned function, and then see how much the cost changed.  However, this seems very inefficient.

meheff: sounds reasonable. Thanks for the explanation. My only suggestion then is to make it as simple as possible to rollback the nodes that were added to minimize the possibility of leaving stuff in the graph.

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -194,6 +244,168 @@ absl::StatusOr<Node*> IntegrationFunction::InsertNode(const Node* to_insert) {
...
+    XLS_ASSIGN_OR_RETURN(
+        Node * merged,
+        node_a->CloneInNewFunction(unified_operands.operands, function()));
+    MergeNodesBackendResult merge_result;
+    merge_result.can_merge = true;
+    merge_result.target_a = merged;
+    merge_result.target_b = merged;
+    merge_result.added_muxes = unified_operands.added_muxes;
+    merge_result.other_added_nodes.insert(merged);
+    return merge_result;
```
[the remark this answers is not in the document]
jbaileyhandle: Done.  Every day I learn about a new bit of hideous and amazing c++ syntax

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Done (although order is now deterministic)

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Done.

meheff (approved):

PR #183, opened 2020-11-06
Adding DynamicCountedFor op to the IR.

jbaileyhandle:
This is like CountedFor, except that the trip count and stride are dynamic. Plumbing has been added along with some implementation components (function builder, ir matcher, parser, verifier).

### DIFF
--- a/docs_src/ir_semantics.md
+++ b/docs_src/ir_semantics.md
@@ -673,6 +673,60 @@ body(i, loop_carry_data[, invariant_arg0, invariant_arg1, ...])
 Note that we currently inspect the `body` function to see what type of induction
 variable (`i` above) it accepts in order to pass an `i` value of that type.
 
+#### **`dynamic_counted_for`**
+
+Invokes a dynamic-trip count loop.
+
+**Syntax**
+
+```
+result = counted_for(init, trip_count, stride, body=<body>, invariant_args=<inv_args>)
+```
+
+**Types**
+
+Value        | Type
+------------ | ----
+`init`       | `T`
+`trip_count` | `bits[N], treated as unsigned`
+`stride`     | `bits[M], treated as signed`,
+`result`     | `T`
+
+**Keyword arguments**
+
+| Keyword          | Type     | Required | Default | Description               |
+| ---------------- | -------- | -------- | ------- | ------------------------- |
+| `invariant_args` | array of | yes      |         | Names of the invariant    |
+:                  : operands :          :         : operands as the loop body :
+| `body`           | `string` | yes      |         | Name of the function to   |
+:                  :          :          :         : use as the loop body      :
+
+`dynamic_counted_for` invokes the function `body` `trip_count` times, passing
+loop-carried data that starts with value `init`. The induction variable is 
+incremented by `stride` after each iteration.
+
+*   The first argument passed to `body` is the induction variable -- presently,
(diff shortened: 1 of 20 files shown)


xls/ir/function_builder.h
```diff
@@ -396,6 +396,29 @@ class BuilderBase {
                     absl::optional<SourceLocation> loc = absl::nullopt,
                     absl::string_view name = "");
 
+  // Adds a "dynamic counted for-loop" to the computation, having an
+  // unknown-constant number of loop iterations and stride.
```
meheff, line 400: unknown-constant? Is it just "unknown"?

docs_src/ir_semantics.md
```diff
@@ -673,6 +673,60 @@ body(i, loop_carry_data[, invariant_arg0, invariant_arg1, ...])
...
+```
+result = counted_for(init, trip_count, stride, body=<body>, invariant_args=<inv_args>)
+```
+
+**Types**
+
+Value        | Type
+------------ | ----
+`init`       | `T`
+`trip_count` | `bits[N], treated as unsigned`
```
meheff, line 691: I realize we call it "trip_count" in counted_for, but it is not really trip count. It is a bound on the range of the  induction variable. So if you have trip_count=10, stride=2, the trip count is actually 5. Mind filing a github issue for this? thanks.

meheff (changes requested):

docs_src/ir_semantics.md
```diff
@@ -673,6 +673,60 @@ body(i, loop_carry_data[, invariant_arg0, invariant_arg1, ...])
...
+
+Therefore `body` should have a signature that matches the following:
+
+```
+body(i, loop_carry_data, [invariant_arg0, invariant_arg1, ...])
+```
+
+Note that we currently inspect the `body` function to see what type of induction
+variable (`i` above) it accepts in order to pass an `i` value of that type.
+`trip_count` must have fewer bits than `i` and `stride` should have fewer than
```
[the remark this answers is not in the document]
jbaileyhandle: fewer than.  trip_count is treated as unsigned, while stride and index are treated as signed.  To do signed arithmetic, we zero-extend trip_count by 1 bit.  However, we don't want the trip_count to have more bits than the index after sign extension.

xls/ir/function_builder.h
jbaileyhandle: Clarified.

docs_src/ir_semantics.md
jbaileyhandle: The implementations of the unroll pass (https://github.com/google/xls/blob/main/xls/passes/unroll_pass.cc#L43), ir interpreter (https://github.com/google/xls/blob/main/xls/interpreter/ir_interpreter.cc#L219), and llvm jit (https://github.com/google/xls/blob/main/xls/jit/function_builder_visitor.cc#L349) all treat trip_count as the number of trips, where the maximum value of the induction variable is (trip_count - 1) * stride.

meheff: You are right. I misread the interpreter implementation. Sorry about that!

meheff (approved):

jbaileyhandle:
It looks like someone with write access will have to resolve the conflict in one of the files (looks trivial, though).

PR #213, opened 2020-11-20
"Added DynamicCountedFor to IR interpreter and jit."

jbaileyhandle:


xls/interpreter/ir_interpreter.cc
```diff
@@ -256,7 +256,44 @@ absl::Status IrInterpreter::HandleDecode(Decode* decode) {
...
+  // interpreter Run() on the body function -- the new accumulator value is the
+  // return value of interpreting body.
+  while (!bits_ops::SEqual(index, index_limit)) {
+    std::vector<Value> args_for_body = {Value(index), loop_state};
+    for (const auto& value : invariant_args) {
+      args_for_body.push_back(value);
+    }
+    XLS_ASSIGN_OR_RETURN(loop_state, Run(body, args_for_body, stats_));
+
+    index = bits_ops::Add(index, extended_stride);
```
meheff, line 293: How do we know that this never overflows? 

And relatedly do we check that the index parameter to the body is wide enough to hold the max value?

meheff (changes requested):
Nice!

xls/interpreter/ir_interpreter.cc
jbaileyhandle: At the interpeter / jit level, we don't know and can't control if the index is wide enough / doesn't overflow.  The interpeter's (and jit's) job is just to execute the code faithfully, even if the code has an infinite loop.  This is the same as with something like c++ - if the input code has an infinite loop, the computer should execute an infinite loop.

However, we will be able to avoid infinite loops in practice because we should only ever emit a dynamic counted for by combining 2 or more static counted fors.  In this case, we can determined the width of the index for each static counted for and then set the width of the index for the dynamic counted for to be the maximum of these.

meheff (approved):

Issue #142, opened 2020-09-22
Create FPGA LUT-to-Z3 translator


RobSpringer:
To enable LEC'ing FPGA designs, we need a means of translating, e.g., LUT4-based designs into AND/OR/NOT/etc.-defined netlists, since that's the language our LEC tools currently speak.

julianviera@ started on this work during his internship, and demonstrated some good proof-of-concept work, so it'd be good to fully flesh out the work. IIRC, the necessary steps are:
 - ID the functionality of each LUT config ("init"?) in terms of AND/OR/NOT, etc.
 - Rewrite each LUT in terms of those operations (may require declaring add'l cells/wires)

...that's really it. I don't know that it makes sense to define a Bazel macro to do this, since the LUT mapping is potentially per-target, unless we make some default option that can be overridden by a config flag...but I'm not actually sure if that's possible.

I'll initially assign the issue to me, to make sure it doesn't get lost, but this can be picked up by anyone. It's really a decent starter project.

felixzhuologist:
An update on the status of this work and next steps:

#219 adds support for iCE40 LUTs by hardcoding the netlist parser to handle cells called `SB_LUT4`. It handles the LUTs by creating a new `CellLibraryEntry` with a `StateTable` that matches the LUT mask (or retrieving it if it exists already), instead of looking it up in a provided `CellLibrary` which is the direction suggested in the description above.

The next steps could go in two (somewhat related) directions:
* There are [more cells](https://github.com/phanrahan/mantle/blob/master/docs/ice40.md) beyond `SB_LUT4` that need to be supported for the iCE40, though I think just implementing `SB_DFF` would enable a full LEC flow to work with XLS (e.g. with the `add1_8b.x` program). I'm not sure if it would make more sense to create these cells in the netlist parser at runtime, like we do for LUTs in #219, or to specify a separate iCE40 cell library similar to `fake_cell_library`.
* As support for these cells (as well as possibly cells for other FPGAs) are implemented, it might make sense to disentangle the cell name resolution/lookup code from the netlist parser code, i.e. separate the netlist parsing and the cell name lookup (which are both done as part of `xls::netlist::rtl::Parser::ParseNetlist`). In this way, we could have multiple cell name resolution "backends" that each know how to resolve cells for a specific vendor. The hardcoding of handling `SB_LUT4` could then be moved out of the parsing step.

PR #219, opened 2020-11-26
Add basic support for FPGA netlists in netlist parser

felixzhuologist:
This PR adds support for handling LUT4 based designs in the netlist parser. The approach diverges slightly from the one suggested in #142 - instead of figuring out the function based on the LUT mask, and then resolving the LUT4 to a known function from a predefined cell library, this PR takes the naive approach of just constructing a `CellLibraryEntry` "on the fly" by using the mask to specify a `StateTable`. I think this approach results in less code, but I'm not sure what the implications are for the later stages of LEC (it seems like state tables are translated to an if-else chain in Z3, but I'm not sure if that's slower to use than the built in functions) . Since the parsed netlist gets passed directly to the Z3 translator code, I don't think there are other downsides to not resolving the masks to known functions besides the stats output possibly being wrong because we don't identify LUTs that correspond to e.g. XORs (they'd count as "other" cells instead).

Some major caveats:

- I only looked at the LUT netlist format for iCE FPGAs, so I'm not sure if things differ for other vendors. The code is hardcoded to support this case.
  - Part of the reason for this is that I found it hard to find details about the expected .gv format. For iCE, I couldn't find any details about it in the data sheet and ended up just feeding small examples to `synth_ice40` in yosys to figure things out.
- It's still not possible to do LEC on useful examples since it's missing support for certain non LUT cells which I'm assuming are necessary for any multistage programs (e.g. [`SB_DFF`](https://github.com/YosysHQ/yosys/blob/master/techlibs/ice40/cells_sim.v#L249), which shows up when trying to get the [multistage LEC test](https://github.com/google/xls/blob/main/xls/solvers/z3_lec_test.cc#L319) working). I think handling this would involve creating a cell library for these cells, but I haven't looked into it further

I'm hoping this PR is still useful enough to check in. I've kept commits separate to make it easier to read but if this ends up being reviewed I'll squash before submitting.


xls/netlist/cell_library.cc
```diff
@@ -236,10 +283,9 @@ absl::StatusOr<bool> StateTable::MatchRow(
 absl::StatusOr<bool> StateTable::GetSignalValue(
     const InputStimulus& input_stimulus, absl::string_view signal) const {
   // Find a row matching the stimulus or return error.
-  XLS_RET_CHECK(std::find(proto_.internal_names().begin(),
-                          proto_.internal_names().end(),
-                          signal) != proto_.internal_names().end())
-      << "Can't find internal signal \"" << signal << "\"";
+  if (internal_names_.find(signal) == internal_names_.end()) {
```
meheff, line 286: nit: !internal_names_.contains(signal)

xls/netlist/cell_library_test.cc
```diff
@@ -138,6 +139,29 @@ TEST(CellLibraryTest, EvaluateStateTable) {
...
+
+TEST(CellLibraryTest, LutStateTable) {
+  // 4-way AND
+  StateTable table = StateTable::FromLutMask(0x8000);
+  StateTable::InputStimulus stimulus;
+  stimulus["I0"] = true;
+  stimulus["I1"] = true;
+  stimulus["I2"] = true;
+  stimulus["I3"] = true;
+  XLS_ASSERT_OK_AND_ASSIGN(bool signal, table.GetSignalValue(stimulus, "X"));
```
meheff, line 155: You can do:

EXPECT_THAT(table.GetSignalValue(stimulus, "X"), IsOkAndHolds(true));

same below.

xls/netlist/netlist.cc
```diff
@@ -214,6 +214,28 @@ absl::StatusOr<const Module*> Netlist::GetModule(
       absl::StrFormat("Module %s not found in netlist.", module_name));
 }
 
+absl::StatusOr<const CellLibraryEntry*> Netlist::GetOrCreateLut4CellEntry(
+    int64 lut_mask) {
+  if (lut_mask > ((1 << 16) - 1)) {
```
meheff, line 219: Probably want to check that the lut_mask is not negative either.

xls/netlist/netlist_parser_test.cc
```diff
@@ -176,6 +201,46 @@ endmodule)";
...
+  );
+endmodule
+)";
+  Scanner scanner(netlist);
+  XLS_ASSERT_OK_AND_ASSIGN(CellLibrary cell_library, MakeFakeCellLibrary());
+  XLS_ASSERT_OK_AND_ASSIGN(std::unique_ptr<Netlist> n,
+                           Parser::ParseNetlist(&cell_library, &scanner));
+  XLS_ASSERT_OK_AND_ASSIGN(const Module* m, n->GetModule("main"));
+  EXPECT_EQ("main", m->name());
+  XLS_ASSERT_OK(m->ResolveCell("q0_SB_LUT4_O"));
```
meheff, line 241: Anyway to check the 0x8000 value?

xls/netlist/parse_netlist_main_test.sh
```diff
@@ -14,11 +14,16 @@
...
 echo 'entries: { kind: INVERTER, name: "INV" input_names: "A" output_pin_list { pins { name: "ZN" function: "F" } } }' > "${TEST_TMPDIR}/fake_cell_library.textproto"
 echo 'module main(i, o); input i; output o; INV inv_0(.A(i), .ZN(o)); endmodule' > "${TEST_TMPDIR}/netlist.v"
 
 BINPATH=./xls/netlist/parse_netlist_main
 $BINPATH "${TEST_TMPDIR}/netlist.v" "${TEST_TMPDIR}/fake_cell_library.textproto" || exit -1
 
+# Test without a cell library
+echo 'module main(a0, a1, a2, a3, q0); input a0, a1, a2, a3; output q0; SB_LUT4 #(.LUT_INIT(16'"'"'h8000)) q0_lut (.I0(a0), .I1(a1), .I2(a2), .I3(a3), .O(q0)); endmodule' > "${TEST_TMPDIR}/netlist2.v"
+
+$BINPATH "${TEST_TMPDIR}/netlist2.v" || exit -1
```
meheff, line 27: if you make the first line:

#!/bin/bash -e

Then the script will fail if any of the commands inside fail. That way you don't need to add the "|| -1" which is easy to forget.

meheff (changes requested):
Looks great! Just some minor comments.

xls/netlist/cell_library.cc
felixzhuologist: Done

xls/netlist/cell_library_test.cc
felixzhuologist: fancy!

xls/netlist/netlist.cc
felixzhuologist: done

xls/netlist/netlist_parser_test.cc
felixzhuologist: The mask value gets discarded during parsing. I added a check for the cell library entry name (which contains the mask). A more thorough check would be to compare the `StateTable` against an expected value but this would require defining `StateTable::operator==`

xls/netlist/parse_netlist_main_test.sh
felixzhuologist: done

PR #233, opened 2020-12-15
"Added helper functions to 1) detect when 2 nodes are mapping targets…

jbaileyhandle:
… of a common source function 2) check if all of a node's operands have a mapping and 3) for an integration node, set as the return value a tuple of the mappings of the source function return values. Added a basic integration algorithm. Split up ir_integration into multiple files / build targets, moving the IntegrationBuilder and IntegrationOptions into their own files (This was necessary at some point.  Thought doing the split in the this PR would make review easier but in retrospect... probably not).  Note that we have a virtual IntegrationAlgorithm class. This class is a template because one of its functions must be able to create an object of a child class. Because the class is a template, it exists only as a header rather than an independent build target (actually, we have a .h and an *_implementation.h that mimics what we would put in a .cc file if the class was its own build target)."

### DIFF
--- a/xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
+++ b/xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
@@ -1,0 +1,100 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+#include "xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h"
+
+#include "xls/ir/node_iterator.h"
+
+namespace xls {
+
+void BasicIntegrationAlgorithm::EnqueueNodeIfReady(Node* node) {
+  if (!queued_nodes_.contains(node) &&
+      integration_function_->AllOperandsHaveMapping(node)) {
+    ready_nodes_.push_back(node);
+    queued_nodes_.insert(node);
+  }
+}
+
+absl::Status BasicIntegrationAlgorithm::Initialize() {
+  // Make integration function.
+  XLS_ASSIGN_OR_RETURN(integration_function_, NewIntegrationFunction());
+
+  // ID initial nodes with all operands ready.
+  for (const Function* func : source_functions_) {
+    for (Node* node: func->nodes()) {
+      if(node->op() != Op::kParam) {
+        EnqueueNodeIfReady(node);
+      }
+    }
+  }
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::unique_ptr<IntegrationFunction>>
+BasicIntegrationAlgorithm::Run() {
+  while (!ready_nodes_.empty()) {
+    std::optional<BasicIntegrationMove> move;
+    for (auto node_itr = ready_nodes_.begin(); node_itr != ready_nodes_.end();
+         ++node_itr) {
+      // Check insertion cost.
+      XLS_ASSIGN_OR_RETURN(int64 insert_cost,
+                           integration_function_->GetInsertNodeCost(*node_itr));
+      if (!move.has_value() || insert_cost < move.value().cost) {
+        move = MakeInsertMove(node_itr, insert_cost);
+      }
+
+      // Check merge cost.
+      for (Node* internal_node : integration_function_->function()->nodes()) {
+        // TODO(jbaileyhandle): Relax this requirement so that
+        // it only applies to integration-generated muxes.
+        if (!integration_function_->IsMappingTarget(internal_node)) {
+          continue;
+        }
+
+        // Check if mergeable
+        XLS_ASSIGN_OR_RETURN(
+            std::optional<int64> merge_cost,
+            integration_function_->GetMergeNodesCost(*node_itr, internal_node));
+        if (!merge_cost.has_value()) {
+          continue;
+        }
+
+        // Check if lowest cost.
+        XLS_RET_CHECK(move.has_value());
+        if (merge_cost < move.value().cost) {
+          move = MakeMergeMove(node_itr, internal_node, merge_cost.value());
+        }
+      }
+    }
+
+    // Execute lowest-cost move.
+    XLS_RET_CHECK(move.has_value());
+    XLS_RETURN_IF_ERROR(ExecuteMove(integration_function_.get(), move.value()).status());
+
+    // Update ready_nodes_.
+    ready_nodes_.erase(move.value().node_itr);
+    for (Node* user : move.value().node->users()) {
+      EnqueueNodeIfReady(user);
+    }
+  }
+
+  // Finalize.
+  XLS_RETURN_IF_ERROR(integration_function_->MakeTupleReturnValue().status());
+  return std::move(integration_function_);
+}
+
+// Explicit instantiation.
+template class IntegrationAlgorithm<BasicIntegrationAlgorithm>;
+
+}  // namespace xls
--- a/xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
+++ b/xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
@@ -1,0 +1,98 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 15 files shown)


jbaileyhandle:
Note:  ir_integrator.* will be renamed to integration_function.* in a later pull request.  I figured that if that file retained the same name for this pull request, the added helper functions would be easier to spot in the diff.

xls/contrib/integrator/ir_integrator.cc
```diff
@@ -166,6 +193,40 @@ IntegrationFunction::GetSourceFunctionIndexesOfNodesMappedToNode(
...
+absl::StatusOr<bool> IntegrationFunction::NodeSourceFunctionsCollide(
+    const Node* node_a, const Node* node_b) const {
+  // Get the index of the node's function if not owned by the integration
+  // function, otherwise get the source function indexes of the nodes mapped to
+  // the node.
+  auto get_node_source_indexes =
+      [this](const Node* node) -> absl::StatusOr<std::set<int64>> {
+    std::set<int64> sources;
+    if (IntegrationFunctionOwnsNode(node)) {
+      XLS_ASSIGN_OR_RETURN(sources,
```
meheff, line 205: nit:I think  you can just return the result of GetSourceFunctionIndexesOfNodesMappedtoNode.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace xls {
+
+// A naive merging algorithm.  A node is eligible to be added to the
+// integration function when all of its operands have already been added.
+// At each step, adds the eligible node for which the cost of adding it to
+// the function (either by inserting or merging with any integeration function
+// node) is the lowest.
+class BasicIntegrationAlgorithm
+    : public IntegrationAlgorithm<BasicIntegrationAlgorithm> {
```
meheff, line 29: Is CRTP necessary? Why doesn't normal inheritance work?

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
```diff
@@ -0,0 +1,104 @@
...
+
+absl::Status BasicIntegrationAlgorithm::Initialize() {
+  // Make integration function.
+  XLS_ASSIGN_OR_RETURN(integration_function_, NewIntegrationFunction());
+
+  // ID initial nodes with all operands ready.
+  for (const Function* func : source_functions_) {
+    for (const Node* param : func->params()) {
+      for (Node* user : param->users()) {
+        EnqueueNodeIfReady(user);
```
meheff, line 37: OOC, how does this deal with literals which have no operands? Do they get cloned over to the integration function?

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
```diff
@@ -0,0 +1,104 @@
...
+         ++node_itr) {
+      // Check insertion cost.
+      XLS_ASSIGN_OR_RETURN(int64 insert_cost,
+                           integration_function_->GetInsertNodeCost(*node_itr));
+      if (!move.has_value() || insert_cost < move.value().cost) {
+        move = MakeInsertMove(node_itr, insert_cost);
+      }
+
+      // Check merge cost.
+      for (Node* internal_node : TopoSort(integration_function_->function())) {
```
meheff, line 58: Do you need a toposort order here? function->nodes() is a lot faster.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+
+    // Execute lowest-cost move.
+    XLS_RET_CHECK(move.has_value());
+    auto result = ExecuteMove(integration_function_.get(), move.value());
+    if (!result.ok()) {
+      return result.status();
+    }
+    std::vector<Node*> new_nodes = result.value();
```
meheff, line 87: used anywhere? If it is not going to be used you can do:

XLS_RETURN_IF_ERROR(ExecuteMove(...).status());

meheff (changes requested):

xls/contrib/integrator/ir_integrator.cc
jbaileyhandle: Done.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
jbaileyhandle: We'd like to call a single static function to run an algorithm (which is some child class of IntegrationAlgorithm) to completion and return the result.  This is in contrast to a 2-step process where we 1) construct an instance of an algorithm then 2) run the algorithm (more prone to misuse).  XLS uses a similar general strategy (without inheritance) elsewhere (e.g. code generators).  

The issue is that the static function that we call belongs to the parent IntegrationAlgorithm class, but must call the constructor of a child class algorithm.  Currently, I don't know another way of doing this other than CRTP.  However, CRTP is akward to deal with (and advantage of single call vs. build + run calls isn't huge), so I'm open to suggestions for other approaches.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
jbaileyhandle: What will happen is that the algorithm will fail - good catch!  Fixed.

Interestingly, I expect my code at this point to two mux together two literals from different functions even if the literals are the same.  Yet, in a test case I added, I get 1 literal in the integrated function without the mux.  Are there some smarts somewhere besides the optimization passes (which I'm reasonably confident are being run) that would make this happen?

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
jbaileyhandle: Not especially - can you confirm that the ordering of nodes() is deterministic from run to run for testing purposes?

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.cc
jbaileyhandle: Done.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
meheff: That's fine. Worth a comment describing the motivation here.

meheff (approved):

xls/contrib/integrator/integration_algorithms/integration_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+#include "xls/contrib/integrator/ir_integrator.h"
+#include "xls/ir/function.h"
+#include "xls/ir/package.h"
+
+namespace xls {
+
+// An abstract class representing an algorithm to merge 2 or more
+// xls ir Functions.
+template <class AlgorithmType>
+class IntegrationAlgorithm {
```
hongted, line 30: Should provide a virtual destructor.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // Make a BasicIntegrationMove for a merge.
+  inline BasicIntegrationMove MakeMergeMove(std::list<Node*>::iterator node_itr,
+                                            Node* merge_node, int64 cost) {
+    return BasicIntegrationMove{{.node = *node_itr,
+                                 .move_type = IntegrationMoveType::kMerge,
+                                 .merge_node = merge_node,
+                                 .cost = cost},
+                                .node_itr = node_itr};
```
hongted, line 65: Designated initializers are a C++20 extension, suggest not using for greater compatibility.

If used, then code should not mix designated and non-designated, nor nest designated initializers.

ex.
```
    return BasicIntegrationMove{{.node = *node_itr,
                                 .move_type = IntegrationMoveType::kInsert,
                                 .cost = cost},
                                 node_itr};
```

xls/contrib/integrator/integration_algorithms/integration_algorithm.h
jbaileyhandle: Done.

xls/contrib/integrator/integration_algorithms/basic_integration_algorithm.h
jbaileyhandle: Ack.

"Designated initializers are a C++20 extension, suggest not using for greater compatibility."
Designated initializers are already in use in multple parts of XLS

"If used, then code should not mix designated and non-designated, nor nest designated initializers."
I'm a little unclear on what specifically in the initializations this is refering to - I'm assuming this inner bracketed component?
`{.node = *node_itr,
  .move_type = IntegrationMoveType::kInsert,
  .cost = cost}`

Maybe the issue is that this looks like its own designated intializer, with the resulting struct being passed into the outer designated initializer without a designation  / label?
The designated initialzers here are creating BasicIntegrationMove structs.  This struct inherets from the IntegrationMove struct which has fields node, move_type, merge_node, and cost.  So here, we are actually assigning the node, move_type, merge_node, and cost fields of the BasicIntegrationMove struct with designations.  The extra brackets are necessary because these fields are inhereted.

hongted: Issue is the outer bracketed component.  The first is using a non-designated initialization, while the second is using the designated initialization of .node_itr.  Suggestion as commented above -- remove the designated initialization of the outer bracket.

jbaileyhandle: Done.

hongted (approved):

PR #284, opened 2021-01-28
Added a dslx implementation of a fast fourier transform (fft) based o…

jbaileyhandle:
…n the one from MachSuite (https://breagen.github.io/MachSuite/). Also added a 32-bit floating point subtr act module (trivial modification on top of adder call) and made fp32 add and mul 'pub' functions so they can be imported and used in other files.  Note that the machsuite BSD-3 license is included in the fft directory.


jbaileyhandle:
User experience note for floating-point modules: floating point is very verbose (eg. fpmul_2x32::fpmul_2x32(x,y)) and it's odd that the floating-point operations are in separate modules. It would be nice if they were all collected and we could do calls like fp:mul_2x32 and fp:add_2x32 etc.  Or better yet if we could overload operators (long term idea - not an immediate priority).

jbaileyhandle:
It seems that fft is too large after unrolling (1000s of nodes at least) for ir optimization to scale well - My desktop (16 GB ram) get bricked when ir optimization runs.  Current 'fix' is just to disable conversion of dslx to ir.

meheff:
I contacted the open source licensing team inside Google and they said the ported code should live in a third_party directory with the original license in the DSLX file. Specifically: google/xls/third_party/machsuite

Inside this directory should be a LICENSE file containing the original MachSuite license. Each file which is ported from machsuite source code should have the *MachSuite* header at the top, not the one we use in XLS. I'd guess the other files in this PR (not the ported ones) can live inside the xls/ directory and be treated just like other code we write. They can just refer to the files in the third_party/machsuite directory for testing purposes etc.

jbaileyhandle:
Thanks for talking to the open source people, Mark.  When we say that the ported files should go under the third_party directory and the non-ported ones are okay where they are, which files would count as ported vs non-ported?  I'd assume fft_reference.cc would count as ported and that the changes to our floating point modules would not (not at all derived from Mach Suite).  I'm fuzzy fft.x and the test data, which follow the MachSuite as a reference but is our own implementation.

meheff:
If in doubt, just put it in third_party (like fft.x). Is the test data a copy of what is in the machsuite? or something you generated on your own, If it's a copy, then put it in third_party.

jbaileyhandle:
Went ahead and moved everything to the third_party directory (I don't think there was anything besides the BUILD file that wasn't at least partly derivative).

xls/third_party/machsuite/fft/fft_reference.cc
```diff
@@ -0,0 +1,165 @@
...
+    if(line.size()) {
+      XLS_CHECK(absl::SimpleAtof(line, &data[data_idx]));
+      data_idx++;
+    }
+  }
+
+  return absl::OkStatus();
+}
+
+typedef union float_uint_union_t {
```
meheff, line 116: union float_unit_union {
...
};

xls/third_party/machsuite/fft/fft_reference.cc
```diff
@@ -0,0 +1,165 @@
...
+// 'size' number of entries.
+void DisplayData(float data[], std::string name, int64 size) {
+  bool display_hex = absl::GetFlag(FLAGS_hex);
+  std::cout << name << " = [" << std::endl;
+
+  for(int64 idx=0; idx < size; ++idx) {
+    if(display_hex) {
+      float_uint_union data_union;
+      data_union.f32 = data[idx];
+      std::cout << std::hex << std::showbase << data_union.u32 << "," << std::endl;
```
meheff, line 131: std::cout << absl::StreamFormat("0x%x,\n", data_union.u32);

meheff (changes requested):
Thanks for adding this!

xls/third_party/machsuite/fft/fft_reference.cc
jbaileyhandle: uint as in uint32_t.

xls/third_party/machsuite/fft/fft_reference.cc
jbaileyhandle: Done.

xls/third_party/machsuite/fft/fft_reference.cc
meheff: I meant change the form of the declaration not the name. That is:

union foo {
};

Not:

typedef union foo_t {
} foo;

I'll make the change as it's uploaded internally.

jbaileyhandle:
Is the auto-puller (don't know real name) getting hung up on something?

meheff:
fyi, we're just waiting on one last approval before this can land internally and mirrored in github. It had to be split into several different changes, so not sure how that will be reflected in this PR after it lands.

meheff:
This change has landed but had to be broken into multiple bits so it's not reflected in the PR state. This can be reverted/withdrawn or whatever one does on Gtihub.

PR #307, opened 2021-02-22
Constant loop body paramters converted to loop-internal constants

jbaileyhandle:
For constants used in a loop body but defined outside the loop, reproduce constant nodes in in the loop body instead of passing these nodes to the loop body function as arguments.  This preserves const-ness. Motivating use case was nested loops where the loop bound was defined by a parametric value. Previously, this would not compile because inner loop bounds would appear as non-const parameters of the outer loop body.


meheff (approved):
Sorry for the delay on this.

cdleary:
Thanks for the excellent fix!

PR #314, opened 2021-02-26
Fix the README

briansrls:
I couldn't find this package, I think it's just a typo.

Here's the package I found:  https://packages.ubuntu.com/focal/python-is-python3


cdleary:
Whoops, I typed this in by hand -- thanks for the fix Biran!

Issue #342, opened 2021-03-17
Assert op should be able to interpolate data operands


cdleary:
Can start out simple, but as an extension to the facilities in #319 we'll likely want something that can capture/interpolate data operands, filing this issue to track. From a @meheff doc that hasn't been turned to markdown quite yet:

> **Richer error messages**
> Optionally,  the assert operation (and potentially assume or fatal) could include an arbitrary number of additional operands whose values are included in a dynamically generated error message. In this case, the message string could be a format string indicating how the operand values should be presented. For example:
> 
> assert(tkn, cond,
>        message=”Foo is too big: 0x%x”,
>        data_operands=[foo])
> 
> In this case, if the failure condition is true a possible error message might be:
> 
> Foo is too big: 0xabcd
> 
> In addition, source line information correlated back to the original program source could be included in the message. Each XLS IR op already includes fields for storing source correlation information.
> 
> The addition of these data operands may affect optimizations by increasing the number of uses of these values. This can prevent dead code elimination and numerous optimizations which are only performed for values which have a single use. This might be mitigated by removing the data operands and the richer messaging it provides prior to code generation.
> 
> An additional use case for these data operands is for error reporting in fatal operations. In this case, the values could be sent to the error reporting along with static information such as error codes.

meheff:
Fixed by https://github.com/google/xls/commit/c166c67e13fe542dd273fb6f28048de35a500cdf

PR #446, opened 2021-06-15
Add basic string interpolation to assert op

felixzhuologist:
This commit adds a `data_operands` field to the Assert node,
and specifies a format string format for interpolating these
operands into the assert message. This format is implemented
in the IR interpreter.

(see #342)

### DIFF
--- a/docs_src/ir_semantics.md
+++ b/docs_src/ir_semantics.md
@@ -1151,6 +1151,7 @@ other side-effecting operations.
 ```
 result = assert(tkn, condition, message=<string>)
 result = assert(tkn, condition, message=<string>, label=<string>)
+result = assert(tkn, condition, message=<string>, data_operands=<operands>, label=<string>)
 ```
 
 **Types**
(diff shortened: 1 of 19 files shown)


cdleary:
Really nice start! Just jotting down a few things we talked about briefly in video chat:

* Could be interesting to separate interpolation string "parsing" (maybe it's indeed just scanning at this point) and interpolation into separate "format string to command sequence data structure" and "apply format command sequence data structure to these values" phases, that way we can also give early errors from IR parsing time if the parsing fails (vs deferring to a late error at IR interpreter time)
* We might want to mull the right place to put this in the directory structure, but I think this is fine for now as we figure out what we think is the right spot -- probably we'll be able to share implementation between the IR interpreter and the DSL interpreter so some common location like this is good

felixzhuologist:
Thanks for writing those down! Some other things I have in my notes:

- Related to the point about refactoring into a command sequence: we'd also like to move the knowledge about what the format specifiers are (e.g. `{:x}`, `{:b}`) into the general util code (i.e. embedded in the command sequence type, I think). The current code makes the util code agnostic about what is inside the `{}`, making the caller responsible for giving them meaning (by handling them in the handler callbacks).
- The current code also doesn't handle errors well, like if a malformed string is used or not enough arguments are provided. Right now it just gives up on interpolation and prints out the raw string, but I think Chris mentioned we'd want to do the "parsing to command sequence" stuff inside the IR validator so that any bad data can be handled there.

felixzhuologist:
This is ready for another look :)

nanavati:
This is pretty similar to how I've been proceeding with supporting trace_fmt! into the IR, with two main differences:
1. I focused on Verilog generation first, so I need to fill in the rest of the IR pieces (and thank you for providing a template I can start from for that).
2. I keep the format in its parsed form (std::vector<FormatStep>) in the AST and retain that in the IR. I think asserts (and any other constructs we add in the future that support formatted output) should do the same thing.

So I'm not sure what the right path forward with this PR is. I want to say change over asserts to use parsed formats as well, but it doesn't make sense to do that until after I've committed the utilities I'm writing to deal with that for traces.

What do you think?

felixzhuologist:
> This is pretty similar to how I've been proceeding with supporting trace_fmt! into the IR, with two main differences:
> 
> 1. I focused on Verilog generation first, so I need to fill in the rest of the IR pieces (and thank you for providing a template I can start from for that).
> 2. I keep the format in its parsed form (std::vector) in the AST and retain that in the IR. I think asserts (and any other constructs we add in the future that support formatted output) should do the same thing.
> 
> So I'm not sure what the right path forward with this PR is. I want to say change over asserts to use parsed formats as well, but it doesn't make sense to do that until after I've committed the utilities I'm writing to deal with that for traces.
> 
> What do you think?

By keeping the format in its parsed form, do you mean parsing the message into the `std::vector<FormatStep>` early and then storing that in the IR instead of the raw message string? That makes sense to me, and I think would simplify things here slightly since we technically do the parsing twice - once during in the IR verifier, and then once at runtime during interpretation (mainly I just wasn't sure how best to do this/structure it for this PR). I'd be happy to wait until some of the changes for trace_fmt! are fleshed out some more (and incorporating things from this CL as needed) before continuing with these changes.

Feel free to let me know on this PR when you think it's a good time to revisit this

PR #468, opened 2021-07-29
[examples] add hsv2rgb

proppy:
Fast HSV to RGB conversion.

Ported from https://www.vagrearg.org/content/hsvrgb (MIT license).

- Translated https://www.vagrearg.org/hsvrgb/fast_hsv2rgb_32bit.chttps://www.vagrearg.org/hsvrgb/fast_hsv2rgb_32bit.c to unsafe Rust.
- Converted pointers swap to a match expression.
- Converted `mut`s and statements to let expressions.
- Ported to DSLX.

### DIFF
--- a/third_party/xls_colors/BUILD
+++ b/third_party/xls_colors/BUILD
@@ -23,3 +23,31 @@ xls_dslx_test(
     name = "hsv2rgb_dslx_test",
     dep = ":hsv2rgb_dslx_module",
 )
+
+cc_test(
+    name = "hsv2rgb_verilog_test",
+    srcs = ["test.cc"],
+    deps = [
+        "//xls/codegen:combinational_generator",
+        "//xls/common:xls_gunit_main",
+        "//xls/common/file:filesystem",
+        "//xls/common/file:get_runfile_path",
+        "//xls/common/status:matchers",
+        "//xls/dslx:ir_converter",
+        "//xls/dslx:parse_and_typecheck",
+        "//xls/ir",
+        "//xls/ir:ir_parser",
+        "//xls/ir:ir_test_base",
+        "@com_google_googletest//:gtest",
+        ":fast_hsv2rgb",
+    ],
+    data = [
+        ":hsv2rgb_dslx_module",
+    ]
+)
+
+cc_library(
+    name = "fast_hsv2rgb",
+    srcs = ["fast_hsv2rgb_32bit.c"],
+    hdrs = ["fast_hsv2rgb.h"],
+)
--- a/third_party/xls_colors/BUILD
+++ b/third_party/xls_colors/BUILD
@@ -1,0 +1,25 @@
+# XLS color library ported from https://www.vagrearg.org/content/hsvrgb
+
+load(
+    "//xls/build_rules:xls_build_defs.bzl",
+    "xls_dslx_module_library",
+    "xls_dslx_test",
+)
+
+package(
+    default_visibility = ["//xls:xls_internal"],
+)
+
+licenses(["notice"])
+
+exports_files(["LICENSE"])
+
+xls_dslx_module_library(
+    name = "hsv2rgb_dslx_module",
+    src = "hsv2rgb.x",
+)
+
+xls_dslx_test(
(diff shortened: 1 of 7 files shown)


xls/examples/hsv2rgb.x
```diff
@@ -0,0 +1,66 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Fast HSV to RGB conversion.
+//
+// Range:
+//   h: [0, 0x5ff], s: [0, 0xff], v: [0, 0xff]
+//   r: [0,  0xff], g: [0, 0xff], b: [0, 0xff]
+//
+// Ported from https://www.vagrearg.org/content/hsvrgb (MIT license).
```
hongted, line 21: Given that this was derived from code licensed by MIT -- I think we need a copy of the MIT license here rather than just a web link.

xls/examples/hsv2rgb.x
```diff
@@ -0,0 +1,66 @@
...
+
+pub fn slope(h_fraction: u16, s: u8, v:u8) -> u8{
+  let u':u32 = (v as u32) * ((u32:255 << 8) - ((s as u16) * h_fraction) as u32);
+  let u'' = u' + (u' >> 8);
+  let u''' = u'' + (v as u32);
```
hongted, line 26: If u' is never used again, it doesn't hurt to just use "let u = ".

xls/examples/hsv2rgb.x
```diff
@@ -0,0 +1,66 @@
...
+  (u''' >> 16)[0+:u8]
+}
+
+pub fn hsv2rgb(h: u16, s: u8, v: u8) -> (u8, u8, u8) {
+  let sextant_:u8 = (h >> 8)[0+:u8];
```
hongted, line 31: Naming convention is that _ is a name to bind to but not read (https://google.github.io/xls/dslx_reference/#identifiers)

hongted, line 31: Can this be replaced with just a bit slice? h[8+=:u8]?

xls/examples/hsv2rgb.x
```diff
@@ -0,0 +1,66 @@
...
+  let sextant = u8:5 if sextant_ > u8:5 else sextant_;
+
+  let ww':u16 = (v as u16) * ((u8:255 - s) as u16);
+  let ww'' = ww' + u16:1;
+  let ww = ww'' + ww'' >> 8;
+  let c = (ww >> 8)[0+:u8];
+
+  let h_fraction:u16 = h[0+:u8] as u16;
```
hongted, line 39: Think it's a bit clearer to use h[0:7] rather than +:

xls/examples/hsv2rgb.x
hongted: Or as later mentioned, use explicit bit slices.

cdleary:
By current policy this needs to live in a spot analogous to xls_go_math, since the original source was MIT licensed before porting to the DSL and it will carry the same MIT license in that new spot.

proppy:
> By current policy this needs to live in a spot analogous to xls_go_math, since the original source was MIT licensed before porting to the DSL and it will carry the same MIT license in that new spot.

`xls_contrib`? :)

cdleary:
> > By current policy this needs to live in a spot analogous to xls_go_math, since the original source was MIT licensed before porting to the DSL and it will carry the same MIT license in that new spot.
> 
> `xls_contrib`? :)

Sorry was on mobile so was hard to look up details -- https://github.com/google/xls/tree/main/third_party subdirectory as xls_hsvrgb would be the current protocol, and see how the `LICENSE`/`BUILD` files are done in those subdirs.

proppy:
> subdirectory as xls_hsvrgb would be the current protocol

added third_party/xls_colors so that there is room to add more helpers, let me know what you think?

proppy:
Added simulation based tests along side the original C++ implementation, PTAL.

cdleary:
CL is being internally reviewed actively now, sorry for the delay.

PR #478, opened 2021-08-09
[tools/codegen_main] accept stdin as positional argument

proppy:


hongted (approved):
LGTM.  Thanks for the PR.

PR #482, opened 2021-08-12
[public]: add python bindings

proppy:
This will allow better integration with other python based EDA tools (like `litex`, `migen` and `nmigen`).

/cc @tcal-x who was interesting into discussing ways to ease XLS binary distribution.

### DIFF
--- a/xls/public/python/runtime_build_actions.cc
+++ b/xls/public/python/runtime_build_actions.cc
@@ -27,23 +27,18 @@ PYBIND11_MODULE(runtime_build_actions, m) {
 
   m.def("convert_dslx_to_ir", [](absl::string_view dslx, absl::string_view path, absl::string_view package, const std::vector<absl::string_view>& additional_search_paths) -> absl::StatusOr<std::string> {
       return ConvertDslxToIr(dslx, path, package, std::vector<std::filesystem::path>(additional_search_paths.begin(), additional_search_paths.end()));
-    });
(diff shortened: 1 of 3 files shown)


xls/public/python/runtime_build_actions.cc
```diff
@@ -0,0 +1,49 @@
...
+#include "xls/common/status/statusor_pybind_caster.h"
+
+namespace py = pybind11;
+
+namespace xls {
+
+PYBIND11_MODULE(runtime_build_actions, m) {
+  ImportStatusModule();
+
+  m.def("convert_dslx_to_ir", [](absl::string_view dslx, absl::string_view path, absl::string_view package, const std::vector<absl::string_view>& additional_search_paths) -> absl::StatusOr<std::string> {
```
proppy, line 28:
```
>>> import runtime_build_actions
>>> help(runtime_build_actions)
Help on module runtime_build_actions:

NAME
    runtime_build_actions

FUNCTIONS
    convert_dslx_path_to_ir(...) method of builtins.PyCapsule instance
        convert_dslx_path_to_ir(arg0: str, arg1: List[str]) -> StatusOr[str]
```

We should add `py::arg` arguments to get better argument names.

xls/public/python/runtime_build_actions.cc
```diff
@@ -0,0 +1,44 @@
...
```
proppy, line 28: we should also add some docstrings.

PR #484, opened 2021-08-18
[public] add ConvertIrToCombinationalVerilog

proppy:
- add ConvertIrToCombinationalVerilog build action
- add python bindings
- use xls.common.testbase

### DIFF
--- a/xls/public/runtime_build_actions.h
+++ b/xls/public/runtime_build_actions.h
@@ -86,6 +86,20 @@ absl::StatusOr<std::string> ProtoToDslx(absl::string_view proto_def,
                                         absl::string_view text_proto,
                                         absl::string_view binding_name);
 
+
+// Converts the specified XLS IR text into a combinational Verilog module.
+//
+// Args:
+//  ir: Text for the XLS IR package.
+//  path: Path to use for source location information for the given text.
+//    Since text may be generated, an empty string or a pseudo path like
+//    "<generated>" is acceptable.
+//  entry: Entry function for the XLS IR package,
+//    if empty entry function of the package is used.
+//  module_name: Name of the Verilog module,
+//    if empty use the mangled IR function name.
+absl::StatusOr<std::string> ConvertIrToCombinationalVerilog(absl::string_view ir, absl::string_view path, absl::string_view entry, absl::string_view module_name);
+
 }  // namespace xls
 
 #endif  // XLS_PUBLIC_RUNTIME_BUILD_ACTIONS_H_
--- a/xls/public/BUILD
+++ b/xls/public/BUILD
@@ -30,10 +30,12 @@ cc_library(
     deps = [
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
+        "//xls/codegen:combinational_generator",
         "//xls/dslx:import_data",
         "//xls/dslx:ir_converter",
         "//xls/dslx:mangle",
         "//xls/dslx:parse_and_typecheck",
+        "//xls/ir:ir_parser",
         "//xls/passes",
         "//xls/tools:opt",
         "//xls/tools:proto_to_dslx",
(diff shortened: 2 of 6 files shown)


xls/public/runtime_build_actions.h
```diff
@@ -86,6 +86,20 @@ absl::StatusOr<std::string> ProtoToDslx(absl::string_view proto_def,
...
+// Args:
+//  ir: Text for the XLS IR package.
+//  path: Path to use for source location information for the given text.
+//    Since text may be generated, an empty string or a pseudo path like
+//    "<generated>" is acceptable.
+//  entry: Entry function for the XLS IR package,
+//    if empty entry function of the package is used.
+//  module_name: Name of the Verilog module,
+//    if empty use the mangled IR function name.
+absl::StatusOr<std::string> ConvertIrToCombinationalVerilog(absl::string_view ir, absl::string_view path, absl::string_view entry, absl::string_view module_name);
```
proppy, line 101: should we make combinational vs pipeline different entrypoint or a separate argument? (I opted for the former because the arguments needed a significantly differents)

xls/public/runtime_build_actions.h
```diff
@@ -86,6 +86,20 @@ absl::StatusOr<std::string> ProtoToDslx(absl::string_view proto_def,
...
+//
+// Args:
+//  ir: Text for the XLS IR package.
+//  path: Path to use for source location information for the given text.
+//    Since text may be generated, an empty string or a pseudo path like
+//    "<generated>" is acceptable.
+//  entry: Entry function for the XLS IR package,
+//    if empty entry function of the package is used.
+//  module_name: Name of the Verilog module,
+//    if empty use the mangled IR function name.
```
proppy, line 100: should we always enforce public function parameter to be defined? or are we cool with the empty value holding some semantics?

xls/public/runtime_build_actions.h
```diff
@@ -86,6 +86,20 @@ absl::StatusOr<std::string> ProtoToDslx(absl::string_view proto_def,
...
+
+// Converts the specified XLS IR text into a combinational Verilog module.
+//
+// Args:
+//  ir: Text for the XLS IR package.
+//  path: Path to use for source location information for the given text.
+//    Since text may be generated, an empty string or a pseudo path like
+//    "<generated>" is acceptable.
+//  entry: Entry function for the XLS IR package,
+//    if empty entry function of the package is used.
```
proppy, line 98: should we always enforce public function parameter to be defined? or are we cool with the empty value holding some semantics?

xls/public/runtime_build_actions.h
hongted: In think the end-goal should be that there is a single entry point and that whether or not it's combinational/procedural; the entry point; and any generation constraints (ex. stages, input/output flops, signal names, inverting/non-inverting reset, etc...) can be configured.

That said, once we get into generating IR for a set of cooperating procs the needs would become a bit more concrete.

meheff (commented):
I entered these comments a long while back but now realize I didn't not actually send them out, sigh. Sorry about the delay :-(

xls/public/runtime_build_actions.h
proppy: I like really the idea of defining the public interface payloads thru proto, should we go the extra-mile and include a [`service` definition](https://developers.google.com/protocol-buffers/docs/proto3#services)?

I think this could help with:
- tracking changes made to the API and versioning
- help with binding/parameters marshaling for various language (could pass paramater as raw byte and decode using code generated from proto)
- expose the toolchain as a service

WDYT?

cdleary:
Going through old-PR backlog, what was the disposition on this @proppy ?

cdleary:
Closing due to staleness.

proppy:
@cdleary yes, thanks for triaging: we'll likely revisit as part of https://github.com/google/xls/issues/1256

PR #493, opened 2021-09-13
[examples] Integrating bazel rules hdl synthesis.

QuantamHD:
This integrates the synthesis rules into the XLS repo with the IDCT jpeg example.


QuantamHD:
@nanavati

QuantamHD:
@cdleary

hongted:
How are eda tool dependencies handled?  Are they assumed to have been pre-built and located somewhere in the system path?  Should the readme be updated to include any needed changes for these targets?

QuantamHD:
They are built from source in as part of bazel build similar to google3. Every tool used in the flow is built hermetically as part of the build. https://github.com/hdl/bazel_rules_hdl/tree/main/dependency_support

QuantamHD:
@hongted If there aren't any concerns with this would you like to merge?

nanavati (approved):
Looks good to me. Thanks for integrating this!

PR #518, opened 2021-11-09
Bump up BoringSSL commit

dnmiller:
The listed commit for the boringssl dependency points to a version that doesn't build with GCC 11.1, which means that XLS doesn't build with GCC 11.1. The underlying issue was fixed in boringssl commit [92c6fbfc4c44dc8462d260d836020d2b793e7804](https://boringssl.googlesource.com/boringssl/+/92c6fbfc4c44dc8462d260d836020d2b793e7804) and is currently in boringssl `main`. This updates the commit to `main-with-bazel` of boringssl as of 2021-11-01 and allows XLS to build with GCC 11.1.

### DIFF
--- a/dependency_support/load_external.bzl
+++ b/dependency_support/load_external.bzl
@@ -59,9 +59,9 @@ def load_external_repositories():
 
     git_repository(
         name = "boringssl",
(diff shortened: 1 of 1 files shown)


dnmiller:
@googlebot I signed it!

PR #533, opened 2021-12-08
Update bazel-toolchain to 0.6.3

lachlansneff:
This updates the bazel-toolchain dependency to a pinned version as well as removes the google-workstation-workaround patch (I think it should work without it now that the system detection logic has changed, would be great if someone would double-check).


cdleary:
Thanks for the contribution, looks great! Trying to get this merged today, just need to fix what looks like some fallout from new bazel version first.

cdleary:
The new OS detection mechanism worked like a charm.

Just had to update the command in the comment per the repo's current README guidance:

> We no longer support the --crosstool_top=@llvm_toolchain//:toolchain flag, and instead rely on the --incompatible_enable_cc_toolchain_resolution flag.
-- https://github.com/grailbio/bazel-toolchain

See also issue #558

PR #599, opened 2022-03-28
Necessary-but-not-sufficient changes to successfully build on OSX.

RobSpringer:
Found while attempting to build on an M1 Mac. Tested in that environment, but not on Linux and/or x86.

I don't love the `#ifdef LINUX`, but I wasn't able to find another way to specialize that the  non-POSIX `c_line` field. I'd be just as happy to drop it. WDYT?

### DIFF
--- a/xls/ir/caret.cc
+++ b/xls/ir/caret.cc
@@ -95,7 +95,9 @@ std::string PrintCaret(
     int64_t slack =
         terminal_width - (line_contents_owned.size() + line_number_width + 3);
     int64_t size =
-        std::max(0l, static_cast<int64_t>(line_contents_owned.size()) + slack);
+        std::max(
+          static_cast<int64_t>(0),
+          static_cast<int64_t>(line_contents_owned.size()) + slack);
     line_contents_owned = line_contents_owned.substr(0, size - 1);
     absl::StrAppend(&line_contents_owned, "…");
   }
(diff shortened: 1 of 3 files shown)


xls/ir/caret.cc
```diff
@@ -95,7 +95,9 @@ std::string PrintCaret(
     int64_t slack =
         terminal_width - (line_contents_owned.size() + line_number_width + 3);
     int64_t size =
-        std::max(0l, static_cast<int64_t>(line_contents_owned.size()) + slack);
+        std::max(
+          static_cast<int64_t>(0),
```
taktoa, line 99: What's the difference between `0l` and `static_cast<int64_t>(0)`?

taktoa (changes requested):

xls/ir/caret.cc
RobSpringer: The exact definition of `long` is architecture-specific; for example, in ILP64, `int`s, `long`s, and pointers are 64 bits wide, whereas in LLP64, a `long` is 32 bits, whereas `long long`s and pointers are 64 bits. IIRC, OSX is LP64, which _should_ be fine here, but either A) I'm wrong, or B) the compiler is very particular.

taktoa (approved):

cdleary (approved):

Issue #615, opened 2022-05-10
integrating SDC-based pipeline scheduling method into XLS?


zihaoli-cn:
SDC-based scheduling [1] is an HLS scheduling method using Linear Programming(Not the ILP). Although using real-number variables, it is guaranteed to get integer solutions, because the constraint matrix is totally unimodular. It can be solved efficiently.

I read the code in the scheduling part in XLS. The objective is to minimize the pipeline registers. 

- As for the objective function, it is very similar to [2]'s 4.2 part, "Minimum-Lifetime Schedule over SDC". It introduces the life-time variables to the model. For a value **_v_** in XLS IR, bit-count(**_v_**) * life-time(**_v_**) is just its pipeline registers usage.
- As for the cycle period time constraints, section 3.2.2 in [1] uses "critical combinational path" to formally describe it. But I don't know how to implement it efficiently. It needs to compute critical paths between any pair of nodes.

I am interested in integrating the SDC-based scheduling method into XLS and comparing the result and running time with XLS's original algorithm. We can use [OR-Tools](https://developers.google.com/optimization) as the solver.

**[1]** Cong, Jason, and Zhiru Zhang. "An efficient and versatile scheduling algorithm based on SDC formulation." 2006 43rd ACM/IEEE Design Automation Conference. IEEE, 2006.
**[2]** Zhang, Zhiru, and Bin Liu. "SDC-based modulo scheduling for pipeline synthesis." 2013 IEEE/ACM International Conference on Computer-Aided Design (ICCAD). IEEE, 2013.

taktoa:
We have discussed using SDC in our scheduler, especially because we are interested in implementing exact cycle constraints on send/receive nodes for communication with fixed-latency IPs like SRAMs. The current scheduler optimizes for number of registers, number of pipeline stages, and critical combinational path length.

Optimizing for number of registers in the SDC formulation is easy; if the (minimized) objective is `c dot x` then for every edge in the dataflow graph betwen node `i` and node `j` with bitwidth `w`, you subtract `w` from `c[i]` and add `w` to `c[j]`, which is equivalent to adding `w * (x[j] - x[i])` to the objective function (and the edge itself induces a timing constraint such that `x[j] >= x[i]`).

Optimizing for number of pipeline stages is also easy. You add an extra variable `n` as well as constraints that for all `i`, `n - 
x[i] >= 0`. Then if `n` is part of the objective function, it will be the least upper bound of the stage numbers of all nodes, i.e.: the number of stages.

[...]

PR #619, opened 2022-05-11
add support for SDC scheduling

zihaoli-cn:


zihaoli-cn:
related [issue](https://github.com/google/xls/issues/615)

dependency_support/load_external.bzl
```diff
@@ -180,3 +180,57 @@ def load_external_repositories():
...
     )
+
+    git_repository(
+        name = "platforms",
+        tag = "0.0.5",
+        remote = "https://github.com/bazelbuild/platforms.git",
+    )
+
+    http_archive(
+        name = "glpk",
```
taktoa, line 191: GLPK is GPLv3 licensed, so I think we can't link it in without violating the license.

dependency_support/load_external.bzl
```diff
@@ -180,3 +180,57 @@ def load_external_repositories():
...
+    http_archive(
+        name = "bliss",
+        build_file = "@com_google_ortools//bazel:bliss.BUILD",
+        patches = ["@com_google_ortools//bazel:bliss-0.73.patch"],
+        sha256 = "f57bf32804140cad58b1240b804e0dbd68f7e6bf67eba8e0c0fa3a62fd7f0f84",
+        url = "http://www.tcs.hut.fi/Software/bliss/bliss-0.73.zip",
+    )
+
+    new_git_repository(
+        name = "scip",
```
taktoa, line 206: SCIP uses the ZIB Academic License, which is also not compatible with Apache 2.0

dependency_support/load_external.bzl
```diff
@@ -180,3 +180,57 @@ def load_external_repositories():
...
+
+    http_archive(
+        name = "glpk",
+        build_file = "//bazel:glpk.BUILD",
+        sha256 = "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15",
+        url = "http://ftp.gnu.org/gnu/glpk/glpk-5.0.tar.gz",
+    )
+
+    http_archive(
+        name = "bliss",
```
taktoa, line 198: BLISS is LGPL licensed, I'll need to look into whether or not it's compatible with our license (I believe it is, but Google has policies).

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+      result;  // critical combinational path's delay
+
+  for (Node* src : f->nodes()) {
+    for (auto [dst, ccp] : ComputeSingleSourceCCP(src, node_delay)) {
+      std::pair<Node*, Node*> path(src, dst);
+      result.push_back(std::make_pair(path, ccp));
+    }
+  }
+  return result;
+}
```
taktoa, line 217: I believe this is a reimplementation of [LongestNodePaths](https://github.com/google/xls/blob/main/xls/data_structures/graph_contraction.h#L264). Double check that the output is the same before replacing it.

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+    const DelayEstimator& delay_estimator, sched::ScheduleBounds* bounds,
+    int64_t clock_period_ps) {
+  XLS_VLOG(3) << "ScheduleToMinimizeRegistersSDC()";
+  XLS_VLOG(3) << "  pipeline stages = " << pipeline_stages;
+  XLS_VLOG_LINES(4, f->DumpIr());
+
+  XLS_VLOG(4) << "Initial bounds:";
+  XLS_VLOG_LINES(4, bounds->ToString());
+
+  std::unique_ptr<MPSolver> solver(MPSolver::CreateSolver("SCIP"));
```
taktoa, line 242: Use GLOP by default

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  if (!solver) {
+    XLS_LOG(WARNING) << "SCIP solver unavailable.";
+    return absl::UnavailableError("SCIP solver unavailable.");
+  }
+
+  const double infinity = solver->infinity();
+
+  absl::flat_hash_map<Node*, MPVariable*>
+      node_cycle_var;  // Node's cycle after scheduling
```
taktoa, line 251: Put comments on the line before the declaration.

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  absl::flat_hash_map<Node*, MPVariable*>
+      node_lifetime_var;  // Node's lifetime, since it finishes until it's
+                          // consumed by the last user
```
taktoa, line 254: Same here

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
   return cycle_map;
 }
 
+// A helper function to compute each node's delay by calling the delay estimator
+// The result is used by `ComputeCriticalCombinationalPaths`,
+// `ComputeSingleSourceCCP`, `ScheduleToMinimizeRegistersSDC`
+absl::StatusOr<absl::flat_hash_map<Node*, int64_t>> ComputeNodeDelay(
```
taktoa, line 150: `ComputeNodeDelays` would probably be a better name

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  for (Node* src : f->nodes()) {
+    for (auto [dst, ccp] : ComputeSingleSourceCCP(src, node_delay)) {
+      std::pair<Node*, Node*> path(src, dst);
+      result.push_back(std::make_pair(path, ccp));
+    }
+  }
+  return result;
+}
+
+using namespace operations_research;
```
taktoa, line 219: `using namespace` is against our style guide. I think `namespace or_tools = ::operations_research;` would be acceptable.

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  XLS_VLOG(3) << "ScheduleToMinimizeRegistersSDC()";
+  XLS_VLOG(3) << "  pipeline stages = " << pipeline_stages;
+  XLS_VLOG_LINES(4, f->DumpIr());
+
+  XLS_VLOG(4) << "Initial bounds:";
+  XLS_VLOG_LINES(4, bounds->ToString());
+
+  std::unique_ptr<MPSolver> solver(MPSolver::CreateSolver("SCIP"));
+  if (!solver) {
+    XLS_LOG(WARNING) << "SCIP solver unavailable.";
```
taktoa, line 244: No need to log as well as returning an error, just return the error

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  for (Node* node : f->nodes()) {
+    objective->SetCoefficient(node_lifetime_var[node],
+                              node->GetType()->GetFlatBitCount());
+  }
+  objective->SetMinimization();
+
+  const MPSolver::ResultStatus result_status = solver->Solve();
+  // Check that the problem has an optimal solution.
+  if (result_status != MPSolver::OPTIMAL) {
+    XLS_LOG(FATAL) << "The problem does not have an optimal solution!";
```
taktoa, line 300: Same here

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+    return absl::InternalError(
+        "The problem does not have an optimal solution!");
+  }
+
+  // extract result
+  ScheduleCycleMap cycle_map;
+  for (Node* node : f->nodes()) {
+    double cycle = node_cycle_var[node]->solution_value();
+    XLS_CHECK_LE(std::fabs(cycle - std::floor(cycle)), 1e-3)
```
taktoa, line 309: Shouldn't you round rather than floor here? Also, use `XLS_RET_CHECK_LE` or directly call absl::InternalError if that doesn't exist.

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+        0.0, infinity, absl::StrFormat("lifetime_%s", node->GetName()));
+  }
+
+  for (Node* node : f->nodes()) {
+    MPVariable* const var_node = node_cycle_var.at(node);
+    MPVariable* const var_node_lifetime = node_lifetime_var.at(node);
+    for (Node* user : node->users()) {
+      MPVariable* const var_user = node_cycle_var.at(user);
+
+      // Constraint: cycle[i_user] - cycle[i] <= lifetime[i]
```
taktoa, line 268: You also need a constraint `cycle[i_user] - cycle[i] >= 0`, otherwise causality can be violated. Note that this is not equivalent to the `lifetime[i] >= 0` constraint present on line 259.

xls/scheduling/pipeline_schedule.cc
```diff
@@ -141,6 +144,175 @@ absl::StatusOr<ScheduleCycleMap> ScheduleToMinimizeRegisters(
...
+  XLS_ASSIGN_OR_RETURN(auto node_delay, ComputeNodeDelay(f, delay_estimator));
+  for (auto [path, path_delay] :
+       ComputeCriticalCombinationalPaths(f, node_delay)) {
+    if (path_delay < clock_period_ps) {
+      continue;
+    }
+    auto [src, dst] = path;
+    // Constraint: cycle[src] - cycle[dst] <= -path_delay/clock_period_ps
+    MPConstraint* const c =
+        solver->MakeRowConstraint(-infinity, -path_delay / clock_period_ps);
```
taktoa, line 285: `path_delay` and `clock_period_ps` are `int64_t`, so this will compute `-floor(path_delay / clock_period_ps)`, which will include some invalid situations. For example, if the ratio is 5.7, then that means that `cycle[dst] - cycle[src] >= 6` is the right constraint to generate, but you are generating `cycle[src] - cycle[dst] <= -5`.

When I switch over to this formulation, I get "the problem does not have an optimal solution" for some of our internal benchmarks, so something else must be wrong as well. Without that, we get schedules that do not meet timing.

zihaoli-cn:
@taktoa Thanks a lot for your careful code review!

Here are some updates I made for this commit:
- fix the mentioned bugs/bad code styles, including:
  - put comments on the line before the declaration
  - rename `ComputeNodeDelay` to `ComputeNodeDelays`
  - using namespace `or_tools`
  - remove useless logs
  - use "round" rather than "floor"
  - add constraints for DU chains
- remove scip/glpk dependence
  - directly remove glpk's dependence will not bring any trouble. But remove scip's will cause linking error, owing to `linear_solver`'s dependence on `scip` in [ortools/linear_solver/BUILD.bazel:108](https://github.com/google/or-tools/blob/stable/ortools/linear_solver/BUILD.bazel). 
  - `MPSolver` depends on `scip` with the purpose of providing **configurability** between different solvers.
  - Therefore, I directly `glop`'s Linear Programming API rather the wrapped `MPSolver`'s. It can avoid scip's external dependence.

Some other issues:
- I do some experiments, and the `LongestNodePaths` doesn't produce the same result as `ComputeCriticalCombinationalPaths`.
- After these updates, I tested the SDC scheduling method myself by mimicking the way XLS team did the test in [xls/scheduling/function_partition_test.cc : BenchmarkTest](https://github.com/google/xls/blob/main/xls/scheduling/function_partition_test.cc). It can pass `adler32.opt` and `crc32.opt` with verified timing. But it will take a long time to pass `sha256.opt`'s test case, finally, the test would time out. `sha256.opt` has 4929 nodes in IR.
- In this modeling method, the number of constraints is polynomial times of variables' number. If the `glop` uses a `Simplex Algorithm`-based method. If there are many constraints and few variables, it will be more time-efficient to solve the dual problem.

zihaoli-cn:
@taktoa Thanks a lot for your careful code review!

Here are some updates I made for this commit:
- fix the mentioned bugs/bad code styles, including:
  - put comments on the line before the declaration
  - rename `ComputeNodeDelay` to `ComputeNodeDelays`
  - using namespace `or_tools`
  - remove useless logs
  - use "round" rather than "floor"
  - add constraints for DU chains
- remove scip/glpk dependence
  - directly remove glpk's dependence will not bring any trouble. But remove scip's will cause linking error, owing to `linear_solver`'s dependence on `scip` in [ortools/linear_solver/BUILD.bazel:108](https://github.com/google/or-tools/blob/stable/ortools/linear_solver/BUILD.bazel). 
  - `MPSolver` depends on `scip` with the purpose of providing **configurability** between different solvers.
  - Therefore, I directly `glop`'s Linear Programming API rather the wrapped `MPSolver`''s.  to avoid scip's external dependence.

Some other issues:
- I do some experiments, and the `LongestNodePaths` doesn't produce the same result as `ComputeCriticalCombinationalPaths`.
- After these updates, I tested the SDC scheduling method myself by mimicking the way XLS team did the test in [xls/scheduling/function_partition_test.cc : BenchmarkTest](https://github.com/google/xls/blob/main/xls/scheduling/function_partition_test.cc). It can pass `adler32.opt` and `crc32.opt` with verified timing. But it will take a long time to pass `sha256.opt`'s test case, finally, the test would time out. `sha256.opt` has 4929 nodes in IR.
- In this modeling method, the number of constraints is polynomial times of variables' number. If the `glop` uses a `Simplex Algorithm`-based method. If there are many constraints and few variables, it will be more time-efficient to solve the dual problem.

taktoa:
I got this working on all our internal/external benchmarks. Your DFS function was really slow and consumptive of memory, so I replaced it with `LongestNodePaths` and just kept the redundant `target - source >= 1` constraints where the critical path between `source` and `target` is not the smallest critical path that is greater than one clock period. The benchmark results show that the quality of results is pretty much the same as the old scheduler, and execution time was the same in all benchmarks except one, which wasn't degraded too much.

I'll do the remaining work to get this merged.

zihaoli-cn:
@taktoa Thank you for your continued attention to my commitment.

Yeah, the [022ef33](https://github.com/google/xls/pull/619/commits/022ef33077b83101dda4988caa3df901f1eb5bfd) commit can give scheduling result correctly. 

According to the results of my experiments, the min-cut-based scheduling method is a well-designed algorithm, which has a high approximation ratio. 

I also want to ask you a question. **What is the best practice to deal with duplicated flags in the external dependent project?**

- The or-tools have a lot of duplicate flags, such as `vmodule`, `log_prefix`, which are already declared in XLS.

PR #726, opened 2022-09-28
bazelrc: enable llvm toolchain

### LINKED ISSUE #558: Enable build with LLVM toolchain provider via grailbio/bazel-toolchain
Historically m4 had difficulty building with this enabled, now it looks like the immediate issue is with the `dependency_support:embedded_python_interpreter` target e.g. running the following wildcard:

```console
$ bazel test -c opt  --incompatible_enable_cc_toolchain_resolution ...
```

If we want to we could start trying to build up CI for targets that are known to work, but honestly running vs the standard CC toolchain on select OS-es has been working ok for development to date, so marking as "long term enhancement", just know that we're not winning any "hermiticity of the year" awards for our builds.

proppy:
From https://github.com/hdl/conda-eda/blob/master/hls/xls/use-llvm-toolchain.patch

Fixes #558 #141

@cdleary is it fine to enable by default?

### DIFF
--- a/dependency_support/rules_hdl/workspace.bzl
+++ b/dependency_support/rules_hdl/workspace.bzl
@@ -30,8 +30,8 @@ def repo():
     )
 
     # Commit on 2023-06-13, current as of 2023-06-13.
-    git_hash = "e6540a5bccbfb124aec0b19deaa9cf855781b3a5"
-    git_sha256 = "21307b0c14a036f1b4879c8f1d4d50a115053eb87c428307d4d6569c3e7ba859"
+    git_hash = "723965fc52f7a3636b743e12f06047879a7251e0"
+    git_sha256 = "56904adddd2059df17f4113534f4a4a30964948ceeabf86a1a390d068c4ebef0"
 
     maybe(
         http_archive,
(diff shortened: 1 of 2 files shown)


cdleary:
Should be fine to enable by default given that it works in CI.

Unfortunately since we changed Bazel build flags it'll need to rebuild everything which is gonna take a while for us to observe the result, may time out the first go around, but running it locally too and if that looks ok we'll put it in and see how it goes.

proppy:
> Unfortunately since we changed Bazel build flags it'll need to rebuild everything which is gonna take a while for us to observe the result

That shouldn't be a concern anymore with the dedicated runners.

dependency_support/rules_hdl/workspace.bzl
```diff
@@ -30,15 +30,15 @@ def repo():
...
+    git_sha256 = "56904adddd2059df17f4113534f4a4a30964948ceeabf86a1a390d068c4ebef0"
 
     maybe(
         http_archive,
         name = "rules_hdl",
         sha256 = git_sha256,
         strip_prefix = "bazel_rules_hdl-%s" % git_hash,
         urls = [
-            "https://github.com/hdl/bazel_rules_hdl/archive/%s.tar.gz" % git_hash,
+            "https://github.com/proppy/bazel_rules_hdl/archive/%s.tar.gz" % git_hash,
```
proppy, line 42: trying https://github.com/hdl/bazel_rules_hdl/pull/184 out, and see if it fixes the CI; will update once merged.

proppy:
Superseeded by https://github.com/google/xls/commit/72b254fddf8f356420fc0c9831dd7c461d41f37d.

PR #807, opened 2022-12-14
delay_model/sky130: alias onehot to or

### LINKED ISSUE #806: one_hot not supported by delay_model for sky130
When trying to codegen https://github.com/google/xls/blob/main/xls/modules/fp/fp32_fast_rsqrt.x, I get the following error:
```
F1214 05:27:40.019014    4105 codegen_main.cc:278] Check failed: ::absl::OkStatus() == (xls::RealMain(ir_path)) (OK vs. UNIMPLEMENTED: Unhandled node for delay estimation in delay model 'sky130': one_hot.5472: bits[32] = one_hot(concat.6705: bits[31], lsb_prio=true, id=5472, pos=[(0,196,14)])) 
```

proppy:
Fixes #806 

@vincent-mirian-google should we have a warning/note about the model not being accurate?

### DIFF
--- a/xls/delay_model/models/sky130.textproto
+++ b/xls/delay_model/models/sky130.textproto
@@ -345,6 +345,7 @@ op_models { op: "kArrayIndex" estimator { alias_op: "kOr" } }
 op_models { op: "kNor" estimator { alias_op: "kOr" } }
 op_models { op: "kNand" estimator { alias_op: "kAnd" } }
 op_models { op: "kOneHotSel" estimator { alias_op: "kOr" } }
+op_models { op: "kOneHot" estimator { alias_op: "kOr" } }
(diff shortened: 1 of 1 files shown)


vincent-mirian-google:
@proppy There is already a comment in the file touching on the following ops being modeled by 'bogus alias ops'.

proppy:
> comment in the file touching on the following ops being modeled by 'bogus alias ops'

Yep 👍 I saw, I was thinking of some runtime thing to inform the user that the scheduling wouldn't leverage accurate timing info if their IR has those alias ops'.

vincent-mirian-google:
(I figured that 's what you meant originally.)...

That's a good point. Typically, models are an approximation. We can add a warning that the op was not characterized. That being said, there is no infrastructure that enables the latter.

PR #855, opened 2023-02-09
Don't use hardcoded paths to /bin/bash /usr/bin/python

hzeller:
In more pure environments (such as nixos), these paths do not exist.
In general, the recommended way in the unix environment is to use /usr/bin/env to resolve binaries, so

  * `#!/usr/bin/env bash`
  * `#!/usr/bin/env python`

Signed-off-by: Henner Zeller <h.zeller@acm.org>


cdleary:
Still some failures to resolve from CI? LMK when to PTAL.

hzeller:
Once #855 is merged into git, I can rebase and then see the failure.
The tests in question don't run yet at all on my machine, so I can't see what is wrong there.

PR #856, opened 2023-02-09
Output errors in tests.

hzeller:
This makes make the CI outputs actionable.

... this will help to see why the CI tests in #855 fail :)


meheff:
Thanks!

PR #876, opened 2023-03-03
README: add conda package

proppy:
- add badge
- add instructions
- reword "build from source" intro

/cc @cdleary

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,12 @@
 
 # **XLS**: Accelerated HW Synthesis
 
-[**Docs**](https://google.github.io/xls/) | [**Quick Start**](https://google.github.io/xls/tools_quick_start/) [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/chipsalliance/silicon-notebooks/blob/main/xls-adder-openlane.ipynb) | [**Tutorials**](https://google.github.io/xls/tutorials/) ![Ubuntu 20.04](https://github.com/google/xls/workflows/Continuous%20Integration/badge.svg) ![Ubuntu 22.04](https://github.com/google/xls/workflows/Nightly%20Ubuntu%2022.04/badge.svg)
+[**Docs**](https://google.github.io/xls/) | [**Quick Start**](https://google.github.io/xls/tools_quick_start/) [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/chipsalliance/silicon-notebooks/blob/main/xls-adder-openlane.ipynb) | [**Tutorials**](https://google.github.io/xls/tutorials/)
+
+[**Conda packages**](https://anaconda.org/litex-hub/xls) [![Anaconda-Server Badge](https://anaconda.org/litex-hub/xls/badges/latest_release_date.svg)](https://anaconda.org/litex-hub/xls) [![Anaconda-Server Badge](https://anaconda.org/litex-hub/xls/badges/platforms.svg)](https://anaconda.org/litex-hub/xls)
+
+![Ubuntu 20.04](https://github.com/google/xls/workflows/Continuous%20Integration/badge.svg) ![Ubuntu 22.04](https://github.com/google/xls/workflows/Nightly%20Ubuntu%2022.04/badge.svg)
+
 
 ## What is XLS?
 
@@ -43,9 +48,22 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
 [reporting bugs](https://github.com/google/xls/issues), and letting us know
 what you think!
 
+## Install using conda
+
+```bash
+curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+bash Miniconda3-latest-Linux-x86_64.sh -p conda-env/ -b
+source conda-env/bin/activate
+conda install --yes -c litex-hub xls
+interpreter_main --version
+ir_converter_main --version
+opt_main --version
+codegen_main --version
+```
+
 ## Building From Source
 
-Currently, XLS must be built from source using the Bazel build system.
+You can build XLS from source using the Bazel build system.
 
 *Note:* Binary distributions of the XLS library are not currently available, but
 we hope to enable them via continuous integration,


README.md
```diff
@@ -43,9 +43,21 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
 [reporting bugs](https://github.com/google/xls/issues), and letting us know
 what you think!
 
+## Install using conda
+
+```console
```
proppy, line 48: wonder if that's the actual right decorator

README.md
```diff
@@ -43,9 +43,21 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
...
+curl -O https://repo.anaconda.com/miniconda/Miniconda3-py37_23.1.0-1-Linux-x86_64.sh
```
proppy, line 49: @cdleary do you want shell prompt and output in there?

I personally prefer it that way so that you can easily cut and paste.

README.md
```diff
@@ -43,9 +43,21 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
...
+bash Miniconda3-py37_23.1.0-1-Linux-x86_64.sh -p conda-env/ -b
+source conda-env/bin/activate
+interpreter_main --version
```
proppy, line 52: @cdleary this currently retusn `interpreter_main` is that expected?

README.md
```diff
@@ -43,9 +43,21 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
...
+ir_converter_main --version
+opt_main --version
+codegen_main --version
+```
+
 ## Building From Source
 
-Currently, XLS must be built from source using the Bazel build system.
+Can yan build XLS from source using the Bazel build system.
```
proppy, line 60: typo

README.md
proppy: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L6384

proppy: if we don't have the prompt `bash` is probably better here:
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L6302

README.md
cdleary: Go for it!

README.md
cdleary: We've never done anything in particular to hook into that version flag output, must be what absl flags impl does by default.

README.md
```diff
@@ -43,9 +43,22 @@ out, running through [some tutorials](https://google.github.io/xls/tutorials/),
 [reporting bugs](https://github.com/google/xls/issues), and letting us know
 what you think!
 
+## Install using conda
+
+```bash
+curl -O https://repo.anaconda.com/miniconda/Miniconda3-py37_23.1.0-1-Linux-x86_64.sh
+bash Miniconda3-py37_23.1.0-1-Linux-x86_64.sh -p conda-env/ -b
```
cdleary, line 50: OOC does this only work for Python 3.7?

proppy: since include `python 3.7`, I wanted to default on this since some of the other conda-eda packages only support 3.7 right now, see https://github.com/hdl/conda-eda/issues/72 and https://github.com/hdl/conda-eda/issues/260

README.md
```diff
@@ -4,7 +4,7 @@
 
 # **XLS**: Accelerated HW Synthesis
 
-[**Docs**](https://google.github.io/xls/) | [**Quick Start**](https://google.github.io/xls/tools_quick_start/) | [**Tutorials**](https://google.github.io/xls/tutorials/) ![Ubuntu 20.04](https://github.com/google/xls/workflows/Continuous%20Integration/badge.svg) ![Ubuntu 22.04](https://github.com/google/xls/workflows/Nightly%20Ubuntu%2022.04/badge.svg)
+[**Docs**](https://google.github.io/xls/) | [**Quick Start**](https://google.github.io/xls/tools_quick_start/) | [**Tutorials**](https://google.github.io/xls/tutorials/) | [**Conda packages**](https://anaconda.org/litex-hub/xls) [![Anaconda-Server Badge](https://anaconda.org/litex-hub/xls/badges/latest_release_date.svg)](https://anaconda.org/litex-hub/xls) [![Anaconda-Server Badge](https://anaconda.org/litex-hub/xls/badges/platforms.svg)](https://anaconda.org/litex-hub/xls) | ![Ubuntu 20.04](https://github.com/google/xls/workflows/Continuous%20Integration/badge.svg) ![Ubuntu 22.04](https://github.com/google/xls/workflows/Nightly%20Ubuntu%2022.04/badge.svg)
```
cdleary, line 7: Nit: can you put this on a different line? (Otherwise it starts to look quite crowded.)

README.md
cdleary: Can you include at the start of the section in the markdown whether these instruction will /only/ work for Python 3.7? e.g. right now I have Python 3.10 and I'm not sure if this means "needs at least 3.7" or " needs exactly 3.7".

If it's "exactly 3.7" but we anticipate being able to support more point-versions if we could point at a GitHub issue that'd be swell.

cdleary (commented):
Do you mind also rebaseing to HEAD? I don't think it'll affect the content of the PR but it may help the way it syncs with the repo, thanks!

README.md
proppy: Sorry I misspelled my previous comment, what I meant is that the miniconda installer by default *includes* its own distribution of python 3.7, there is no requirements on the version of python installed on the host (could be none or different, it wouldn't affect the packages executed from the conda environment, and wouldn't be applicable here anyway since we're only packaging the tool executables)

proppy: So the reason I'm using 3.7 here is to ensure better compatiblity with the *others* conda-eda packages, but we could default on https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh if we didn't want to be opiniated here.

proppy: reverter to Miniconda3-latest-Linux-x86_64.sh i

proppy:
> Do you mind also rebaseing to HEAD? I don't think it'll affect the content of the PR but it may help the way it syncs with the repo, thanks!

Done.

README.md
proppy: Done.

proppy:
PTAL

README.md
cdleary: Thanks for explaining!

PR #928, opened 2023-04-10
github/workflows: move all remaining jobs to bigger runners

proppy:
- run on bigger runners
- enable disabled targets and tests
- remove cache cleanup

### DIFF
--- a/.github/workflows/build-mkdocs-commit-html.yml
+++ b/.github/workflows/build-mkdocs-commit-html.yml
@@ -10,7 +10,8 @@ on:
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: ubuntu-22.04-4core
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -12,7 +12,8 @@ on:
 jobs:
   build:
     name: Nightly Fuzz
-    runs-on: ubuntu-20.04
+    runs-on:
+      labels: ubuntu-20.04-64core
     steps:
       - uses: actions/checkout@v2
       - name: Mount Bazel Cache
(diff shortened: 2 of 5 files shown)


meheff (approved):
Looks good to me but @hongted should probably look at it.

.github/workflows/build-mkdocs-commit-html.yml
```diff
@@ -10,7 +10,8 @@ on:
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: ubuntu-22.04-4core
```
hongted, line 14: Is this the smallest machine available?

.github/workflows/nightly-fuzz.yml
```diff
@@ -35,52 +36,11 @@ jobs:
...
+      - name: Bazel Run Fuzz (opt)
         run: |
-          "${GITHUB_WORKSPACE}/bin/bazel" run -c opt xls/fuzzer:run_fuzz_multiprocess -- --crash_path "${GITHUB_WORKSPACE}/crashers" --sample_count=1024 --codegen=false --simulate=false --emit_loops=False --summary_path "${GITHUB_WORKSPACE}/crashers"
+          "${GITHUB_WORKSPACE}/bin/bazel" run -c opt xls/fuzzer:run_fuzz_multiprocess -- --crash_path "${GITHUB_WORKSPACE}/crashers" --sample_count=2048 --summary_path "${GITHUB_WORKSPACE}/crashers"
       - name: Upload Fuzz Crashers
         uses: actions/upload-artifact@v2
         with:
           name: crashers
           path: "${GITHUB_WORKSPACE}/crashers"
-  cleanup:
```
hongted, line 48: I think we still want to keep the cleanup.  Even with the faster machines, the cache is still useful for even more speedup.

proppy: This still use the cache, it's just that we don't need to work around the invalidation logic anymore because a cache miss would still make the build complete in a reasonable time.

.github/workflows/build-mkdocs-commit-html.yml
proppy: Using larger runner yes, see: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners

hongted: In that case, I'd suggest moving to a smaller one.  I don't think mkdocs needs a larer runner.

.github/workflows/nightly-fuzz.yml
hongted: The idea is that Github's default invalidation logic doesn't work for us.  If we don't have this then our cache will be filled with caches created from CL checkins and the nightly builds will likely not have a cache to use.

This logic keeps the caches clean -- we'll only have the latest cache for each OS around (assuming Github doesn't evict us for going over their 10GB limit).

proppy: > nightly builds will likely not have a cache to use

would it make sense to use a different cache key for the nightly build? That way its eviction lifecycle would be decouple from pending PRs.

proppy: Additionally, even without caching the nightly build would still complete in a reasonable time without timeoutong (<1 hour) now that it leverages larger runner.

proppy: Another reason I'm disabling the cache cleanup in this PR, is that #925 highlighted some issue between the current `gh` tool and the larger runner runtime environment, see https://github.com/actions/gh-actions-cache/issues/56#issuecomment-1499829778

hongted: There should already be two cache keys in use (and the cleanup rule is so that the the less-used cache key isn't evicted by the more often used one).  Github eviction policy is global and not on a per-key basis.

But if this doesn't work on the larger runners, then ok to remove.

PR #932, opened 2023-04-14
dslx/stdlib: add iterative_div_mod

proppy:
- add `iterative_div_mod` based on https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581
- unify implementation with existing `iterative_div` function

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -199,35 +199,69 @@ fn smul_test() {
     assert_eq(s4:0b1111, smul(s2:0b11, s2:0b01));
 }
 
-// Calculate x / y one bit at a time. This is an alternative to using
-// the division operator '/' which may not synthesize nicely.
-pub fn iterative_div<N: u32, DN: u32 = {N * u32:2}>(x: uN[N], y: uN[N]) -> uN[N] {
-    let init_shift_amount = ((N as uN[N]) - uN[N]:1);
-    let x = x as uN[DN];
+// Returns unsigned division of `n` (N bits) and `d` (M bits) as quotient (N bits) and remainder (M
+// bits).
+// If dividing by `0`, returns alls `1`s for quotient and `n` for remainder.
+// Based on the implementation in //xls/ir/abstract_evaluator:UDivMod;
+// should be expected to use a large number of gates and have a slow
+// critical path when using combinational codegen.
+fn iterative_div_mod<N: u32, M: u32>(n: uN[N], d: uN[M]) -> (uN[N], uN[M]) {
+    // Zero extend divisor by 1 bit.
+    let divisor = d as uN[M + u32:1];
 
-    let (_, _, _, div_result) = for
-        (idx, (shifted_y, shifted_index_bit, running_product, running_result)) in range(u32:0, N) {
-        // Increment running_result by current power of 2
-        // if the prodcut running_result * y < x.
-        let inc_running_result = running_result | shifted_index_bit;
-        let inc_running_product = running_product + shifted_y;
-        let (running_result, running_product) = if inc_running_product <= x {
-            (inc_running_result, inc_running_product)
+    for (i, (q, r)): (u32, (uN[N], uN[M])) in range(u32:0, N) {
+        // Shift the next bit of n into r.
+        let r = r ++ n[(N - u32:1 - i)+:u1];
+        let (q, r) = if r >= divisor {
+            (q as uN[N - u32:1] ++ u1:1, r - divisor)
         } else {
-            (running_result, running_product)
+            (q as uN[N - u32:1] ++ u1:0, r)
         };
+        // Remove the MSB of r; guaranteed to be 0 because r < d.
+        (q, r[0:M as s32])
+    }((uN[N]:0, uN[M]:0))
+}
 
-        // Shift to next (lower) power of 2.
-        let shifted_y = shifted_y >> uN[N]:1;
-        let shifted_index_bit = shifted_index_bit >> uN[N]:1;
+// Returns unsigned division of `n` (N bits) and `d` (M bits) as quotient (N bits).
+// If dividing by `0`, returns alls `1`s for quotient.
+fn iterative_div<N: u32, M: u32>(n: uN[N], d: uN[M]) -> uN[N] {
+    let (q, r) = iterative_div_mod(n, d);
+    q
+}
 
-        (shifted_y, shifted_index_bit, running_product, running_result)
-    }((
-        (y as uN[DN]) << (init_shift_amount as uN[DN]), uN[N]:1 << init_shift_amount, uN[DN]:0,
-        uN[N]:0,
-    ));
+#[test]
+fn iterative_div_mod_test() {
+    // Power of 2.
+    assert_eq((u4:0, u4:8), iterative_div_mod(u4:8, u4:15));
+    assert_eq((u4:1, u4:0), iterative_div_mod(u4:8, u4:8));
+    assert_eq((u4:2, u4:0), iterative_div_mod(u4:8, u4:4));
+    assert_eq((u4:4, u4:0), iterative_div_mod(u4:8, u4:2));
+    assert_eq((u4:8, u4:0), iterative_div_mod(u4:8, u4:1));
+    assert_eq((u4:8 / u4:0, u4:8), iterative_div_mod(u4:8, u4:0));
+    assert_eq((u4:15, u4:8), iterative_div_mod(u4:8, u4:0));
 
-    div_result
+    // Non-powers-of-2.
+    assert_eq((u32:6, u32:0), iterative_div_mod(u32:18, u32:3));
+    assert_eq((u32:6, u32:0), iterative_div_mod(u32:36, u32:6));
+    assert_eq((u32:6, u32:0), iterative_div_mod(u32:48, u32:8));
+    assert_eq((u32:20, u32:0), iterative_div_mod(u32:900, u32:45));
+
+    // Results w/ remainder.
+    assert_eq((u32:6, u32:2), iterative_div_mod(u32:20, u32:3));
+    assert_eq((u32:6, u32:5), iterative_div_mod(u32:41, u32:6));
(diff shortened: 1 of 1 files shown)


meheff (approved):
Thanks!

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
...
+  let remainder: bits[N] = dividend;
+  let term: bits[DN] = bits[DN]:1 << (N as bits[DN]);
+  let product: bits[DN] = (divisor as bits[DN]) << (N as bits[DN]);
+  let (quotient, remainder, product, term) =
+     for (i, (quotient, remainder, product, term)): (bits[N], (bits[N], bits[N], bits[DN], bits[DN]))
+     in range(bits[N]:0, (N as bits[N])) {
+       let product = product >> bits[DN]:1;
+       let term = term >> bits[DN]:1;
+       let (new_q, new_r) : (bits[N], bits[N]) =
+         match product <= (remainder as bits[DN]) {
```
proppy, line 193: @grebe commented in #935

> Seems simpler to do an if else than matching on true/false

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
...
 
+// Binary Division algorithm as outlined by:
+// http://www.dragonwins.com/domains/GetTechEd/de248/binary_division.htm
+fn iterative_binary_div<N: u32, DN: u32 = {N * u32:2}>(dividend: bits[N], divisor: bits[N]) -> (bits[N], bits[N]) {
+  let quotient: bits[N] = bits[N]:0;
+  let remainder: bits[N] = dividend;
+  let term: bits[DN] = bits[DN]:1 << (N as bits[DN]);
+  let product: bits[DN] = (divisor as bits[DN]) << (N as bits[DN]);
+  let (quotient, remainder, product, term) =
+     for (i, (quotient, remainder, product, term)): (bits[N], (bits[N], bits[N], bits[DN], bits[DN]))
```
proppy, line 188: @grebe commented in #935 

> Making your loop indices u32 seems like it will make this a little more readable- I forget, can you do range(0, N) or will you have to do range(u32:0, N)

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
   ()
 }
 
+// Binary Division algorithm as outlined by:
+// http://www.dragonwins.com/domains/GetTechEd/de248/binary_division.htm
+fn iterative_binary_div<N: u32, DN: u32 = {N * u32:2}>(dividend: bits[N], divisor: bits[N]) -> (bits[N], bits[N]) {
```
proppy, line 182: @grebe commented in https://github.com/google/xls/pull/935

> May want a comment describing return values (and their order)

proppy, line 182: @grebe commented in https://github.com/google/xls/pull/935

> Do you think it would be much more work to let dividend and divisor have different widths? I think you'd just adjust the remainder width, perhaps extending when you do the comparison. Maybe that would add some special cases when divisor width > dividend width? I haven't thought through it all, and no worries if you'd rather not worry about it for now and just want to submit what's here.

@ericastor also commented:

> Agreed that it's not too complicated; actually, we have an implementation of a full arbitrary-width integer DivMod (both signed & unsigned) as part of the AbstractEvaluator inside of XLS here: https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581
> It should be much nicer in DSLX!

proppy: ported the IR arbitrary-width version you [referenced](https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581) to DSLX.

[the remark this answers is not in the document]
proppy: ditched the internal `//platforms` version this PR was adding in favor of the a DSLX port of the existing IR `DivMod` that @ericastor [pointed](https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581).
also unified the existing `iterative_div` implementation to simply drop the remainder as suggested by @hongted.

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
   ()
 }
 
+// Binary Division algorithm as outlined by:
+// http://www.dragonwins.com/domains/GetTechEd/de248/binary_division.htm
```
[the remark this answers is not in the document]
proppy: should we link to https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581 now?

proppy:
PTAL, unified the DSLX stdlib division with the existing IR divmod implementation.

xls/dslx/stdlib/std.x
```diff
@@ -113,61 +113,69 @@ fn umul_test() {
...
+fn iterative_div_mod_test() {
   // Power of 2.
-  assert_eq(u4:0, iterative_div(u4:8, u4:15));
-  assert_eq(u4:1, iterative_div(u4:8, u4:8));
-  assert_eq(u4:2, iterative_div(u4:8, u4:4));
-  assert_eq(u4:4, iterative_div(u4:8, u4:2));
-  assert_eq(u4:8, iterative_div(u4:8, u4:1));
-  assert_eq(u4:8 / u4:0, iterative_div(u4:8, u4:0));
-  assert_eq(u4:15, iterative_div(u4:8, u4:0));
+  let _ = assert_eq((u4:0, u4:8), iterative_div_mod(u4:8, u4:15));
```
proppy, line 144: we shouldn't reintroduce the `let _ = `

xls/dslx/stdlib/std.x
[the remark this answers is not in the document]
ericastor: Probably not a bad idea to make the link explicit, yeah.

Also probably worth adding a comment (like I did in the AbstractEvaluator) calling out that this is a slow operation to do combinationally, and might be worth pipelining in some circumstances.

As a separate followup, it's probably worth it for us to verify which synthesizes better in practice for combinational implementations at different sizes: this, restoring division (a really small modification, just replacing (r >= d) with an unconditional subtract, test if negative, & adding d back if the result is negative), or non-restoring division (per https://en.wikipedia.org/wiki/Division_algorithm#Non-restoring_division).

xls/dslx/stdlib/std.x
```diff
@@ -113,61 +113,67 @@ fn umul_test() {
...
-    let shifted_index_bit = shifted_index_bit >> uN[N]:1;
-
-    (shifted_y, shifted_index_bit, running_product, running_result)
-  } (( (y as uN[DN]) << (init_shift_amount as uN[DN]),
-       uN[N]:1 << init_shift_amount,
-       uN[DN]:0,
-       uN[N]:0));
-
-  div_result
+// Returns unsigned division of n (N bits) and d (M bits) as quotient (N bits) and remainder (M bits).
```
ericastor, line 116: Can we document the behavior of this implementation when dividing by zero?

If I'm not mistaken, as written, if d == 0, the output will be (q = all 1's, r = n).

xls/dslx/stdlib/std.x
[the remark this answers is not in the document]
proppy: do we want to support the 3 variants with parametrics?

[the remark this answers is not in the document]
ericastor: Interesting thought. I think for reasonable ease of use, that would mean ~parametrizing on an *enum*... and I'm not sure we can do that. Let's go ahead with this as-is for now, and consider our options later.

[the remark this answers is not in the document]
proppy: I think it would work (although we might have to have additional wrapper functions to address partial specialization and default value); I'd be curious if the IR optimization pass would be capable of optimization the constant and the match expression away if we made it a regular parameter.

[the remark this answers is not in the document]
proppy: > consider our options later

filed #1046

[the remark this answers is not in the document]
proppy: updated docstring to mention the original IR implementation and current limitations.

xls/dslx/stdlib/std.x
proppy: Done.

proppy:
PTAL

ericastor (approved):

cdleary:
I think this need a rebase to merge, thanks.

xls/dslx/stdlib/std.x
```diff
@@ -199,35 +199,69 @@ fn smul_test() {
...
-        (idx, (shifted_y, shifted_index_bit, running_product, running_result)) in range(u32:0, N) {
-        // Increment running_result by current power of 2
-        // if the prodcut running_result * y < x.
-        let inc_running_result = running_result | shifted_index_bit;
-        let inc_running_product = running_product + shifted_y;
-        let (running_result, running_product) = if inc_running_product <= x {
-            (inc_running_result, inc_running_product)
+// Returns unsigned division of `n` (N bits) and `d` (M bits) as quotient (N bits) and remainder (M
+// bits).
+// If dividing by `0`, returns alls `1`s for quotient and `n` for remainder.
```
ericastor, line 204: nit, spelling: "alls" -> "all" (here and below).

xls/dslx/stdlib/std.x
```diff
@@ -199,35 +199,69 @@ fn smul_test() {
...
+// Based on the implementation in //xls/ir/abstract_evaluator:UDivMod;
```
ericastor, line 205: Rather than citing that implementation, can we point out that this is actually binary long division?

PR #935, opened 2023-04-20
examples: add gcd

proppy:
- stacked on top of #932
- add `gcd_euclidean`
- add `gcd_binary`
- cross-test both implementation using `quicktest`

### DIFF
--- a/xls/examples/gcd.x
+++ b/xls/examples/gcd.x
@@ -1,0 +1,69 @@
+// Copyright 2023 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import std;
+
+// https://en.wikipedia.org/wiki/Greatest_common_divisor#Euclidean_algorithm
+fn gcd_euclidean<N: u32, DN: u32 = {N * u32:2}>(a: uN[N], b: uN[N]) -> uN[N] {
+    let (_, gcd) = for (i, (a, b)) in range(u32:0, DN) {
+        let (d, r) = std::iterative_div_mod(a, b);
+        if (r == uN[N]:0) {
+            (a, b)
+        } else {
+            (b, r)
+        }
+    }((a, b));
+    gcd
+}
+
+fn gcd_binary_match<N: u32>(a: uN[N], b: uN[N], d: uN[N]) -> (uN[N], uN[N], uN[N]) {
+    match (a[0:1], b[0:1]) {
+      (u1:0, u1:1) => (b, a >> 1, d),
+      (u1:1, u1:0) => (a, b >> 1, d),
+      (u1:0, u1:0) => (a >> 1, b >> 1, d+uN[N]:1),
+      (u1:1, u1:1) => ((a - b) >> 1, b, d),
+      _ => fail!("unsupported_case", (uN[N]:0, uN[N]:0, uN[N]:0)),
+    }
+}
+
+// https://en.wikipedia.org/wiki/Greatest_common_divisor#Binary_GCD_algorithm
+fn gcd_binary<N: u32, DN: u32 = {N * u32:2}>(a: uN[N], b: uN[N]) -> uN[N] {
+    let (a, _, d) = for (i, (a, b, d)) in range(u32:0, DN) {
+        if (a == b) {
+            (a, b, d)
+        } else if (a < b) {
+            gcd_binary_match(b, a, d)
+        } else {
+            gcd_binary_match(a, b, d)
+        }
+    }((a, b, uN[N]:0));
+    (a << d)
+}
+
+#[test]
+fn gcd_euclidean_test() {
+    assert_eq(u8:6, gcd_euclidean(u8:48, u8:18));
+    assert_eq(u8:6, gcd_euclidean(u8:18, u8:48));
+}
+
+#[test]
+fn gcd_binary_test() {
+    assert_eq(u8:6, gcd_binary(u8:48, u8:18));
+    assert_eq(u8:6, gcd_binary(u8:18, u8:48));
+}
+
+#[quickcheck(test_count=50000)]
+fn prop_gcd_equal(a: u32, b: u32) -> bool {
+    gcd_euclidean<u32:32>(a, b) == gcd_binary<u32:32>(a, b)
+}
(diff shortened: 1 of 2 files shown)


proppy:
rebased

xls/examples/gcd.x
```diff
@@ -0,0 +1,74 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import std
+
+// https://en.wikipedia.org/wiki/Greatest_common_divisor#Euclidean_algorithm
+fn gcd_euclidean<N: u32, DN: u32 = {N * u32:2}>(a: uN[N], b: uN[N]) -> uN[N] {
+    let (_, gcd) = for (i, (a, b)) in range(u32:0, DN) {
+        let (d, r) = std::iterative_binary_div(a, b);
```
proppy, line 20: should move this in the else case, so that we don't perform unnecessary division when the remainder is `0`.

meheff: Not sure I understand. How can this move into the else case if the result of the computation is the condition?

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
...
+  let remainder: bits[N] = dividend;
+  let term: bits[DN] = bits[DN]:1 << (N as bits[DN]);
+  let product: bits[DN] = (divisor as bits[DN]) << (N as bits[DN]);
+  let (quotient, remainder, product, term) =
+     for (i, (quotient, remainder, product, term)): (bits[N], (bits[N], bits[N], bits[DN], bits[DN]))
+     in range(bits[N]:0, (N as bits[N])) {
+       let product = product >> bits[DN]:1;
+       let term = term >> bits[DN]:1;
+       let (new_q, new_r) : (bits[N], bits[N]) =
+         match product <= (remainder as bits[DN]) {
```
grebe, line 193: Seems simpler to do an if else than matching on true/false

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
...
+// Binary Division algorithm as outlined by:
+// http://www.dragonwins.com/domains/GetTechEd/de248/binary_division.htm
+fn iterative_binary_div<N: u32, DN: u32 = {N * u32:2}>(dividend: bits[N], divisor: bits[N]) -> (bits[N], bits[N]) {
+  let quotient: bits[N] = bits[N]:0;
+  let remainder: bits[N] = dividend;
+  let term: bits[DN] = bits[DN]:1 << (N as bits[DN]);
+  let product: bits[DN] = (divisor as bits[DN]) << (N as bits[DN]);
+  let (quotient, remainder, product, term) =
+     for (i, (quotient, remainder, product, term)): (bits[N], (bits[N], bits[N], bits[DN], bits[DN]))
+     in range(bits[N]:0, (N as bits[N])) {
```
grebe, line 189: Making your loop indices u32 seems like it will make this a little more readable- I forget, can you do `range(0, N)` or will you have to do `range(u32:0, N)`

xls/dslx/stdlib/std.x
```diff
@@ -177,6 +177,53 @@ fn iterative_div_test () {
   ()
 }
 
+// Binary Division algorithm as outlined by:
+// http://www.dragonwins.com/domains/GetTechEd/de248/binary_division.htm
+fn iterative_binary_div<N: u32, DN: u32 = {N * u32:2}>(dividend: bits[N], divisor: bits[N]) -> (bits[N], bits[N]) {
```
grebe, line 182: May want a comment describing return values (and their order)

grebe, line 182: Do you think it would be much more work to let dividend and divisor have different widths? I think you'd just adjust the remainder width, perhaps extending when you do the comparison. Maybe that would add some special cases when divisor width > dividend width? I haven't thought through it all, and no worries if you'd rather not worry about it for now and just want to submit what's here.

xls/examples/gcd.x
grebe: I think gcd is a good candidate for proc implementations, which would let you do the "early return" I think you're describing. Each proc tick would do one division (or `gcd_binary_match`), and you send the output when you reach the stop condition.

Fine to leave as a TODO, and these function implementations are nice because they work well w/ quickcheck.

grebe (changes requested):
This is great!

xls/dslx/stdlib/std.x
ericastor: Agreed that it's not too complicated; actually, we have an implementation of a full arbitrary-width integer DivMod (both signed & unsigned) as part of the AbstractEvaluator here inside of XLS here: https://github.com/google/xls/blob/fa7af6cbda69d02d2d50b28fa5d95f51fe8cdfb1/xls/ir/abstract_evaluator.h#L513-L581

It should be much nicer in DSLX!

proppy: sure, there is actually a separate PR for that part of the change: #932, should we move the discussion there?

xls/examples/gcd.x
proppy: > Not sure I understand. How can this move into the else case if the result of the computation is the condition?

@meheff since we always pass `r` to the next iteration, I was thinking we could `sel` based on the result of the previous division, rather than always performing the same division again and again until the end of the loop.

proppy: > I think gcd is a good candidate for proc implementations, which would let you do the "early return" I think you're describing.

It could be interesting to compare a proc implementation to the automatically pipelined version.

xls/dslx/stdlib/std.x
proppy: I think this was done in #932.

xls/dslx/stdlib/std.x
proppy: I believe this was addressed in #932

xls/dslx/stdlib/std.x
proppy: I believe this was addressed/obsoleted by https://github.com/google/xls/pull/932

xls/dslx/stdlib/std.x
proppy: I believe this was addressed in https://github.com/google/xls/pull/932

proppy:
> I think gcd is a good candidate for proc implementations

Filed #1444

xls/examples/gcd.x
proppy: > @meheff since we always pass `r` to the next iteration, I was thinking we could `sel` based on the result of the previous division, rather than always performing the same division again and again until the end of the loop.

As discussed simplified the implementation to only trigger div_mod in `for` recurrence w/ a non-zero modulo.

This apparently result in a reduced critical path and bom, see https://colab.research.google.com/gist/proppy/01142e33a2be8b71d6b317ee4ce7fde4/xls-playground-sandbox.ipynb

proppy:
PTAL

PR #956, opened 2023-05-14
LFSR example

mdagois:
The example contains
* a template to flexibly define LFSR of any bit width with any number of taps
* several examples of maximal LFSRs (2-bit to 24-bit) using the template
* tests for 7-bit and 8-bit LFSRs


xls/examples/lfsr.x
```diff
@@ -0,0 +1,518 @@
+////////////////////////////////////////////////////////////////////////////////
+// Linear-feedback shift register (LFSR)
+// A template function to compute the next value of an LFSR counter.
```
proppy, line 3: s/template/parametric/
see: https://google.github.io/xls/dslx_reference/#parametric-functions

xls/examples/lfsr.x
```diff
@@ -0,0 +1,518 @@
...
+	let new_bit = for (index, xor_bit): (u32, u1) in range(u32:0, BIT_WIDTH) {
+		if tap_mask[index+:u1] == u1:0 {
+			xor_bit
+		} else {
+			xor_bit ^ current_value[index+:u1]
+		}
+	} (u1:0);
+
+	// Kick the high bit and insert the new bit
+	current_value[0:{(BIT_WIDTH - u32:1) as s32}] ++ new_bit
```
proppy, line 21: curious if https://google.github.io/xls/dslx_reference/#width-slice would make things easier there?

xls/examples/lfsr.x
```diff
@@ -0,0 +1,518 @@
...
+
+fn lfsr<BIT_WIDTH: u32>(current_value: uN[BIT_WIDTH], tap_mask: uN[BIT_WIDTH]) -> uN[BIT_WIDTH] {
+	// Compute the new bit from the taps
+	let new_bit = for (index, xor_bit): (u32, u1) in range(u32:0, BIT_WIDTH) {
+		if tap_mask[index+:u1] == u1:0 {
+			xor_bit
+		} else {
+			xor_bit ^ current_value[index+:u1]
+		}
+	} (u1:0);
```
proppy, line 18: is the input of the lfsr always 0? or should it be a parameter?

xls/examples/lfsr.x
```diff
@@ -0,0 +1,518 @@
...
+	// Kick the high bit and insert the new bit
+	current_value[0:{(BIT_WIDTH - u32:1) as s32}] ++ new_bit
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// A series of maximal LFSRs for different bit widths.
+// Source: https://en.wikipedia.org/wiki/Linear-feedback_shift_register
+////////////////////////////////////////////////////////////////////////////////
+
+fn lfsr2(n: u2) -> u2 {
```
proppy, line 29: not sure if you need those wrapper, the interpreter should be able to infer the function type parameters.

xls/examples/lfsr.x
mdagois: Fixed.

xls/examples/lfsr.x
mdagois: I used the width slice instead of the cast.

xls/examples/lfsr.x
```diff
@@ -0,0 +1,518 @@
...
+fn lfsr24(n: u24) -> u24 {
+	lfsr(n, u24:0b111000010000000000000000)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Tests
+////////////////////////////////////////////////////////////////////////////////
+
+#[test]
+fn lfsr7_test() {
```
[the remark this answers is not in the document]
mdagois: Generated with some scrappy scripts.
Isn't cc_test more cumbersome? You'd need a compiler, while you can develop DSLX algorithms entirely with the interpreter if the test are written in DSLX.

xls/examples/lfsr.x
mdagois: Yes, it is always 0, because xor'ing a bit with 0 always yields that bit.
That 0 has basically no impact on the result of the for loop, which is the intended behavior, as the algorithm only cares about the xor of the tap bits in the current_value.

xls/examples/lfsr.x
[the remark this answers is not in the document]
proppy: it's good to have some simple dslx unittests, but it can also be nice to show validation against a software implementation in `cc_test` (similar to what's done in `//third_party/xls_colors`), the bazel workspace already assume a c++ compiler to build XLS itself ;)

xls/examples/lfsr.x
proppy: ah I see, so maybe we can only keep a few ones that are actually tested and and a descriptive comment (X-bit lfsr with a period of XXX)?

also curious if it's possible to predict the period from the tap masks?

mdagois: No, you cannot predict the period from the tap masks. All you know is that the maximal theoretical period is 2^n - 1.
I think all you can do is try all kind of masks until you find a maximal one.
For a low number of bits, it is fast enough to try all combination and find a sequence that has the exact period you want. Assuming such a sequence exists of course.
I think Xilinx has discovered a maximal LFSR on 168 bits. It is the longest I have heard of, and it uses only 4 tap bits.

xls/examples/lfsr.x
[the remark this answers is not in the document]
mdagois: OK, so I completely agree it is nice to have cc_test in the context of xls. For example, if the LFSR were in std, it would make sense to add cc_test's and the user could assume the function just works. But, for a sample in the example directory, isn't it better to keep it all in pure DSLX?
Maybe I am biased by my setup, but basically, I develop all my samples in a separate private repo, without any bazel, as if I were a developer to which the binaries (interpreter_main, etc.) had been provided. I don't develop anything inside the xls repo.
I hope it makes sense.

Now, if running those tests is expensive or too long, I could definitely understand we don't want too many of them. The LFSR ones are fast, but for the bitonic sample I have in another PR, the tests do take several seconds.

xls/examples/lfsr.x
proppy: Thanks for the detailed explanation!

Would it make sense to capture some of those "famous" LFSR in the examples?

xls/examples/lfsr.x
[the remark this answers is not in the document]
proppy: we do have some cc reference implementation of the example, see https://github.com/google/xls/blob/main/xls/examples/crc32_reference.cc

I wish those were actual tests that validate the DSLX implementation, I'm curious what other thinks?
@meheff @cdleary @antmicro/xls

xls/examples/lfsr.x
mdagois: I removed  the superfluous examples, keeping only the tested ones.
I also added succint comments about the number of bits and period.

xls/examples/lfsr.x
```diff
@@ -0,0 +1,438 @@
...
+    let _ = assert_eq(lfsr7(u7:64), u7:1);
+	()
+}
+
+#[test]
+fn lfsr8_test() {
+	let _ = assert_eq(lfsr8(u8:0), u8:0);
+	let _ = assert_eq(lfsr8(u8:1), u8:2);
+	let _ = assert_eq(lfsr8(u8:2), u8:4);
+	let _ = assert_eq(lfsr8(u8:4), u8:8);
```
proppy, line 183: if we want to keep only dslx test, maybe we could make thiose more succint to ease maintainance (test the first N values and then do noop iteration for `PERIOD - N` and test the next `N` value to verifie it does cycle?

wdyt?

xls/examples/lfsr.x
mdagois: So, I actually tried to test the period for a 24-bit LFSR, but the execution time was horrenduous.
A cc_test would probably fair better, but in the interpreter, it is just not pratical.
To assert the period, one can start from a given value (let's say 1), make a full cycle (by calling the function "period" times) and finally check that the last yielded value is the starting one (i.e. 1). Additionally, one would probably also test that the LFSR does not only return the same value all the time.

proppy: I'm not sure if the `cc_test` would be faster: it would just help to generate the value from a reference implementation rather than hardcoding them; but it would still use the same interpreted code under the hood (unless you use the JIT!)

xls/examples/lfsr.x
mdagois: Yes, I changed the tests to only include the sanity test (with 0), a few random values and the cycle.

xls/examples/lfsr.x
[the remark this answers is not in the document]
proppy: The test comparing against a C++ implementation could come in a separate PR.

xls/examples/lfsr.x
proppy: we could leverage the JIT for this to be faster, but this could be done in a separate PR.

proppy (approved):

PR #958, opened 2023-05-16
[DSLX/IR] Fix ```assert_eq``` in ```#[test_proc()]```

mtdudek:
TestProc and TestFunction were handled differently by GetOrder() function. For TestFunction types calling
```AddToReady``` was an no-op as ```IsReady()``` was
always true.

TestProc handling did not call ```AddToReady``` and
it was adding all proc functions to ready vector.

### DIFF
--- a/xls/examples/BUILD
+++ b/xls/examples/BUILD
@@ -163,9 +163,6 @@ xls_dslx_library(
 xls_dslx_test(
     name = "fir_filter_dslx_test",
     srcs = ["fir_filter.x"],
-    dslx_test_args = {
-        "compare": "none",
-    },
     deps = [":fir_filter_dslx"],
 )
 
@@ -174,9 +171,6 @@ xls_dslx_test(
     srcs = ["matmul_4x4.x"],
     # TODO(rspringer): 2022-05-24: Does not successfully lower to IR.
     # Needs support for arrays of channels.
-    dslx_test_args = {
-        "compare": "none",
-    },
     deps = [
         "//xls/modules/fp:fp32_mul_2",
     ],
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by proppy** [comment]
#673

**Comment by mtdudek** [comment]
Yes, this PR should fix #673.

**Comment by cdleary** [comment]
PRs that both delete code /and/ fix bugs are by far the best. :laughing: Great patch, thanks, testing it downstream.

File: xls/examples/BUILD [hunk 1]
```diff
@@ -163,20 +163,12 @@ xls_dslx_library(
...
-    dslx_test_args = {
-        "compare": "none",
-    },
     deps = [":fir_filter_dslx"],
 )
 
 xls_dslx_test(
     name = "matmul_4x4_dslx_test",
     srcs = ["matmul_4x4.x"],
-    # TODO(rspringer): 2022-05-24: Does not successfully lower to IR.
```
hongted, line 175: Can this TODO be kept to keep a record that we still support lowering this proc to IR?

hongted:
Also, while this PR removes the "compare: 'none'" flags from dslx_tests the toolchain won't compare the DLSX interpretation vs the Jitted IR.  We should add a followup issue for this.  Let me know if you'd want us to create that issue or leave it to you.

xls/examples/BUILD
mtdudek: I've change commit to keep this TODO

mtdudek:
I'll open an issue for dslx_tests not compering interpreter with JIT.
Only after you pointed me to `matmul_4x4`, I noticed it.

Issue #959, opened 2023-05-16
[IR optimization] Over eager IR optimization that leads to incorrect graph


mtdudek:
DSLX allows for process to spawn subprocesses.
It also allows for `procs` to be parameterizable,
which leads to situations were user would like to specialize
`proc <T>A` by spawning its instance in `proc B`'s config
and leave `proc B`'s next function empty (Please see attached file chan_min.x). 

Converting chan_min.x to IR creates 2 processes, inner and outer,
outer is top level process. Outer's next function is empty,
while inner's next function performs some operations on IO channels.
`bazel-bin/xls/dslx/ir_convert/ir_converter_main --top="outer" --dslx_path=./ chan_min.x`
<details><summary>unoptimized IR</summary>

 ```
  package chan_min
  
  file_number 0 "../chan_min.x"
  chan chan_min___in(bits[32], id=0, kind=streaming, ops=receive_only, flow_control=ready_valid, metadata="""""")
  chan chan_min___out(bits[32], id=1, kind=streaming, ops=send_only, flow_control=ready_valid, metadata="""""")
  
  fn __chan_min__inner.init__32() -> () {
    param: bits[32] = literal(value=32, id=1, pos=[(0,0,15)])
    ret tuple.2: () = tuple(id=2, pos=[(0,3,11)])
  }
  
  top proc __chan_min__outer_0_next(__token: token, __state: (), init={()}) {
    literal.5: bits[1] = literal(value=1, id=5)
    tuple.6: () = tuple(id=6, pos=[(0,23,8)])
    after_all.7: token = after_all(__token, id=7)
    next (after_all.7, tuple.6)
  }
  
  proc __chan_min__outer__inner_0__32_next(__token: token, __state: (), init={()}) {
    receive.12: (token, bits[32]) = receive(__token, channel_id=0, id=12)
    tok: token = tuple_index(receive.12, index=0, id=14, pos=[(0,8,13)])
    data: bits[32] = tuple_index(receive.12, index=1, id=15, pos=[(0,8,18)])
    tuple_index.13: token = tuple_index(receive.12, index=0, id=13)
    tok__1: token = send(tok, data, channel_id=1, id=16)
    literal.10: bits[1] = literal(value=1, id=10)
    param: bits[32] = literal(value=32, id=11, pos=[(0,0,15)])
    tuple.17: () = tuple(id=17, pos=[(0,10,8)])
    after_all.18: token = after_all(__token, tuple_index.13, tok__1, id=18)
[...]

Issue #950, opened 2023-05-02
Multi-Proc Codegen


grebe:
We've talked about it a bunch in other issues, but I don't see a dedicated multi-proc codegen issue. I'll open this one an we can fill it in more when we have a more detailed plan about what it will entail.

proppy:
@grebe can this be closed?

PR #964, opened 2023-05-17
xls/build_rules: Add `top` to the list of valid opt_ir flags

rw1nkler:
The `top` flag exposed by the `opt_main` was not added to the list of available flags in the bazel rules. This addition allows the `top` flag to be used as an argument of the `opt_ir_args` parameter of the `xls_dslx_verilog` bazel rule.

This may be used to create more concise `BUILD` files. For example, three rules used to build the delay example may be substituted with only one rule:
```
diff --git a/xls/examples/BUILD b/xls/examples/BUILD
index a39c10af..51e5d460 100644
--- a/xls/examples/BUILD
+++ b/xls/examples/BUILD
@@ -474,22 +474,13 @@ xls_dslx_test(
     library = ":delay_dslx",
 )
 
-xls_dslx_ir(
-    name = "delay_ir",
+xls_dslx_verilog(
+    name = "delay_sv",
     dslx_top = "Delay32x2048_init3",
-    ir_file = "delay.ir",
     library = ":delay_dslx",
-)
-
-xls_ir_opt_ir(
-    name = "delay_opt_ir",
-    src = "delay.ir",
-    top = "__delay__Delay32x2048_init3__Delay__DelayInternal_0__10_32_2048_64_1024_3_next",
-)
-
-xls_ir_verilog(
-    name = "delay_sv",
-    src = ":delay_opt_ir.opt.ir",
+    opt_ir_args = {
+        "top": "__delay__Delay32x2048_init3__Delay__DelayInternal_0__10_32_2048_64_1024_3_next",
+    },
     codegen_args = {
         "module_name": "delay_top",
         "generator": "pipeline",
```

Are `xls_dslx_verilog` and `xls_cc_verilog` the only rules that implement both the `CodegenInfo` and `OptIRInfo` providers used by the `xls_benchmark_verilog` bazel rule? I'm wondering if having this flag isn't the only option to generate Verilog benchmarks for examples that suffer from #959 (e.g., `delay.x` example) or the benchmarks can be somehow generated in other way in this case.

### DIFF
--- a/xls/build_rules/xls_ir_rules.bzl
+++ b/xls/build_rules/xls_ir_rules.bzl
@@ -197,6 +197,7 @@ def _optimize_ir(ctx, src):
     opt_ir_tool = get_executable_from(get_xls_toolchain_info(ctx).opt_ir_tool)
     opt_ir_args = dict(ctx.attr.opt_ir_args)
     IR_OPT_FLAGS = (
+        "top",
         "ir_dump_path",
         "run_only_passes",
         "skip_passes",


proppy:
Note that this is blocking #994 (as #974 current epends on it)

proppy (approved):

proppy:
Removing the blocker flag as there is an acceptable workaround (using separate rules instead of `xls_dslx_verilog`).

proppy:
There is also a `dslx_top` argument available in the `xls_dslx_verilog` rule: https://google.github.io/xls/bazel_rules_macros/#xls_dslx_verilog-dslx_top (with the caveat that multi proc codegen isn't yet supported, see https://github.com/google/xls/issues/950)

mtdudek:
`dslx_top` argument is not enough (see #959),
Without passing `--top` to opt_ir, optimizer will remove correct top.

cdleary:
I think usually we put small representative tests in https://github.com/google/xls/tree/main/xls/build_rules/tests -- could you add one that shows use of the new flag (as discussed in this thread)? With a small test we can merge.

rw1nkler:
@cdleary It seems that this change has already been added by @hongted some time ago in https://github.com/google/xls/commit/5026f009d015eebfadf3ef72e27c919aa5ea0576.

PR #973, opened 2023-05-23
[DSLX] Add functions that calculate the length of an array to stdlib

rw1nkler:
This PR adds functions to calculate the length of an array to XLS stdlib.

CC @proppy

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -281,6 +281,30 @@ fn find_index_test() {
   ()
 }
 
+// Returns the length of an array of unsigned values
+pub fn array_length_unsigned<BITS: u32, ELEMS: u32>(array: uN[BITS][ELEMS]) -> u32 {
+  ELEMS
+}
+
+#[test]
+fn array_length_unsigned_test() {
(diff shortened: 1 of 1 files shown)


cdleary:
It's nice to have anything that can be done as a function in userspace be done that way, less stuff for the compiler to know / more being exercised without the special superpowers the frontend has access to. Looks great to have, thanks for adding!

cdleary:
Oh actually, I realize @proppy 's point now -- we don't have a type that's polymorphic on signedness, e.g. what you really want to write is:

```
fn array_size<BITS: u32, ELEMS: u32, xN: signedness>(a: xN[BITS][ELEMS]) -> u32 {
  ELEMS
}
```

That does currently require builtin superpowers. Sorry I misunderstood the point! I'll get you that quickly.

cdleary:
(Of course, feel free to use this impl in your implementation as a workaround until the builtin shows up.)

cdleary:
Really appreciate the PR, but I think https://github.com/google/xls/commit/ab140b54173a8288db6cc211a021be58f211e9bc should do what you're looking for slightly better given what we discussed in this thread. LMK if you have any questions/concerns!

rw1nkler:
I don't have any. Thank you for adding the built-in!

Issue #81, opened 2020-08-09
[DSLX] quickcheck support for parametric functions


cdleary:
cc @hmontero1205 

In #70 support for quickchecking non-parametric functions is introduced -- we'll want a way to let the quickcheck "driver" instantiate parametric functions with the ability to put constraints on the parametric values. One nice aspect of this is the quickcheck driver can use normal functions to test properties of normal values (e.g. filtering via `is_even` as a predicate on a given parametric `N: u32`) and then use that value to drive the instantiation. This will make the quickcheck less vectorizable, but we can vector-test some number of samples for any given parametric `N` chosen to get some batch efficiency.

cdleary:
Not currently on the "implement soon" radar so marking as long-term enhancement.

proppy:
This also seems to be transitive (i.e: quickcheck can't test any function that uses parametric functions)

PR #974, opened 2023-05-23
Add RLE encoder

### LINKED ISSUE #994: add DSLX sample for Run Length Encoding (RLE) encoder/decoder
See https://en.wikipedia.org/wiki/Run-length_encoding

This will help demonstrate how to implement streaming algorithms with DSLX procs.

rw1nkler:
This PR adds a simple implementation of the RLE Encoder ([Run Length Encoding](https://en.wikipedia.org/wiki/Run-length_encoding)). The introduced changes contain the RLE encoder, tests for the module, rules for Verilog generation, and rules to build IR and Verilog benchmarks.

The changes rely on other features added in separate PRs:
* https://github.com/google/xls/pull/964
* ~~https://github.com/google/xls/pull/973~~

Fixes #994 

CC @proppy

### DIFF
--- a/xls/modules/rle/BUILD
+++ b/xls/modules/rle/BUILD
@@ -1,0 +1,95 @@
+# Copyright 2023 The XLS Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build rules for XLS RLE Encoder implementation.
+
+load(
+    "//xls/build_rules:xls_build_defs.bzl",
+    "xls_dslx_library",
+    "xls_dslx_test",
+    "xls_dslx_ir",
+    "xls_ir_opt_ir",
+    "xls_ir_verilog",
+    "xls_benchmark_ir",
+)
+
+package(
+    default_applicable_licenses = ["//:license"],
+    default_visibility = ["//xls:xls_users"],
+    licenses = ["notice"],
+)
+
+xls_dslx_library(
+    name = "rle_dslx",
+    srcs = ["rle.x"],
+)
+
+xls_dslx_test(
+    name = "rle_dslx_test",
+    dslx_test_args = {
+        "compare": "none",
+    },
+    library = "rle_dslx",
+)
+
+xls_dslx_test(
+    name = "rle_dslx_ir_test",
+    dslx_test_args = {
+        "compare": "interpreter",
+    },
+    library = "rle_dslx",
+)
+
+xls_dslx_test(
+    name = "rle_dslx_jit_test",
+    dslx_test_args = {
+        "compare": "jit",
+    },
+    library = "rle_dslx",
+)
+
+xls_dslx_ir(
+    name = "rle_enc_ir",
+    dslx_top = "RunLengthEncoder32",
+    library = "rle_dslx",
+    ir_file = "rle_enc.ir",
+)
+
+xls_ir_opt_ir(
+    name = "rle_enc_opt_ir",
+    src = "rle_enc.ir",
+    top = "__rle__RunLengthEncoder32__RunLengthEncoder_0__2_32_next",
+)
+
+xls_ir_verilog(
+    name = "rle_enc_verilog",
+    src = ":rle_enc_opt_ir.opt.ir",
+    verilog_file = "rle_enc.v",
+    codegen_args = {
+        "module_name": "rle_enc",
+        "delay_model": "unit",
+        "pipeline_stages": "2",
+        "reset": "rst",
+        "use_system_verilog": "false",
+    },
+)
+
+xls_benchmark_ir(
+    name = "rle_enc_ir_benchmark",
+    src = ":rle_enc_opt_ir.opt.ir",
+    benchmark_ir_args = {
+        "pipeline_stages": "2",
+        "delay_model": "unit",
+    }
+)
--- a/xls/modules/rle/rle.x
+++ b/xls/modules/rle/rle.x
@@ -1,0 +1,449 @@
+// Copyright 2023 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file implements a parametric RLE encoder
+//
+// The encoder uses Run Length Encoding (RLE) to compress the input stream of
+// repeating symbols to the output stream that contains the symbols and
+// the number of its consequect occurrences in the input stream.
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+//                      ──────╥─────╥─────╥─────╥─────╥─────╥─────╥─────╥────
+// next evaluation      XXXXXX║ 0   ║ 1   ║ 2   ║ 3   ║ 4   ║ 5   ║ 6   ║ ...
+//                      ──────╨─────╨─────╨─────╨─────╨─────╨─────╨─────╨────
+//                      ──────╥───────────╥─────╥─────╥─────╥─────╥──────────
+// symbol               XXXXXX║ A         ║ B   ║XXXXX║ B   ║ C   ║XXXXXXXXXX
+// (input channel)      ──────╨───────────╨─────╨─────╨─────╨─────╨──────────
+// last                                   ┌─────┐           ┌─────┐
+// (input channel)      ──────────────────┘     └───────────┘     └──────────
+//                      ╥─────╥─────╥─────╥─────╥─────╥─────╥─────╥──────────
+// state.prev_symbol    ║ 0   ║ A   ║ A   ║ B   ║ 0   ║ B   ║ C   ║ 0
+// (set state value)    ╨─────╨─────╨─────╨─────╨─────╨─────╨─────╨──────────
+//                      ╥─────╥─────╥─────╥─────╥─────╥─────╥─────╥──────────
+// state.prev_count     ║ 0   ║ 1   ║ 2   ║ 1   ║ 0   ║ 1   ║ 1   ║ 0
+// (set state value)    ╨─────╨─────╨─────╨─────╨─────╨─────╨─────╨──────────
+//
+// do_send                                ┌───────────┐     ┌───────────┐
+//                      ──────────────────┘           └─────┘           └────
+//                      ──────────────────╥─────╥─────╥─────╥─────╥─────╥────
+// symbol, count        XXXXXXXXXXXXXXXXXX║ A,2 ║ B,1 ║XXXXX║ B,1 ║ C,1 ║XXXX
+// (output channel)     ──────────────────╨─────╨─────╨─────╨─────╨─────╨────
+// last                                         ┌─────┐           ┌─────┐
+// (output channel)     ────────────────────────┘     └───────────┘     └────
+
+import std
+
+// structure containing input data for an RLE encoder
+pub struct EncInData<SYMBOL_WIDTH: u32> {
+    symbol: bits[SYMBOL_WIDTH], // input symbol to compress
+    last: bool,                 // if this is the last symbol
+}
+
+// structure containing output data from an RLE encoder
+pub struct EncOutData<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
+    symbol: bits[SYMBOL_WIDTH], // compressed symbol
+    count: bits[COUNT_WIDTH],   // symbol count
+    last: bool,                 // if this is the last symbol
+}
+
+// structure to preserve the state of an RLE encoder
+struct RunLengthEncoderState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
+    // symbol from the previous RunLengthEncoder::next evaluation,
+    // valid if prev_count > 0
+    prev_symbol: bits[SYMBOL_WIDTH],
+    // symbol count from the previous RunLengthEncoder::next evaluation.
+    // zero means that the previous evaluation sent all the data and
+    // we start counting from the beginning
+    prev_count: bits[COUNT_WIDTH],
+    // flag indicating that the previous symbol was the last one
+    // in the transmission
+    prev_last: bool,
+}
+
+// RLE encoder implementation
+pub proc RunLengthEncoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
+    input_r: chan<EncInData<SYMBOL_WIDTH>> in;
+    output_s: chan<EncOutData<SYMBOL_WIDTH, COUNT_WIDTH>> out;
+
+    init {(
+        RunLengthEncoderState<SYMBOL_WIDTH, COUNT_WIDTH> {
+            prev_symbol: bits[SYMBOL_WIDTH]:0,
+            prev_count: bits[COUNT_WIDTH]:0,
+            prev_last: false,
+        }
+    )}
+
+    config (
+        input_r: chan<EncInData<SYMBOL_WIDTH>> in,
+        output_s: chan<EncOutData<SYMBOL_WIDTH, COUNT_WIDTH>> out,
+    ) {(input_r, output_s)}
+
+    next (tok: token, state: RunLengthEncoderState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+        let zero_input = EncInData {
+          symbol: bits[SYMBOL_WIDTH]:0,
+          last: false
+        };
+        let (input_tok, input) = recv_if(
+            tok, input_r, !state.prev_last, zero_input);
+
+        let prev_symbol_valid = state.prev_count != bits[COUNT_WIDTH]:0;
+        let symbol_differ = prev_symbol_valid && (
+            input.symbol != state.prev_symbol);
+        let overflow =
+            state.prev_count == std::unsigned_max_value<COUNT_WIDTH>();
+
+        let (symbol, count, last) = if (state.prev_last) {
+            (
+                bits[SYMBOL_WIDTH]:0,
+                bits[COUNT_WIDTH]:0,
+                false
+            )
+        } else if (symbol_differ || overflow) {
+            (
+                input.symbol,
+                bits[COUNT_WIDTH]:1,
+                input.last,
+            )
+        } else {
+            (
+                input.symbol,
+                state.prev_count + bits[COUNT_WIDTH]:1,
+                input.last,
+            )
+        };
+
+        let data = EncOutData {
+            symbol: state.prev_symbol,
+            count: state.prev_count,
+            last: state.prev_last
+        };
+
+        let do_send = state.prev_last || symbol_differ || overflow;
+        let data_tok = send_if(input_tok, output_s, do_send, data);
+
+        RunLengthEncoderState {
+            prev_symbol: symbol,
+            prev_count: count,
+            prev_last: last,
+        }
+    }
+}
+
+// RLE encoder specialization for the codegen
+proc RunLengthEncoder32 {
+
+    init {()}
+
+    config (
+        input_r: chan<EncInData<32>> in,
+        output_s: chan<EncOutData<32, 2>> out,
+    ) {
+        spawn RunLengthEncoder<u32:32, u32:2>(input_r, output_s);
+        ()
+    }
+
+    next (tok: token, state: ()) {
+        ()
+    }
+}
+
+// Tests
+
+const TEST_COMMON_SYMBOL_WIDTH = u32:32;
+// Make counter large enough so that it overflows only in overflow testcase.
+const TEST_COMMON_COUNT_WIDTH = u32:32;
+
+type TestCommonSymbol     = bits[TEST_COMMON_SYMBOL_WIDTH];
+type TestCommonCount      = bits[TEST_COMMON_COUNT_WIDTH];
+type TestCommonEncInData  = EncInData<TEST_COMMON_SYMBOL_WIDTH>;
+type TestCommonEncOutData =
+    EncOutData<TEST_COMMON_SYMBOL_WIDTH, TEST_COMMON_COUNT_WIDTH>;
+
+// Simple transaction without overflow
+const CountSymbolTestSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const CountSymbolTestCountWidth  = TEST_COMMON_COUNT_WIDTH;
+
+type CountSymbolTestStimulus   = TestCommonSymbol;
+type CountSymbolTestSymbol     = TestCommonSymbol;
+type CountSymbolTestCount      = TestCommonCount;
+type CountSymbolTestEncInData  = TestCommonEncInData;
+type CountSymbolTestEncOutData = TestCommonEncOutData;
+
+#[test_proc]
+proc RunLengthEncoderCountSymbolTest {
+  terminator: chan<bool> out;
+  enc_input_s: chan<CountSymbolTestEncInData> out;
+  enc_output_r: chan<CountSymbolTestEncOutData> in;
+
+  init {()}
+  config (terminator: chan<bool> out) {
+    let (enc_input_s, enc_input_r) = chan<CountSymbolTestEncInData>;
+    let (enc_output_s, enc_output_r) = chan<CountSymbolTestEncOutData>;
+
+    spawn RunLengthEncoder<CountSymbolTestSymbolWidth, CountSymbolTestCountWidth>(
+        enc_input_r, enc_output_s);
+    (terminator, enc_input_s, enc_output_r)
+  }
+  next (tok: token, state:()) {
+    let CountSymbolTestTestStimuli: CountSymbolTestStimulus[4] = [
+      CountSymbolTestStimulus:0xA, CountSymbolTestStimulus:0xA,
+      CountSymbolTestStimulus:0xA, CountSymbolTestStimulus:0xB,
+    ];
+    let tok = for ((counter, symbol), tok):
+                  ((u32, CountSymbolTestStimulus) , token)
+                  in enumerate(CountSymbolTestTestStimuli) {
+      let last = counter == (array_size(CountSymbolTestTestStimuli) - u32:1);
+      let stimulus = CountSymbolTestEncInData{symbol: symbol, last: last};
+      let tok = send(tok, enc_input_s, stimulus);
+      trace_fmt!("Sent {} stimuli, symbol: 0x{:x}, last: {}",
+        counter, stimulus.symbol, stimulus.last);
+      (tok)
+    }(tok);
+    let CountSymbolTestTestOutput:
+        (CountSymbolTestSymbol, CountSymbolTestCount)[2] = [
+      (CountSymbolTestSymbol:0xA, CountSymbolTestCount:0x3),
+      (CountSymbolTestSymbol:0xB, CountSymbolTestCount:0x1),
+    ];
+    let tok = for ((counter, (symbol, count)), tok):
+        ((u32, (CountSymbolTestSymbol, CountSymbolTestCount)) , token)
+        in enumerate(CountSymbolTestTestOutput) {
+      let last = counter == (array_size(CountSymbolTestTestOutput) - u32:1);
+      let expected = CountSymbolTestEncOutData{
+          symbol: symbol, count: count, last: last};
+      let (tok, enc_output) = recv(tok, enc_output_r);
+      trace_fmt!(
+        "Received {} pairs, symbol: 0x{:x}, count: {}, last: {}",
+        counter, enc_output.symbol, enc_output.count, enc_output.last
+      );
+      assert_eq(enc_output, expected);
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
+  }
+}
+
+// Transaction with counter overflow
+const OverflowSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const OverflowCountWidth  = u32:2;
+
+type OverflowStimulus   = TestCommonSymbol;
+type OverflowSymbol     = TestCommonSymbol;
+type OverflowCount      = bits[OverflowCountWidth];
+type OverflowEncInData  = TestCommonEncInData;
+type OverflowEncOutData =
+    EncOutData<OverflowSymbolWidth, OverflowCountWidth>;
+
+#[test_proc]
+proc RunLengthEncoderOverflowTest {
+  terminator: chan<bool> out;
+  enc_input_s: chan<OverflowEncInData> out;
+  enc_output_r: chan<OverflowEncOutData> in;
+
+  init {()}
+  config (terminator: chan<bool> out) {
+    let (enc_input_s, enc_input_r) = chan<OverflowEncInData>;
+    let (enc_output_s, enc_output_r) = chan<OverflowEncOutData>;
+
+    spawn RunLengthEncoder<OverflowSymbolWidth, OverflowCountWidth>(
+        enc_input_r, enc_output_s);
+    (terminator, enc_input_s, enc_output_r)
+  }
+  next (tok: token, state:()) {
+    let OverflowTestStimuli: OverflowStimulus[14] = [
+      OverflowStimulus:0xB, OverflowStimulus:0xB,
+      OverflowStimulus:0x1, OverflowStimulus:0xC,
+      OverflowStimulus:0xC, OverflowStimulus:0xC,
+      OverflowStimulus:0xC, OverflowStimulus:0xC,
+      OverflowStimulus:0xC, OverflowStimulus:0x3,
+      OverflowStimulus:0x3, OverflowStimulus:0x3,
+      OverflowStimulus:0x2, OverflowStimulus:0x2,
+    ];
+    let tok = for ((counter, symbol), tok):
+                  ((u32, OverflowStimulus) , token)
+                  in enumerate(OverflowTestStimuli) {
+      let last = counter == (
+          array_size(OverflowTestStimuli) - u32:1);
+      let stimulus = OverflowEncInData{symbol: symbol, last: last};
+      let tok = send(tok, enc_input_s, stimulus);
+      trace_fmt!("Sent {} stimuli, symbol: 0x{:x}, last: {}",
+        counter, stimulus.symbol, stimulus.last);
+      (tok)
+    }(tok);
+    let OverflowTestOutput:
+        (OverflowSymbol, OverflowCount)[6] = [
+      (OverflowSymbol:0xB, OverflowCount:0x2),
+      (OverflowSymbol:0x1, OverflowCount:0x1),
+      (OverflowSymbol:0xC, OverflowCount:0x3),
+      (OverflowSymbol:0xC, OverflowCount:0x3),
+      (OverflowSymbol:0x3, OverflowCount:0x3),
+      (OverflowSymbol:0x2, OverflowCount:0x2),
+    ];
+    let tok = for ((counter, (symbol, count)), tok):
+        ((u32, (OverflowSymbol, OverflowCount)) , token)
+        in enumerate(OverflowTestOutput) {
+      let last = counter == (array_size(OverflowTestOutput) - u32:1);
+      let expected = OverflowEncOutData{
+          symbol: symbol, count: count, last: last};
+      let (tok, enc_output) = recv(tok, enc_output_r);
+      trace_fmt!(
+        "Received {} pairs, symbol: 0x{:x}, count: {}, last: {}",
+        counter, enc_output.symbol, enc_output.count, enc_output.last
+      );
+      assert_eq(enc_output, expected);
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
+  }
+}
+
+// Check that RLE encoder will create 2 `last` output packets,
+// when 2 `last` input packets were consumed.
+const LastAfterLastSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const LastAfterLastCountWidth  = TEST_COMMON_COUNT_WIDTH;
+
+type LastAfterLastStimulus   = TestCommonEncInData;
+type LastAfterLastSymbol     = TestCommonSymbol;
+type LastAfterLastCount      = TestCommonCount;
+type LastAfterLastEncInData  = TestCommonEncInData;
+type LastAfterLastEncOutData = TestCommonEncOutData;
+type LastAfterLastOutput     = TestCommonEncOutData;
+
+#[test_proc]
+proc RunLengthEncoderLastAfterLastTest {
+  terminator: chan<bool> out;
+  enc_input_s: chan<LastAfterLastEncInData> out;
+  enc_output_r: chan<LastAfterLastEncOutData> in;
+
+  init {()}
+  config (terminator: chan<bool> out) {
+    let (enc_input_s, enc_input_r) = chan<LastAfterLastEncInData>;
+    let (enc_output_s, enc_output_r) = chan<LastAfterLastEncOutData>;
+
+    spawn RunLengthEncoder<LastAfterLastSymbolWidth, LastAfterLastCountWidth>(
+        enc_input_r, enc_output_s);
+    (terminator, enc_input_s, enc_output_r)
+  }
+  next (tok: token, state:()) {
+    let LastAfterLastTestStimuli: LastAfterLastStimulus[2] = [
+      LastAfterLastStimulus {symbol: LastAfterLastSymbol:0x1, last: true},
+      LastAfterLastStimulus {symbol: LastAfterLastSymbol:0x1, last: true},
+    ];
+    let tok = for ((counter, stimuli), tok):
+        ((u32, LastAfterLastStimulus) , token)
+        in enumerate(LastAfterLastTestStimuli) {
+      let tok = send(tok, enc_input_s, stimuli);
+      trace_fmt!("Sent {} transactions, symbol: 0x{:x}, last: {}",
+        counter, stimuli.symbol, stimuli.last);
+      (tok)
+    }(tok);
+    let LastAfterLastTestOutput: LastAfterLastOutput[2] = [
+      LastAfterLastOutput {
+        symbol: LastAfterLastSymbol:0x1,
+        count: LastAfterLastCount:0x1,
+        last:true},
+      LastAfterLastOutput {
+        symbol: LastAfterLastSymbol:0x1,
+        count: LastAfterLastCount:0x1,
+        last:true},
+    ];
+    let tok = for ((counter, expected), tok):
+        ((u32, LastAfterLastOutput) , token)
+        in enumerate(LastAfterLastTestOutput) {
+      let (tok, enc_output) = recv(tok, enc_output_r);
+      trace_fmt!(
+        "Received {} pairs, symbol: 0x{:x}, count: {}, last: {}",
+        counter, enc_output.symbol, enc_output.count, enc_output.last
+      );
+      assert_eq(enc_output, expected);
+
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
+  }
+}
+
+// Check overflow condition trigger on packet with `last`
+const OverflowWithLastSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const OverflowWithLastCountWidth  = u32:2;
+
+type OverflowWithLastStimulus   = TestCommonSymbol;
+type OverflowWithLastSymbol     = TestCommonSymbol;
+type OverflowWithLastCount      =
+    bits[OverflowWithLastCountWidth];
+type OverflowWithLastEncInData  = TestCommonEncInData;
+type OverflowWithLastEncOutData =
+    EncOutData<OverflowWithLastSymbolWidth,
+               OverflowWithLastCountWidth>;
+
+#[test_proc]
+proc RunLengthEncoderOverflowWithLastTest {
+  terminator: chan<bool> out;
+  enc_input_s: chan<OverflowWithLastEncInData> out;
+  enc_output_r: chan<OverflowWithLastEncOutData> in;
+
+  init {()}
+  config (terminator: chan<bool> out) {
+    let (enc_input_s, enc_input_r) = chan<OverflowWithLastEncInData>;
+    let (enc_output_s, enc_output_r) =
+        chan<OverflowWithLastEncOutData>;
+
+    spawn RunLengthEncoder<OverflowWithLastSymbolWidth,
+                 OverflowWithLastCountWidth>(
+        enc_input_r, enc_output_s);
+    (terminator, enc_input_s, enc_output_r)
+  }
+  next (tok: token, state:()) {
+    let OverflowWithLastTestStimuli: OverflowWithLastStimulus[4] = [
+      OverflowWithLastStimulus:0xC, OverflowWithLastStimulus:0xC,
+      OverflowWithLastStimulus:0xC, OverflowWithLastStimulus:0xC,
+    ];
+    let tok = for ((counter, symbol), tok):
+                  ((u32, OverflowWithLastStimulus) , token)
+                  in enumerate(OverflowWithLastTestStimuli) {
+      let last = counter == (
+          array_size(OverflowWithLastTestStimuli) - u32:1);
+      let stimulus = OverflowWithLastEncInData{symbol: symbol, last: last};
+      let tok = send(tok, enc_input_s, stimulus);
+      trace_fmt!("Sent {} stimuli, symbol: 0x{:x}, last: {}",
+        counter, stimulus.symbol, stimulus.last);
+      (tok)
+    }(tok);
+    let OverflowWithLastTestOutput:
+        (OverflowWithLastSymbol, OverflowWithLastCount)[2] = [
+      (OverflowWithLastSymbol:0xC, OverflowWithLastCount:0x3),
+      (OverflowWithLastSymbol:0xC, OverflowWithLastCount:0x1),
+    ];
+    let tok = for ((counter, (symbol, count)), tok):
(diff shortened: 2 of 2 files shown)


xls/modules/rle/BUILD
```diff
@@ -0,0 +1,90 @@
...
+    library = "rle_dslx",
+    opt_ir_args = {
+        "top": "__rle__RLEEnc32__RLEEnc_0__2_32_next",
+    },
+    codegen_args = {
+        "module_name": "rle_enc",
+        "delay_model": "unit",
+        "pipeline_stages": "2",
+        "reset": "rst",
+        "use_system_verilog": "false",
```
proppy, line 74: curious if the system verilog output was working?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+
+// structure containing output data from an RLE encoder
+struct EncOutData<SYMB_WIDTH: u32, CNT_WIDTH: u32> {
+    symb: bits[SYMB_WIDTH], // symbol
+    cnt: bits[CNT_WIDTH],   // symbol counter
+}
+
+// structure to preserve the state of an RLE encoder for the next iteration
+pub struct RLEEncState<SYMB_WIDTH: u32, CNT_WIDTH: u32> {
+    prev: EncOutData<SYMB_WIDTH, CNT_WIDTH>, // previous symbol and symbol counter
```
proppy, line 31: I think we could simply name this `count`?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+    flush: bool,                             // if flush was requested
```
proppy, line 32: I'm curious about the semantic of the `flush` and why it needs to be kept as a state, could you add some rational in the PR description or in the code comment?

I would have naively thought that the encoder would emit on every symbol change, but it seems that you are giving more control with that flag.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+        let _ = join(flush_tok, data_tok);
+
+        let prev_symb_valid = state.prev.cnt != bits[CNT_WIDTH]:0;
+        let symb_differ = prev_symb_valid && (symb != state.prev.symb);
+        let overflow = state.prev.cnt == std::unsigned_max_value<CNT_WIDTH>();
+
+        let (do_send, next_state) = if (state.flush) {
+            (true, RLEEncState {
+                prev: EncOutData<SYMB_WIDTH, CNT_WIDTH> {
+                    symb: bits[SYMB_WIDTH]:0,
```
proppy, line 69: isn't this effectively loosing the current symbol if different? or missing one count if the same?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+    next (tok: token, state: RLEEncState<SYMB_WIDTH, CNT_WIDTH>) {
+        let (flush_tok, flush) = recv_if(tok, flush_r, !state.flush, false);
+        let (data_tok, symb) = recv_if(tok, symb_r, !state.flush, bits[SYMB_WIDTH]:0);
+        let _ = join(flush_tok, data_tok);
+
+        let prev_symb_valid = state.prev.cnt != bits[CNT_WIDTH]:0;
+        let symb_differ = prev_symb_valid && (symb != state.prev.symb);
+        let overflow = state.prev.cnt == std::unsigned_max_value<CNT_WIDTH>();
+
+        let (do_send, next_state) = if (state.flush) {
```
proppy, line 66: I wonder if this would be more readable to separate the state update logic from the communication one?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+// The encoder uses Run Length Encoding (RLE) to compress the input stream
+// with symbols (data to compress). The output contains packets of data with
+// the symbols and the number of their occurrences in the input stream.
+
+import std
+
+// structure containing output data from an RLE encoder
+struct EncOutData<SYMB_WIDTH: u32, CNT_WIDTH: u32> {
+    symb: bits[SYMB_WIDTH], // symbol
+    cnt: bits[CNT_WIDTH],   // symbol counter
```
proppy, line 26: `count`?

xls/modules/rle/BUILD
rw1nkler: It works fine. However, I also wanted to add a simulation with Iverilog simulator. As far as I know, it does not support SystemVerilog. I can add a separate target that generates SystemVerilog if that is preferred.

xls/modules/rle/rle.x
rw1nkler: It should not because we don't read new symbols when `flush` is asserted. I will extend tests to show multiple uses of `flush`

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+    config (
+        symb_r: chan<bits[SYMB_WIDTH]> in,
+        flush_r: chan<bool> in,
+        data_s: chan<EncOutData<SYMB_WIDTH, CNT_WIDTH>> out,
+    ) {(symb_r, flush_r, data_s)}
+
+    next (tok: token, state: RLEEncState<SYMB_WIDTH, CNT_WIDTH>) {
+        let (flush_tok, flush) = recv_if(tok, flush_r, !state.flush, false);
+        let (data_tok, symb) = recv_if(tok, symb_r, !state.flush, bits[SYMB_WIDTH]:0);
+        let _ = join(flush_tok, data_tok);
```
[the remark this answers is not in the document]
rw1nkler: Yes, we should. Thanks!

xls/modules/rle/rle.x
rw1nkler: A request on the `flush` channel forces the encoder to dump all data to the output stream. Both the `flush` and the channel with input symbols should be synchronized. The information about the flush request is stored in the state of the encoder because it is used to skip reading symbols in the next cycle after the request. That "skipped" cycle is used to send old data (about the last symbol when flush occurred). 

This implementation was chosen because it is currently impossible to send data two times in a row to the same channel in one evaluation of the `next` function.

The typical use of this channel is to set flush after all symbols have been sent to force the encoder to output a final result. Otherwise, the last symbol will be "stuck" in the encoder.

Of course, I will update the code with the proper explanation about the `flush` channel and the `flush` in the state.

xls/modules/rle/rle.x
rw1nkler: For me, it was more obvious to think about it in this way, for example:

* if the `flush` was requested we want to send the data and our next state is ABC
* if we have an overflow or change of the symbol, this means that we have to send the data and our next state is XYZ

But of course, I can separate that if you want

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,253 @@
...
+//
+// The encoder uses Run Length Encoding (RLE) to compress the input stream
+// with symbols (data to compress). The output contains packets of data with
+// the symbols and the number of their occurrences in the input stream.
+
+import std
+
+// structure containing output data from an RLE encoder
+struct EncOutData<SYMB_WIDTH: u32, CNT_WIDTH: u32> {
+    symb: bits[SYMB_WIDTH], // symbol
```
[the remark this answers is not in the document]
rw1nkler: fixed

xls/modules/rle/rle.x
rw1nkler: fixed

xls/modules/rle/rle.x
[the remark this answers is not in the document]
rw1nkler: fixed

xls/modules/rle/rle.x
rw1nkler: Renamed to `COUNT_WIDTH`

xls/modules/rle/rle.x
rw1nkler: I've extended the tests to use flush multiple times

xls/modules/rle/BUILD
proppy: Make sense, filed https://github.com/google/xls/issues/980 and https://github.com/hdl/bazel_rules_hdl/issues/163 to discuss how we could streamline this.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+// read at the same time. The first one contains the symbol to encode,
+// while the second contains request for dumping data from the encoder.
+// Typical use of the flush_r channel is to set it to true at the end of the
+// transmission (together with the last symbol send to the symbol_r channel),
+// to obtain the last packet of data from the encoder.
+//
+// The information about the flush request is stored in the state of the
+// encoder because it is used to skip reading symbols in the next cycle after
+// the request. That "skipped" cycle is used to send already proceeded data
+//over the data_s channel. This behavior was presented on the waveform below:
```
proppy, line 32: missing space.

xls/modules/rle/rle.x
proppy: > This implementation was chosen because it is currently impossible to send data two times in a row to the same channel in one evaluation of the next function.

Are you referring to https://github.com/google/xls/blob/30f192cb091f3f36e459bf4738b4447543f9177a/xls/ir/verifier.cc#L1604 ?

Curious if https://github.com/google/xls/blob/main/docs_src/design_docs/legalize_multiple_channel_ops_per_channel.md would help /cc @grebe 

Interestingly it seems that it doesn't currently fails the interpreter, nor IR conversion, but IR optimization, see:
https://colab.research.google.com/gist/proppy/9be8ce46181b007d1cbe2a4329923623/dslx-proc-playground.ipynb

xls/modules/rle/rle.x
proppy: Sorry about the confusion, I was referring the `prev` field:  describe it as something used to track the count for the current symbol (`counter`) rather that `prev` might be more informative.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+    )}
+
+    config (
+        symbol_r: chan<bits[SYMBOL_WIDTH]> in,
+        flush_r: chan<bool> in,
+        data_s: chan<EncOutData<SYMBOL_WIDTH, COUNT_WIDTH>> out,
+    ) {(symbol_r, flush_r, data_s)}
+
+    next (tok: token, state: RLEEncState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+        let (flush_tok, flush) = recv_if(tok, flush_r, !state.flush, false);
```
proppy, line 100: couldn't that `recv` be also non blocking, that would allow producer not to send `flush: false` on every tick?

If we do want to enforce it, maybe we could make the `chan` type a struct instead so that we're guarantee by construct for the two to be always consistently received together (and remove the note about `(should be read together with symbol_r)`).

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+        let (symbol_tok, symbol) = recv_if(tok, symbol_r, !state.flush, bits[SYMBOL_WIDTH]:0);
+        let recv_tok = join(flush_tok, symbol_tok);
+
+        let prev_symbol_valid = state.prev.count != bits[COUNT_WIDTH]:0;
+        let symbol_differ = prev_symbol_valid && (symbol != state.prev.symbol);
+        let overflow = state.prev.count == std::unsigned_max_value<COUNT_WIDTH>();
+
+        let next_state = if (state.flush) {
+            RLEEncState {
```
proppy, line 109: maybe we could have the if simply return `EncOutData` (or `symbol, count`) since `flush` is the same across all branches.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+                prev: EncOutData<SYMBOL_WIDTH, COUNT_WIDTH> {
+                    symbol: symbol,
+                    count: bits[COUNT_WIDTH]:1,
+                },
+                flush: flush,
+            }
+        } else {
+            RLEEncState {
+                prev: EncOutData<SYMBOL_WIDTH, COUNT_WIDTH> {
+                    symbol: symbol,
```
proppy, line 127: even if that's equivalent, I think `state.prev.symbol` would be more readable.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+
+// Tests
+
+type SendDataPair = (bits[32], bool);     // (symbol, flush)
+type RecvDataPair = (bits[32], bits[2]);  // (symbol, count)
+
+const TEST_SEND_LEN = u32:20;
+const TEST_RECV_LEN = u32:10;
+
+const TEST_SEND_LOOKUP = SendDataPair[TEST_SEND_LEN]:[
```
proppy, line 167: if you want to do table-based testing, you might consider interleaving input and output in a `struct` so that it can read more fluently.

xls/modules/rle/rle.x
rw1nkler: Yes, they are almost equivalent. Except from the first iteration when the `state.prev.symbol` is `0`. So actually using `symbol` does not require any logic to distinguish between the first and any other iteration.

xls/modules/rle/rle.x
rw1nkler: fixed

xls/modules/rle/rle.x
rw1nkler: Changed to `(symbol, count)`

xls/modules/rle/rle.x
rw1nkler: fixed by using a single EncInData

xls/modules/rle/rle.x
rw1nkler: Yes, that is the problem that I encountered.

xls/modules/rle/rle.x
rw1nkler: The problem lies in the fact that not every send request has a corresponding receive request. So in general there will be empty spaces in the struct that collect the transactions, plus some sort of flag indicating that the receive is expected would be required. 

At the same time, there is a need to check for the send that has a flush flag set. I'm not sure whether having one struct for everything could simplify the current mechanism without providing much more complexity in the structure of data. 

But maybe I don't see something obvious here. Do you have anything specific in mind?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,314 @@
...
+
+// This file implements a parametric RLE encoder
+//
+// The encoder uses Run Length Encoding (RLE) to compress the symbols from
+// the input stream. The output contains packets of data with the symbols and
+// the number of their occurrences in the input stream.
+//
+// The RLE encoder exposes one input and one output channel. The input channel
+// contains the symbol to compress and flush request. The request is used to
+// dump all the data from the encoder. A typical use of the flush field is
```
proppy, line 23: I wonder if it would make sense to make the "typical use" the "only supported use" in order simplify the protocol, if we only use flash to end (or start the transmission), maybe we could simply ignore the `symbol` part of the packet?

That would require the user to do always send the last `symbol` before the `flash` cycle, and remove the need for saving the `prev_flush` state?

wdyt?

proppy: Looking at the [AXIS stream protocol specification](https://developer.arm.com/documentation/ihi0051/latest/) (is that the right version to look at?)

It does seem to allow `TLAST` with no accompanying data, see:

> 2.6.1 Transfer with zero data or position bytes
> A transfer can have TLAST asserted but contain no data or position bytes. This can be used to:
> • Indicate the end of a packet when there are no more data or position bytes to transmit.
> • Push through any data that is held in intermediate buffers.
> • Complete an operation at an end-point that is expecting a TLAST at the end of a packet.

Some key differences I see with an empty delimiter packet:
a. the proc is guaranted to receive packet on each tick
b. the receiver can be guaranted that the last `EncOutData` will be emitted in the same clock cycle as `flush`
c. the producer potentially needs to buffer one `symbol` in case the flush needs to happen without interrupting the stream

(a) and (b) sound desirable, (c) seems like a limitation but it seems that it would also happen with the current implementation: as the proc would not attempt to read when "acting" on the `flush` state and I'm not sure how common it would be: naively I would though that the need to flush would be strongly associated with some boundary being detected in the input stream.

xls/modules/rle/rle.x
rw1nkler: Having the `last` signal exposed as an output as described in https://github.com/google/xls/pull/974#discussion_r1209390982, simplifies also the checks of the data received from the encoder.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+        } else {
+            (
+                input.symbol,
+                state.prev_count + bits[COUNT_WIDTH]:1,
+                input.last,
+            )
+        };
+
+        let do_send = state.prev_last || symbol_differ || overflow;
+        if do_send {
```
proppy, line 123: `send_if`?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+
+        let prev_data_valid = state.recv_count != u32:0;
+        let data_repeats = prev_data_valid && (state.prev.symbol == enc_output.symbol);
+        let max_count = std::unsigned_max_value<TEST_COUNT_WIDTH>();
+        let overflow = prev_data_valid && state.prev.count == max_count;
+        let last = prev_data_valid && state.prev.last;
+
+        let _ = if data_repeats {
+            let _ = assert_eq(overflow || last, true);
+        } else {()};
```
proppy, line 286: is there something we can assert in the other case?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+
+        // check the total count after the last expected receive matches
+        // with the expected value
+
+        let exp_count_reached = recv_count == array_size(TEST_RECV_LOOKUP);
+        let exp_total_reached = total_count == array_size(TEST_SEND_LOOKUP);
+        let _ = if exp_count_reached {
+            let _ = assert_eq(exp_total_reached, true);
+            let _ = send(recv_tok, terminator, true);
+        } else {()};
```
proppy, line 296: is there something we can assert in the other case?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,306 @@
...
+        RLEEncState {
+            prev_symbol: symbol,
+            prev_count: count,
+            prev_last: last,
+        }
+    }
+}
+
+// RLE encoder specialization for the codegen
+proc RLEEnc32 {
```
proppy, line 141: is the specialization necessary? or can we just use `type` definition on procs? (if not, maybe we should log an issue to allow it?)

xls/modules/rle/rle.x
rw1nkler: fixed

xls/modules/rle/rle.x
rw1nkler: Added more assertions related to the number of received transactions

xls/modules/rle/rle.x
rw1nkler: A similar approach is used in the `delay.x` example:
https://github.com/google/xls/blob/main/xls/examples/delay.x#L202

The purpose of this specialization is to be able to generate the Verilog out of it, because currently it seems to be impossible to generate Verilog out of the parametric proc:
```
3 ir_converter_main.cc:163] Check failed: ::absl::OkStatus() == (status) (OK vs. NOT_FOUND: Top-level type info not found for proc "RLEEnc".)
```

xls/modules/rle/rle.x
rw1nkler: If the data repeats, there are no specials cases to handle related to count of the symbols or whether the packet is at or not. However, I added a general assert which check if the received count is greater than zero and less or equal to maximal possible count.

xls/modules/rle/rle.x
rw1nkler: Is the current solution acceptable for you @proppy?

xls/modules/rle/rle.x
proppy: Oh I see, this does sound different from #959, it would be nice to have a way to alias procs like we do it with other type, see https://github.com/google/xls/blob/main/xls/examples/delay.x#L202.

Filed #998

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,322 @@
...
+        }
+    )}
+
+    config (
+        input_r: chan<EncInData<SYMBOL_WIDTH>> in,
+        output_s: chan<EncOutData<SYMBOL_WIDTH, COUNT_WIDTH>> out,
+    ) {(input_r, output_s)}
+
+    next (tok: token, state: RLEEncState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+        let zero_input = EncInData { symbol: bits[SYMBOL_WIDTH]:0, last: false };
```
proppy, line 95: nit: wonder if we could move this as a top level `const`, wdyt?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,322 @@
...
+        let max_count = std::unsigned_max_value<TEST_COUNT_WIDTH>();
+
+        let _ = assert_eq(enc_output.count > TestCount:0, true);
+        let _ = assert_eq(enc_output.count <= max_count, true);
+
+        // if the symbol repeats, check that the symbol counter has reached
+        // its maximum value or if the symbol was last in the transaction
+
+        let prev_data_valid = state.recv_count != u32:0;
+        let data_repeats = prev_data_valid && (state.prev.symbol == enc_output.symbol);
```
proppy, line 286: nit: should we wrap all of this in a big `if prev_data_valid`? as it doesn't seems that the other condition are used somewhere else.

xls/modules/rle/rle.x
proppy: what about:
```
assert_eq(overflow || last, data_repeats);
```
would that work?

proppy:
> Please let me know when you are satisfied with the changes. 

Yes, please rebase and remove the `draft` label so that other people on the team can review.

Thanks for working incrementally on this!

xls/modules/rle/rle.x
rw1nkler: fixed

xls/modules/rle/rle.x
rw1nkler: This is a parametric struct, I tried to set `zero_input` as a parametric value of `RLEEnc`, but it seems that a parametric value cannot be of parametric struct type. I may open a separate issue for that, unless it is an expected behavior.

proppy:
Confirmed with @meheff, @cdleary and @hzeller that the current handling of the `_last` symbol would correspond more or less to what the toolchain would generate if @grebe's https://github.com/google/xls/blob/main/docs_src/design_docs/legalize_multiple_channel_ops_per_channel.md was implemented and multiple send per tick were allowed.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,324 @@
...
+        // if the symbol repeats, check that the symbol counter has reached
+        // its maximum value or if the symbol was last in the transaction
+
+        let prev_data_valid = state.recv_count != u32:0;
+        let _ = if prev_data_valid {
+            let data_repeats = state.prev.symbol == enc_output.symbol;
+            let overflow = state.prev.count == max_count;
+            let last = state.prev.last;
+
+            let _ = if data_repeats {
```
proppy, line 291: wouldn't `assert_eq(data-repeats, overflow || last)` do the same?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,324 @@
...
+
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH = u32:2;
+
+type TestSymbol = bits[TEST_SYMBOL_WIDTH];
+type TestCount = bits[TEST_COUNT_WIDTH];
+type TestEncInData = EncInData<TEST_SYMBOL_WIDTH>;
+type TestEncOutData = EncOutData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+
+const TEST_SEND_LOOKUP = TestEncInData[20]:[
```
proppy, line 167: I'd be curious to see what the test would looks like if all those payload were inlined in the test itself instead?

The code might be a little more verbose, but the flow between data send and assertion on recv'ed message would be more readable (currently you have to make an interection between the two tables to figure that out).

proppy: @meheff also suggested that you might have more control if testing those in C++.

xls/modules/rle/rle.x
mtdudek: Not exactly, as we want to skip assert entirely if data_didn't repeat.
It can be changed to `assert_eq(data_repeats && (overflow || last), data_repeats)`,
but I'm not sure if this will make this code more readable.

xls/modules/rle/rle.x
mtdudek: It is possible, but test_proc's next function would be enormous.
It would also remove ability to quickly extend test's inputs and outputs.

As for C++ tests, on one hand,it would give as access to procs state and JIT.
On the other hand, it would remove a non-trivial test_proc example,
that uses a separate proc for stimuli and different proc for scoring DUT behavior.

xls/build_rules/xls_ir_rules.bzl
```diff
@@ -197,6 +197,7 @@ def _optimize_ir(ctx, src):
     opt_ir_tool = get_executable_from(get_xls_toolchain_info(ctx).opt_ir_tool)
     opt_ir_args = dict(ctx.attr.opt_ir_args)
     IR_OPT_FLAGS = (
+        "top",
```
proppy, line 200: can we refactor this to use separate rules or the `dslx_top` argument, to decouple this work from https://github.com/google/xls/pull/964?

xls/modules/rle/rle.x
proppy: > It is possible, but test_proc's next function would be enormous.

I'm not sure if it would be that enormous, as you would be able to leverage for loops for sending multiple packets and assert each packet being received right after each group of data.

> As for C++ tests, on one hand,it would give as access to procs state and JIT 
> On the other hand, it would remove a non-trivial test_proc example,
> that uses a separate proc for stimuli and different proc for scoring DUT behavior.

Just something (C++ tests) to keep in mind for more complicated test (not necessarly this one).

Note that I do agree that having non-trivial dslx test is generally useful and that it's interesting to demonstrate using separate proc for stimuli and assertion. For that particular test, we have a tight correlation both temporal and in the data being tested between the stimuli and assertion: so I feel that we loose a bit of the this relation  with the current data-driven design.

xls/modules/rle/rle.x
proppy: I do agree that the current assertion condition feels weird, and I'm not a fan of having multiple nested empty else either as it makes the control flow difficult to follow.

I wonder if we could encode this as a state machine w/ https://google.github.io/xls/dslx_reference/#match-expression instead? (or maybe this consideration become moot if we have more explicit test as suggested earlier w/ a fat `test_proc`?)

xls/modules/rle/rle.x
proppy: > So I've tried to create rather elegant solution using for over TEST_SEND_LOOKUP and recv_non_blocking to get response from RLEEnc

What happend if you interleave "for over TEST_SEND_LOOKUP + send" and "blocking recv"?

mtdudek: It can be done, but it will use recv_if.
I'll have to specify when I expect to receive data from the encoder.

I must also handle double sends from RLE.
Array will by of TEST_SEND_LEN length, and store values: 0,1,2, so it can be concatenated with TEST_SEND array.

I'll push commit with such change.

mtdudek: There is one more thing, I use multi process approach in other PR, and in some test cases I directly reuse `RLEEncTestSender`

mtdudek: I've changed thigh test to interleave send with blocking recv.
I'm not sure if this is better/more readable than one with multiple procs.

proppy: ah, sorry for the missunderstanding.

What I meant by interleaving was more something like this:
```
let sent_packets: u8[4] = [0xA, 0xA, 0xA, 0xB];
let tok = for ((i, p), (tok)) in enumerate(sent_packets) {
   let tok = send(tok, enc_input_s, p);
   tok
}(tok);
let expected_packets (u8, u32)[2] = [(0xA, 3), (0xB, 1)];
let tok = for ((i, p), (tok)) in enumerate(expected_packets) {
   let (tok, output) = recv(tok, enc_output_r);
   assert_eq(output.symbol, p[0]);
   assert_eq(output.count, p[1]);
   tok
}(tok);
```
would an approach like this work? The advantage I see is that you can directly correlate test input and output for each encoding "sessions".

mtdudek: Also I think I would be nice if DSLX would allow for passing variable reference to test_procs, something like:
```
#[test_proc_def] 
proc Tester<A> ....
```
and
```
#[test_proc<A=some_constant_variable>]
Tester
```
This way universal Tester could be written, and run multiple times with different payload.

proppy: I think that's could be interesting to discuss this as part of https://github.com/google/xls/issues/81, in a way a very hard constraint on quickcheck would be to have a list of data to test with.

See also what python has for https://docs.pytest.org/en/7.1.x/example/parametrize.html

mtdudek: I agree, so any way I'll change RLEEncTightTester to look as you suggested and keep old test, as I'm using parts of it in other PR.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+        let (enc_input_s, enc_input_r) = chan<TestEncInData>;
+        let (enc_output_s, enc_output_r) = chan<TestEncOutData>;
+
+        spawn RLEEnc<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>(enc_input_r, enc_output_s);
+        (terminator, enc_input_s, enc_output_r)
+    }
+    next (tok: token, state:()) {
+
+        // Clear on `last`
+        let trans_send: TestEncInData[4] = [
```
proppy, line 189: can we simplify the structure into a array? and craft the `TestEncInData` on the fly? (we set the the `last` flag when we're on the last element)

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            let (tok, enc_output) = recv(tok, enc_output_r);
+            let _ = trace_fmt!(
+                "Received {} transactions, symbol: 0x{:x}, count: {}, last: {}",
+                counter, enc_output.symbol, enc_output.count, enc_output.last
+            );
+            let _ = assert_eq(enc_output, data_out);
+            let max_count = std::unsigned_max_value<TEST_COUNT_WIDTH>();
+
+            let _ = assert_eq(enc_output.count > TestCount:0, true);
+            let _ = assert_eq(enc_output.count <= max_count, true);
```
proppy, line 215: is that really necessary to test this? (since we already test against the expected count)

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            let data_repeats = prev_symbol == enc_output.symbol;
+            let _ = assert_eq(data_repeats && can_repeat, data_repeats);
```
proppy, line 217: I don't think this can even fail for the given test data: we never got over max_count, and we never iterate past the last symbol.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            ((tok, (enc_output.count == max_count)|| enc_output.last, enc_output.symbol))
+        }((tok, true, TestSymbol:0));
+
+        // Counter Overflow
+        let trans_send: TestEncInData[14] = [
+            TestEncInData {symbol: TestSymbol:0xB, last: false},
```
proppy, line 223: same comment about simplifying the type.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            TestEncInData {symbol: TestSymbol:0x2, last: true},
+        ];
+        let tok = for ((counter, data_in), tok): ((u32, TestEncInData) , token) in enumerate(trans_send) {
+            let tok = send(tok, enc_input_s, data_in);
+            let _ = trace_fmt!("Sent {} transactions, symbol: 0x{:x}, last: {}",
+                counter, data_in.symbol, data_in.last);
+            (tok)
+        }(tok);
+        let trans_recv: TestEncOutData[6] = [
+            TestEncOutData {symbol: TestSymbol:0xB, count: TestCount:0x2, last:false},
```
proppy, line 245: same comment about simplifying the type.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+        let (tok, _, _) = for ((counter, data_out), (tok, can_repeat, prev_symbol)): ((u32, TestEncOutData) , (token, bool, TestSymbol)) in enumerate(trans_recv) {
+            let (tok, enc_output) = recv(tok, enc_output_r);
+            let _ = trace_fmt!(
+                "Received {} transactions, symbol: 0x{:x}, count: {}, last: {}",
+                counter, enc_output.symbol, enc_output.count, enc_output.last
+            );
+            let _ = assert_eq(enc_output, data_out);
+            let max_count = std::unsigned_max_value<TEST_COUNT_WIDTH>();
+
+            let _ = assert_eq(enc_output.count > TestCount:0, true);
```
proppy, line 261: same comment about test that can never fail.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            let _ = assert_eq(enc_output.count <= max_count, true);
+            let data_repeats = prev_symbol == enc_output.symbol;
+            let _ = assert_eq(data_repeats && can_repeat, data_repeats);
```
proppy, line 264: same comment about test that can never fail.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            ((tok, (enc_output.count == max_count)|| enc_output.last, enc_output.symbol))
+        }((tok, true, TestSymbol:0));
+
+        // `last` after `last` check
+        let trans_send: TestEncInData[2] = [
```
proppy, line 269: find to keep this as it here, since we're testing a specific corner case.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            TestEncInData {symbol: TestSymbol:0x1, last: true},
+            TestEncInData {symbol: TestSymbol:0x2, last: true},
```
proppy, line 271: should we use the same `TestSymbol` to make the test more relevant? (that way we can demonstrate that the count will be `1` for both)

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+        let (tok, _, _) = for ((counter, data_out), (tok, can_repeat, prev_symbol)): ((u32, TestEncOutData) , (token, bool, TestSymbol)) in enumerate(trans_recv) {
+            let (tok, enc_output) = recv(tok, enc_output_r);
+            let _ = trace_fmt!(
+                "Received {} transactions, symbol: 0x{:x}, count: {}, last: {}",
+                counter, enc_output.symbol, enc_output.count, enc_output.last
+            );
+            let _ = assert_eq(enc_output, data_out);
+            let max_count = std::unsigned_max_value<TEST_COUNT_WIDTH>();
+
+            let _ = assert_eq(enc_output.count > TestCount:0, true);
```
proppy, line 292: should we have a separate test for the overflow?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            let _ = assert_eq(enc_output.count <= max_count, true);
+            let data_repeats = prev_symbol == enc_output.symbol;
+            let _ = assert_eq(data_repeats && can_repeat, data_repeats);
```
proppy, line 295: would that be triggered if we were using the symbol here? or do we need a separate test?

xls/modules/rle/rle.x
proppy: Thanks, I find this a lot easier to follow! wdyt?

xls/modules/rle/rle.x
mtdudek: Sounds good to me, I'll change it

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,465 @@
...
+            ((tok, (enc_output.count == max_count)|| enc_output.last, enc_output.symbol))
+        }((tok, true, TestSymbol:0));
+        let _ = send(tok, terminator, true);
+        ()
+    }
+}
+
+
+// Data-driven testing approach
```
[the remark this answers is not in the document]
mtdudek: I'm using parts of this test in other RLE parts, so I'd rather keep it.
Also it's a showcase for more complex tests.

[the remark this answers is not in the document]
mtdudek: Removed, I'll update other PRs once this one is merged

mtdudek:
@proppy Can you check if PR is ready to be merged?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,272 @@
...
+
+//const MAX_COUNT = {std::unsigned_max_value<TEST_COUNT_WIDTH>()};
+const MAX_COUNT = u2:3;
+
+type TestSymbol = bits[TEST_SYMBOL_WIDTH];
+type TestCount = bits[TEST_COUNT_WIDTH];
+type TestEncInData = EncInData<TEST_SYMBOL_WIDTH>;
+type TestEncOutData = EncOutData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+
+// Tightly coupled testing approach
```
proppy, line 170: nit: I don't think we need this comment anymore, since there is only one approach in the file now.

xls/modules/rle/rle.x
proppy: @mtdudek can you add one last test for the overflow?

mtdudek: in line 212 there is a test that checks counter overflow

xls/modules/rle/BUILD
```diff
@@ -0,0 +1,119 @@
...
+    },
+)
+
+#xls_dslx_verilog (
```
proppy, line 90: should we remove this as well?

xls/modules/rle/BUILD
```diff
@@ -0,0 +1,119 @@
...
+xls_benchmark_ir(
+    name = "rle_enc_ir_benchmark",
+    src = ":rle_enc_opt_ir.opt.ir",
+    benchmark_ir_args = {
+        "pipeline_stages": "2",
+        "delay_model": "unit",
+    }
+)
+
+#xls_benchmark_verilog(
```
proppy, line 116: should we remove this?

mtdudek: Removed

xls/modules/rle/BUILD
mtdudek: Removed

mtdudek:
Should I squash all review commits before we merge it?

proppy:
> Should I squash all review commits before we merge it?

Yes, please!

Did you want to file an issue for adding a test for the overflow too?

mtdudek:
What exactly do you have in mind for overflow test?
One where we sand single symbol more times, than counter can hold?
Second one has such sequence, but I can add 4th test that does only that.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,267 @@
...
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH = u32:2;
+
+type TestSymbol = bits[TEST_SYMBOL_WIDTH];
+type TestCount = bits[TEST_COUNT_WIDTH];
+type TestEncInData = EncInData<TEST_SYMBOL_WIDTH>;
+type TestEncOutData = EncOutData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+
+#[test_proc]
+proc RLEEncTightTester {
```
proppy, line 168: nit: I think we might want to rename this to capture what the test is actually testing.

proppy:
> @proppy What exactly do you have in mind for overflow test?
> One where we send single symbol more times, than counter can hold?
> Second testcase has such sequence, but I can add 4th test that does only that in new PR.

I think it would make sense to move it to a separare `test_proc`, so that later when it fails we can more quickly narrow down the regression: the test runner will state the number of total tests failing and the name of the failing tests.

Would it make sense to do the following?
- move transactions to a separate `test_proc` when applicable, and name them after what's actually being tested
- define different input/output `type` alias that correspond to what's being tested (i.e: smaller size of overflow test)

Happy to file separate issue about this if you prefer to address this as a follow up PR.

mtdudek:
@proppy I think I've addressed all your comments

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+
+    next (tok: token, state: ()) {
+        ()
+    }
+}
+
+// Tests
+
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH = u32:2;
```
proppy, line 160: what do you think of using a larger count width for this test (and test more transactions) and using `2` for the overflow one?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+      );
+      let _ = assert_eq(enc_output, expected);
+      (tok)
+    }(tok);
+    let _ = send(tok, terminator, true);
+    ()
+  }
+}
+
+// `last` after `last` check
```
proppy, line 291: can you add a bit more details about what this is testing?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH = u32:2;
+
+type TestSymbol     = bits[TEST_SYMBOL_WIDTH];
+type TestCount      = bits[TEST_COUNT_WIDTH];
+
+type TestEncInData  = EncInData<TEST_SYMBOL_WIDTH>;
+type TestEncOutData = EncOutData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+
+// Simple transaction without overflows
```
proppy, line 168: s/overflows/overflow/

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+type SimpleTransactionStimulus = TestSymbol;
+type SimpleTransactionSymbol   = TestSymbol;
+type SimpleTransactionCount    = TestCount;
+
+#[test_proc]
+proc RLEEncSimpleTransactionTest {
```
proppy, line 174: nit: I think we can remove "Transaction" from the test name, and simply use something that describe what this is testing: `RLECountSymbolTest`?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+  }
+}
+
+// Transaction with counter overflow
+type OverflowTransactionStimulus = TestSymbol;
+type OverflowTransactionSymbol  = TestSymbol;
+type OverflowTransactionCount   = TestCount;
+
+#[test_proc]
+proc RLEEncOverflowTransactionTest {
```
proppy, line 231: `RLEOverflowTest`?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,338 @@
...
+
+// structure containing output data from an RLE encoder
+pub struct EncOutData<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
+    symbol: bits[SYMBOL_WIDTH], // compressed symbol
+    count: bits[COUNT_WIDTH],   // symbol count
+    last: bool,                 // if this is the last symbol
+}
+
+// structure to preserve the state of an RLE encoder
+pub struct RLEEncState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
```
proppy, line 63: nit: I think we can simply name this RLEncoderState (since the `E` of RLE is for encoding).

proppy:
I think this is ready to import, most of the changes I requested are purely cosmetic.

xls/modules/rle/rle.x
mtdudek: I would rather keep it, but I also see no issues with removing it.
I'll change it in this PR and in RLE decoder PR from `RLEDecState` to `RLDecoderState`.

proppy: Maybe we could spell it completely? `RunLengthEncoder` and and `RunLengthDecoder`.

I'm also not sure if we need to prefix the type of the state by anything, since those are not public and only used in this file.

mtdudek:
@proppy can you take a look? Unfortunately I can't mark comments as `Resolved`.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+    // zero means that the previous evaluation sent all the data and
+    // we start counting from the beginning
+    prev_count: bits[COUNT_WIDTH],
+    // flag indicating that the previous symbol was the last one
+    // in the transmission
+    prev_last: bool,
+}
+
+// RLE encoder implementation
+pub proc RLEEnc<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
```
proppy, line 77: should we also rename this to be coherent with the name of the type for the state?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+const CountSymbolTestCountWidth  = TEST_COMMON_COUNT_WIDTH;
+
+type CountSymbolTestStimulus   = TestCommonSymbol;
+type CountSymbolTestSymbol     = TestCommonSymbol;
+type CountSymbolTestCount      = TestCommonCount;
+type CountSymbolTestEncInData  = TestCommonEncInData;
+type CountSymbolTestEncOutData = TestCommonEncOutData;
+
+#[test_proc]
+proc RLEEncCountSymbolTest {
```
proppy, line 186: same comment here

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+
+type OverflowStimulus   = TestCommonSymbol;
+type OverflowSymbol     = TestCommonSymbol;
+type OverflowCount      = bits[OverflowCountWidth];
+type OverflowEncInData  = TestCommonEncInData;
+type OverflowEncOutData =
+    EncOutData<OverflowSymbolWidth, OverflowCountWidth>;
+
+#[test_proc]
+proc RLEEncOverflowTest {
```
proppy, line 251: same comment here

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+
+type LastAfterLastStimulus   = TestCommonEncInData;
+type LastAfterLastSymbol     = TestCommonSymbol;
+type LastAfterLastCount      = TestCommonCount;
+type LastAfterLastEncInData  = TestCommonEncInData;
+type LastAfterLastEncOutData = TestCommonEncOutData;
+type LastAfterLastOutput     = TestCommonEncOutData;
+
+#[test_proc]
+proc RLEEncLastAfterLastTest {
```
proppy, line 327: same comment here.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+type OverflowWithLastSymbol     = TestCommonSymbol;
+type OverflowWithLastCount      =
+    bits[OverflowWithLastCountWidth];
+type OverflowWithLastEncInData  = TestCommonEncInData;
+type OverflowWithLastEncOutData =
+    EncOutData<OverflowWithLastSymbolWidth,
+               OverflowWithLastCountWidth>;
+
+#[test_proc]
+proc RLEEncOverflowWithLastTest {
```
proppy, line 395: same comment here.

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+type TestCommonCount      = bits[TEST_COMMON_COUNT_WIDTH];
+type TestCommonEncInData  = EncInData<TEST_COMMON_SYMBOL_WIDTH>;
+type TestCommonEncOutData =
+    EncOutData<TEST_COMMON_SYMBOL_WIDTH, TEST_COMMON_COUNT_WIDTH>;
+
+// Simple transaction without overflow
+const CountSymbolTestSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const CountSymbolTestCountWidth  = TEST_COMMON_COUNT_WIDTH;
+
+type CountSymbolTestStimulus   = TestCommonSymbol;
```
proppy, line 179: curious why we need to prefix the type with the test name since they are basically shared across all tests?

xls/modules/rle/rle.x
```diff
@@ -0,0 +1,449 @@
...
+  }
+}
+
+// Transaction with counter overflow
+const OverflowSymbolWidth = TEST_COMMON_SYMBOL_WIDTH;
+const OverflowCountWidth  = u32:2;
+
+type OverflowStimulus   = TestCommonSymbol;
+type OverflowSymbol     = TestCommonSymbol;
+type OverflowCount      = bits[OverflowCountWidth];
```
proppy, line 245: it would be nice to have support for local `type` in `proc`, should we file a feature request, wdyt?

proppy:
This is ready to copybara import (remaining comments are mostly cosmetic).

xls/modules/rle/rle.x
mtdudek: I think it would be great to have `type`s local to `proc`s.
I think it may add new issue with accessing them, maybe `::` syntax, but I don't know.

mtdudek:
I'll change RLEEnc to RunLengthEncoder and squash all commits into single in preparations for merging

xls/modules/rle/rle.x
mtdudek: it's not necessary for test to work.
For me it was more descriptive to have testcase specific types.

xls/modules/rle/rle.x
mtdudek: Opened it as #1050

xls/modules/rle/rle.x
mtdudek: Fixed

xls/modules/rle/rle.x
mtdudek: Fixed

xls/modules/rle/rle.x
mtdudek: Fixed

xls/modules/rle/rle.x
mtdudek: Fixed

xls/modules/rle/rle.x
mtdudek: @proppy wdyt?

proppy: That makes sense, thanks for providing the rational.

xls/modules/rle/BUILD
```diff
@@ -0,0 +1,96 @@
...
+
+load(
+    "//xls/build_rules:xls_build_defs.bzl",
+    "xls_dslx_library",
+    "xls_dslx_test",
+    "xls_dslx_ir",
+    "xls_ir_opt_ir",
+    "xls_ir_verilog",
+    "xls_benchmark_ir",
+    "xls_benchmark_verilog",
```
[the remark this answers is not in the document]
mtdudek: `"xls_benchmark_verilog"` is unused.
I've removed it and updated PR

proppy:
@mtdudek actually, I just realized that if we want to land this and #1006 concurrently we need to decouple the `BUILD` file too. Do you mind if we split this in separate directory and then unify them in a follow up PR (sorry for all the back and forth on this).

Issue #976, opened 2023-05-23
[DSLX/IR/Codegen] Inlining IR with inter proc channels fails, fifo_depth must be < 1 but was  (none)


mtdudek:
I've encountered an issue with a design that uses 2 procs.
The first proc receives data from external channel, does some processing and
it sends data to the second proc. Second proc also does some processing and
it sends its results to an external channel.

During code generation I get error:
```
Only streaming channels of FIFO depth one or less are supported in proc inlining. Channel `proc1_proc2` has depth: (none)
```
 I've modified DLSX to IR converter, so that every intra DLSX channel has depth of 1.
It was enough to get codegen working but as i can see in #950 multiproc codegen is being developed.
I think this error is similar to #708.

I'll create PR with my current fix.
/cc @proppy

PR: https://github.com/google/xls/pull/977

grebe:
I don't think we want fifo_depth to be hardcoded to any value, and none is the correct default value.

Instead, I think it would be better for internal channels to optionally set the fifo depth. For example:

```
let (chan0_p, chan0_c) = chan<F32>();  // fifo depth 'none' by default
let (chan1_p, chan1_c) = chan<F32>(fifo_depth=3);
```

mtdudek:
I don't know, if adding fifo_depth to DSLX is a great idea.
From DSLX point of view channels are non-blocking(#965),
so they can be seen as infinite. 
How about adding flag to ir_converter that can set inter proc fifo_depth?
Also if `none` is a valid value for fifo_depth, how it should be handled by proc inline?

grebe:
I think `none` should be interpreted as 'unspecified' (essentially infinite) and the compiler should be free to insert FIFOs of whatever size it likes as long as it doesn't cause deadlock. In practice, we don't currently insert those FIFOs. Proc inlining won't inline channels w/ fifo depth of `none`.

I think a flag that sets all internal channels' fifo depths is not ideal- internal channel names are very brittle, you probably don't want them to all have the same depth, and I think in general we have too many flags and not enough source-level configuration.

It seems reasonable to me to set fifo depth in the source (potentially via parameter), and if you don't want to have to specify it everywhere have a pass that does analysis (or maybe just blanket applies a configurable default) to any unspecified depths.

[...]

PR #977, opened 2023-05-23
Force fifo_depth 1 for inter proc channels

mtdudek:
DSLX to IR conversion creates channels with
undefined fifo_depth. This causes proc inlining
to fail as it expects fifo_depth of 0 or 1.

I've modified DSLX to IR conversion so that
fifo_depth is always set to 1.


proppy:
@mtdudek let's punt on this until we converge to a solution in #976

PR #982, opened 2023-05-24
Improve zero! builtin

mtdudek:
zero! builtin did not work with parametric structs and with arrays of structs.
This commit changes ParseParametrics to allow for
parametric structs and struct arrays.
This commit also adds tests to improve
`zero!<>()` coverage.


proppy:
@mtdudek can you open a separate PR for the tests? and open a separate issue to discuss the problem and possible implementation with the rest of the team?

Thanks iin advance.

Issue #984, opened 2023-05-26
`zero!` should support parametric structs and arrays of structs


proppy:
https://google.github.io/xls/dslx_reference/#zerot doesn't seems to work with parametric strucst and arrays of structs.

Is that a known limitation, are there possible workaround?

cdleary:
Re: #982 -- we generally don't want to introduce more backtracking into the parser as an implementation strategy, since it makes for bad error messages and slow parsing. I was hoping to restructure the grammar around types a bit, this is one of the first instances where we have parametric types so the grammar was a bit overfit for parametric-values-only.

The test cases looked good though, maybe those can be used with a DISABLED_ and a TODO referencing this issue?

See also related #904

cdleary:
Oh, and the workaround is to define a type alias and then use that as the parametric argument to `zero!`

mtdudek:
I've created PR #985 with tests and disabled ones that don't work right now.
I think #976  can benefit from adding type parsing to parametric pass.

I've also created draft PR #986 with a possible solution to parse parametric types.
It doesn't parse `::` after first `<` or `[`, it also doesn't allow for using
fields from const arrays, like:
```
const a = u32[2] : {1,2}

fn b<A:u32>(x: u32) -> u32 {
  x + A
}

b<a[0]>(a[1])
```
As far as I can tell, using const array's fields was never allowed, so proposed solution doesn't break it.

proppy:
@mtdudek is that blocking #994 and #995?

mtdudek:
It's not blocking. There is workaround to pass any type to zero!<> using type definitions.

mikex-oss:
+1. Just ran into this myself.

PR #986, opened 2023-05-26
[DSLX] Improve type handling in parametrics

mtdudek:
Some buildins like `zero!` did not work with
parametric structs and with arrays of structs.

ParseParametrics returns TypeAnnotation only if
token is a buildin type. In all other cases
identifiers are expected to resolve to TypeRef or
constant values.

This commit changes ParseParametrics to allow for
parametric structs and struct arrays types.
This change only extends Parser capabiliteis.

Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/xls/dslx/frontend/parser.cc
+++ b/xls/dslx/frontend/parser.cc
@@ -2779,9 +2779,80 @@ absl::StatusOr<std::vector<ExprOrType>> Parser::ParseParametrics(
     }
 
     if (peek->kind() == TokenKind::kIdentifier) {
+      // We have to distinguish between:
+      // type identifiers and
+      // value identifiers
       XLS_ASSIGN_OR_RETURN(auto nocr, ParseNameOrColonRef(bindings));
-      return ToExprNode(nocr);
+      XLS_ASSIGN_OR_RETURN(bool complex_,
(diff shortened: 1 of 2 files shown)


proppy:
is that ready to review? is that blocking #994?

proppy:
@tmichalak can you confirm if this is still needed?

tmichalak:
> @tmichalak can you confirm if this is still needed?

So, the zero! macro is still limited and you cannot use parametric structs with it. Currently, as a workaround you need to define a type alias that renames a parametric type to a new one without explicit parameters, therefore having a more advanced zero! macro would be useful.

proppy:
> the zero! macro is still limited and you cannot use parametric structs with it.

is that already captured as an issue?

tmichalak:
> > the zero! macro is still limited and you cannot use parametric structs with it.
> 
> is that already captured as an issue?

Yes, there is [this one](https://github.com/google/xls/issues/984), but I guess that looking at the [comment](https://github.com/google/xls/issues/984#issuecomment-1563748117) and [workaround](https://github.com/google/xls/issues/984#issuecomment-1563748516) we might need to stick to the workaround for the time being.

dplassgit:
Closing due to lack of activity and obsolescence.

PR #987, opened 2023-05-26
[DSLX] Allow for fifo_depth to be specified as second param

mtdudek:
DSLX to IR conversion creates channels with
undefined fifo_depth. This causes proc inlining
to fail as it expects fifo_depth of 0 or 1.

I've modified DSLX, DSLX-to-IR so that fifo depth
can be passed as optional paramiter to channel.

It depends on :
#986 

Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/xls/dslx/frontend/ast.cc
+++ b/xls/dslx/frontend/ast.cc
@@ -740,7 +740,13 @@ std::string ChannelDecl::ToString() const {
     }
   }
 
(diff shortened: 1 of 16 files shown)


proppy:
closing for the time being in favor of https://github.com/google/xls/commit/9939db01b6a59ac8c226e25a313b9212b859ddd8, let's continue the discussion about depth in type definition in #976.

PR #990, opened 2023-05-26
[DSLX] Propagating type_info through proc frames

### LINKED ISSUE #981: DSLX: Misleading `assert_eq` failure message
I have noticed that when `assert_eq()` fails on a check that includes a state variable, the final error message is misleading.
Here is a small example showing the problem:
```
#[test_proc]
proc Test {
    terminator: chan<bool> out;
    
    init {(u32:0)}

    config(terminator: chan<bool> out) {
        (terminator,)
    }

    next(tok: token, counter: u32) {
        trace_fmt!("counter: {}", counter);
        let _ = assert_eq(counter < u32:10, true);
        counter + u32:1
    }
}
```

And the output:
```
 [ RUN UNITTEST  ] Test
I0524 16:47:25.402100 1264846 bytecode_interpreter.h:85] counter: 0
I0524 16:47:25.402162 1264846 bytecode_interpreter.h:85] counter: 1
I0524 16:47:25.402187 1264846 bytecode_interpreter.h:85] counter: 2
I0524 16:47:25.402191 1264846 bytecode_interpreter.h:85] counter: 3
I0524 16:47:25.402212 1264846 bytecode_interpreter.h:85] counter: 4
I0524 16:47:25.402216 1264846 bytecode_interpreter.h:85] counter: 5
I0524 16:47:25.402220 1264846 bytecode_interpreter.h:85] counter: 6
I0524 16:47:25.402225 1264846 bytecode_interpreter.h:85] counter: 7
I0524 16:47:25.402228 1264846 bytecode_interpreter.h:85] counter: 8
I0524 16:47:25.402233 1264846 bytecode_interpreter.h:85] counter: 9
I0524 16:47:25.402236 1264846 bytecode_interpreter.h:85] counter: 10
E0524 16:47:25.402281 1264846 run_routines.cc:274] Internal error: NOT_FOUND: Could not find concrete type for node: (counter) < (u32:10)
[        FAILED ] Test: internal error: NOT_FOUND: Could not find concrete type for node: (counter) < (u32:10)
[===============] 1 test(s) ran; 1 failed; 0 skipped.
```

[Example on Google Colab](https://colab.research.google.com/gist/rw1nkler/e61b38db7c9c152491e80a362bdfcb86/dslx-proc-playground.ipynb)
CC @proppy

mtdudek:
This commit fixes internal error when testing DSLX procs using `#test_proc()`.
Issue was that only first evaluated frame had `type_info` defined, and all following frames had it undefined. This made ConcreteType lookup fail.

Fixes #981

### DIFF
--- a/xls/dslx/bytecode/bytecode_interpreter.cc
+++ b/xls/dslx/bytecode/bytecode_interpreter.cc
@@ -1503,7 +1503,8 @@ absl::Status ProcConfigBytecodeInterpreter::EvalSpawn(
       std::unique_ptr<BytecodeInterpreter> next_interpreter,
       CreateUnique(import_data, next_bf.get(), full_next_args, options));
   proc_instances->push_back(ProcInstance{proc, std::move(next_interpreter),
-                                         std::move(next_bf), full_next_args});
+                                         std::move(next_bf), full_next_args,
+                                         type_info});
   return absl::OkStatus();
 }
 
@@ -1526,7 +1527,7 @@ absl::StatusOr<ProcRunResult> ProcInstance::Run() {
     }
 
     XLS_RETURN_IF_ERROR(interpreter_->InitFrame(next_fn_.get(), next_args_,
-                                                /*type_info=*/nullptr));
+                                                type_info_));
     return ProcRunResult{.execution_state = ProcExecutionState::kCompleted,
                          .blocked_channel_name = std::nullopt,
                          .progress_made = progress_made};
--- a/xls/dslx/bytecode/bytecode_interpreter_test.cc
+++ b/xls/dslx/bytecode/bytecode_interpreter_test.cc
@@ -1519,6 +1519,39 @@ fn main() {
   EXPECT_TRUE(value.IsUnit());
 }
 
+// https://github.com/google/xls/issues/981
+TEST(BytecodeInterpreterTest, AssertEqFailProcIterations) {
+  constexpr std::string_view kProgram = R"(
+#[test_proc]
+proc BTester {
+    terminator: chan<bool> out;
+
+    init { (u32:0) }
+
+    config(terminator: chan<bool> out) {
+        (terminator,)
+    }
+
+    next(tok: token, state: u32) {
+        assert_eq(state, u32:0);
+        // ensure at least 2 `next()` iterations to create an interpreter frame.
+        let tok = send_if(tok, terminator, state > u32:1, true);
+        (state + u32:1)
+    }
+})";
+
+  XLS_ASSERT_OK_AND_ASSIGN(auto temp_file,
+                           TempFile::CreateWithContent(kProgram, "_test.x"));
+  constexpr std::string_view kModuleName = "test";
+  ParseAndTestOptions options;
+  ::testing::internal::CaptureStderr();
+  absl::StatusOr<TestResult> result = ParseAndTest(
+      kProgram, kModuleName, std::string{temp_file.path()}, options);
+  std::string stdcerr(::testing::internal::GetCapturedStderr());
+  EXPECT_THAT(result, status_testing::IsOkAndHolds(TestResult::kSomeFailed));
+  EXPECT_THAT(stdcerr, HasSubstr("were not equal"));
+}
+
 TEST(BytecodeInterpreterTest, DistinctNestedParametricProcs) {
   // Tests that B, which has one set of parameters, can instantiate A, which has
   // a different set of parameters.
--- a/xls/dslx/bytecode/bytecode_interpreter.h
+++ b/xls/dslx/bytecode/bytecode_interpreter.h
@@ -269,11 +269,13 @@ class ProcInstance {
  public:
   ProcInstance(Proc* proc, std::unique_ptr<BytecodeInterpreter> interpreter,
                std::unique_ptr<BytecodeFunction> next_fn,
-               std::vector<InterpValue> next_args)
+               std::vector<InterpValue> next_args,
+               const TypeInfo* type_info)
       : proc_(proc),
         interpreter_(std::move(interpreter)),
         next_fn_(std::move(next_fn)),
-        next_args_(std::move(next_args)) {}
+        next_args_(std::move(next_args)),
+        type_info_(type_info) {}
 
   // Executes a single "tick" of the ProcInstance.
   absl::StatusOr<ProcRunResult> Run();
@@ -283,6 +285,7 @@ class ProcInstance {
   std::unique_ptr<BytecodeInterpreter> interpreter_;
   std::unique_ptr<BytecodeFunction> next_fn_;
   std::vector<InterpValue> next_args_;
+  const TypeInfo* type_info_;
 };
 
 }  // namespace xls::dslx


mtdudek:
I'll add some tests for asserts in test_procs

mtdudek:
I've added tests to check `assert_eq` and `assert_lt` in `#[test_proc]`.
It required adding new flag to ParseAndTestOptions

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1253,6 +1253,68 @@ fn main() -> MyU9 {
...
+
+    init { (u32:0) }
+
+    config(terminator: chan<bool> out) {
+        (terminator,)
+    }
+
+    next(tok: token, state: u32) {
+        let count = state;
+        let _ = assert_eq(count, u32:0);
```
proppy, line 1270: curious if this is also failing when comparing directly against `state`? (I think it would)

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1253,6 +1253,68 @@ fn main() -> MyU9 {
...
+        let tok = send_if(tok, terminator, count > u32:1, true);
+        (count + u32:1)
+    }
+})";
+
+  XLS_ASSERT_OK_AND_ASSIGN(auto temp_file,
+                           TempFile::CreateWithContent(kProgram, "_test.x"));
+  constexpr std::string_view kModuleName = "test";
+  ParseAndTestOptions options;
+  options.handle_status = false;
```
proppy, line 1280: couldn't find this field in the corresponding struct definition (am I missing something?)

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1253,6 +1253,68 @@ fn main() -> MyU9 {
...
+    init { (u32:0) }
+
+    config(terminator: chan<bool> out) {
+        (terminator,)
+    }
+
+    next(tok: token, state: u32) {
+        let count = state;
+        let _ = assert_eq(count, u32:0);
+        let tok = send_if(tok, terminator, count > u32:1, true);
```
proppy, line 1271: why do we need the `send_if` here?

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1253,6 +1253,68 @@ fn main() -> MyU9 {
...
+})";
+
+  XLS_ASSERT_OK_AND_ASSIGN(auto temp_file,
+                           TempFile::CreateWithContent(kProgram, "_test.x"));
+  constexpr std::string_view kModuleName = "test";
+  ParseAndTestOptions options;
+  options.handle_status = false;
+  absl::StatusOr<TestResult> result = ParseAndTest(
+      kProgram, kModuleName, std::string{temp_file.path()}, options);
+  EXPECT_THAT(result.status(), StatusIs(absl::StatusCode::kInternal,
```
proppy, line 1283: shouldn't this succeed? (i.e: if we do fix the bug we shouldn't get an internal error anymore)

xls/dslx/bytecode/bytecode_interpreter_test.cc
proppy: ah I see that's being added below. Can you document why the new field is necessary (either in the PR description or in the struct itself).

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: It's necessary, as I found that first evaluation will yield correct error message, that is why I have to delay send, so that in first iteration send will not happen, and in second iteration assert will fail

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: I think it does.
Should I change this to reduce number of variables?

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: When assert expression is faulse, interpreter returns ststus that is `kInternal`.
I've taken this check from assert_eq test for functions(https://github.com/google/xls/blob/d86ff6c19790ddb8f390898c7814e77e4c0c05eb/xls/dslx/bytecode/bytecode_interpreter_test.cc#L212)

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: I'll add comment to struct definition to explain existence of this field.

xls/dslx/bytecode/bytecode_interpreter_test.cc
proppy: as you want, let me know what you find more readable.

xls/dslx/bytecode/bytecode_interpreter_test.cc
proppy: ah, I see maybe you can add a comment to make that clearer? also since we are only testing dslx interpreter here (and not ir conversion) maybe concurrently send would work?

xls/dslx/run_routines.cc
```diff
@@ -258,6 +258,8 @@ absl::StatusOr<TestResult> ParseAndTest(std::string_view program,
 
   auto handle_error = [&](const absl::Status& status,
                           std::string_view test_name, bool is_quickcheck) {
+    if (!options.handle_status)
```
proppy, line 261: can you add more context in the PR description about why this change is necessary?

just looking at the change, it's not clear to me why you wouldn't want to always return the status to the caller.

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: I tried it. Error only triggers after new interpreter frame is created, so test has to reach end of next function at least once and start once again.

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: I think I'll remove count and use state directly

xls/dslx/run_routines.cc
mtdudek: With this option I was able to keep default behavior of ParseAndTest.

ParseAndTest would return on first failed test, if it would return test status. 
Right now, it logs cause of the failure and continues to execute remaining tests.

proppy: ah, so this flag is mostly there to verify that the test case that reproduce the bug fails early?

if yes, maybe we should modify the the assertion so that they ignore other test results?

mtdudek: I don't think it's a good idea to modify assertion behavior.
What I see right now, is that, all DSLX tests use assertions to check functions/proc outputs.
Early escape on assertion failure would bring us back to skipping tests after first fail.

proppy: > Early escape on assertion failure would bring us back to skipping tests after first fail.

But making this behavior configurable (being able to basically `fail_fast`) seems orthogonal to what this PR is fixing (the type info propagation), do you think it's worth splitting in a separate issue/pr?

proppy: > I can't create tests for assert_* without fast_fail

couldn't you extract the number of failure and the condition without `fast_fail`?

mtdudek: I'm unable to do so from ParseAndTest output in its unmodified version. 
https://github.com/google/xls/blob/d28e813f1d56c8ec1c3502219e14b48b8a82be4f/xls/dslx/run_routines.cc#L371-L375
https://github.com/google/xls/blob/d28e813f1d56c8ec1c3502219e14b48b8a82be4f/xls/dslx/run_routines.cc#L259-L282
Captures all interpreter errors.

proppy: Yes, but since you're the one creating the `test_proc` you know they are only supposed to yield one error and could potentially assert this in the output?

proppy: oh, so you mean that you're unable to get the test to fails with the previous version of the code?

Since your change does fix the type inference, I think what you have with checking that the error message includes "not less than" is enough.

In the passing case (with this patch applied) the error message with be surfaced, without the error would be lost and the test will fail.

Am I missing something?

mtdudek: No, that's exactly what was and is happening there.
I hope now it’s clearer why I had to add this new flag.

proppy: > error message is lost in ParseAndTest error handler

My understanding was that the error message was lost in ParseAndTest error handler because or the concrete type missing error this PR is fixing, or is that another issue?

mtdudek: By saying "error message is lost" I meant that it isn't propagated to ParseAndTest caller.

proppy: Yep, but it sounds that now that you're fixing the type deduction error, the correct message should be surfaced.

Are you worried that you wouldn't be able to identify a regression? (and that's why you're adding the flag?), would there be a downside to always returning the error from `ParseAndTest`?

mtdudek: Before changes ParseAndTest would only return status that either indicated all tests passed or at least one failed, without additional error message. I can't distinguish between test failing on type deduction from test failing on assertion. I'm using status error message to distinguish between these 2 options and without it I can't detect regression.

I think we can always return `status` , I'll change it to do so.

mtdudek:
@proppy wdyt about this PR right now?

xls/dslx/run_routines.cc
```diff
@@ -372,7 +373,9 @@ absl::StatusOr<TestResult> ParseAndTest(std::string_view program,
     if (status.ok()) {
       std::cerr << "[            OK ]" << std::endl;
     } else {
-      handle_error(status, test_name, /*is_quickcheck=*/false);
+      auto handled = handle_error(status, test_name, /*is_quickcheck=*/false);
+      if (!options.handle_status)
```
proppy, line 377: curious why we don't always return `status` here too?

mtdudek: Always returning `status` here would change behavior of `ParseAndTest` to always fail fast. @proppy

xls/dslx/run_routines.cc
```diff
@@ -279,6 +279,7 @@ absl::StatusOr<TestResult> ParseAndTest(std::string_view program,
                                     suffix)
               << "\n";
     failed += 1;
+    return status;
```
proppy, line 282: I don't think we actually need to return anything here, this it's a `const` reference and we already take it as an argument (meaning that the caller is already able to check for its value)

xls/dslx/run_routines.cc
proppy: I see, so we want to:
a/ keep the current behavior (run all tests, count the number of failure and report individual test failure)
b/ make sure that the right error is populated (actual proc test failure, rather than type lookup error)
?

mtdudek: Yes, I want to keep `a` behavior and add `b` through option.

proppy: it seems to me that we can achieve (and test) `b/` without introducing a way to configure the failfast behavior: we just have to check for the presence of the right error message in the output (and the absence of type inference error: if we want to check for regression)

xls/dslx/run_routines.cc
mtdudek: Yes, you're right, I'll remove it in next commit

xls/dslx/run_routines.cc
mtdudek: So you would like to capture std::cerr in testing infrastructure and check it's contents?

mtdudek: In example above test calls `Interpret` method https://github.com/google/xls/blob/main/xls/dslx/bytecode/bytecode_interpreter_test.cc#L400, while `procs` call `TestAndProc`.
`Interpret` directly calls Bytecode interpreter, so status returned represents exactly what has happened inside the function.
Procs due to their stateful nature require evaluation environment with eval loop. This is done by calling ParseAndTest.

proppy: It seems that testing for:
```
status_testing::IsOkAndHolds(TestResult::kSomeFailed)
```
could allow us to test for failure without fail fast.

wdyt?

mtdudek: Yes, it would give us information that tests failed, but it doesn't have reason for why each test failed.
That would lead back to ambiguity for the source of failure.

I checked and GTEST doesn't have publicly defined options to capture std:cerr or std::cout, but it has internal function to do so. I can use it to capture handlers error message and check its contents for the expected string.

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
   EXPECT_TRUE(value.IsUnit());
 }
 
+// GH 981
```
proppy, line 1522: nit: in the absence of well defined convention, let's use the fully qualified url

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
...
+  ParseAndTestOptions options;
+  ::testing::internal::CaptureStderr();
+  absl::StatusOr<TestResult> result = ParseAndTest(
+      kProgram, kModuleName, std::string{temp_file.path()}, options);
+  std::string stdcerr(::testing::internal::GetCapturedStderr());
+  EXPECT_THAT(result, status_testing::IsOkAndHolds(TestResult::kSomeFailed));
+  EXPECT_THAT(stdcerr, HasSubstr("were not equal"));
+}
+
+// GH 981
```
proppy, line 1558: nit: in the absence of well defined convention, let's use the fully qualified url

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
...
+// Assert_lt type deduction didn't fail in
+// the first iteration of the next function.
+// This test requires 2 iterations of next function,
+// which triggered assert_lt bug.
+TEST(BytecodeInterpreterTest, AssertLtFailProcs) {
```
proppy, line 1563: I'm not sure if we need the extra test (as a regression in the codepath would likely trigger the other one as well)

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
   EXPECT_TRUE(value.IsUnit());
 }
 
+// GH 981
+// Assert_eq type deduction didn't fail in
+// the first iteration of the next function.
+// This test requires 2 iterations of next function,
+// which triggered assert_eq bug.
+TEST(BytecodeInterpreterTest, AssertEqFailProcs) {
```
proppy, line 1527: nit: what about naming `AssertEqFailProcIterations` to capture the iteration clause?

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
   EXPECT_TRUE(value.IsUnit());
 }
 
+// GH 981
+// Assert_eq type deduction didn't fail in
+// the first iteration of the next function.
+// This test requires 2 iterations of next function,
```
proppy, line 1525: let's add the command inline next to the `send_if` for better context.

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,78 @@ fn main() {
   EXPECT_TRUE(value.IsUnit());
 }
 
+// GH 981
+// Assert_eq type deduction didn't fail in
```
proppy, line 1523: not sure if we need this extract docstring if we rename the test.

xls/dslx/run_routines.cc
mtdudek: @proppy
I've removed `options.handle_status` completely in the newest commit.
I'm capturing `std::cerr` and result from ParseAndTest. Testing captured `std::cerr` catches regression.

proppy (approved):
Just some nits; also can we name the PR and commits after what the change is doing: "propagating type_info in proc frames" (rather than what it is fixing)

mtdudek:
I'll fix commit message in the next squash/rebase cycle

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: Changed to full URL

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: Is it better now?

xls/dslx/bytecode/bytecode_interpreter_test.cc
```diff
@@ -1519,6 +1519,42 @@ fn main() {
...
+
+    init { (u32:0) }
+
+    config(terminator: chan<bool> out) {
+        (terminator,)
+    }
+
+    next(tok: token, state: u32) {
+        assert_eq(state, u32:0);
+        // Assert_eq type deduction didn't fail in
```
proppy, line 1537: would suggest a slight rephrasing to be shorter:
```
ensure at least 2 `next()` iterations to create an interpreter frame.
```

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: Doc string moved closer to `send_if`.

xls/dslx/bytecode/bytecode_interpreter_test.cc
mtdudek: Fixed.

proppy (approved):
LGTM: feel free to rebase, before I trigger the internal import!

proppy:
Thanks for the hard work on this, and sorry for the long/delayed review.

mtdudek:
Squashed and edited commit message.

xls/dslx/bytecode/bytecode_interpreter.cc
```diff
@@ -1503,7 +1503,7 @@ absl::Status ProcConfigBytecodeInterpreter::EvalSpawn(
       std::unique_ptr<BytecodeInterpreter> next_interpreter,
       CreateUnique(import_data, next_bf.get(), full_next_args, options));
   proc_instances->push_back(ProcInstance{proc, std::move(next_interpreter),
-                                         std::move(next_bf), full_next_args});
+                                         std::move(next_bf), full_next_args, type_info});
```
proppy, line 1506: nit: can you fix this to be `<= 80 columns wide`?

mtdudek: Fixed @proppy

PR #1003, opened 2023-06-01
Add new bazel rule `xls_dslx_multi_test`

mtdudek:
This PR adds new "compare_list" flag to xls_dslx_test rule.
It allows for multiple compare modes in single test target.

This removes necessity to create new test targets for each 
compare tested mode (see test rules in #974).

CC @proppy


proppy:
Is there already an issue filed for the bug/feature that this PR is fulfilling?

PR #1004, opened 2023-06-01
Add simple RLE stream muxer

mtdudek:
This PR:
* adds implementation for a simple stream muxer,
* test and gen verilog rules for this muxer,
* adds intra RLE interface specification,
* moves rle.x to rle_enc.x

Stream muxer is responsible for converting RLE encoder 
(symbol, count) pair stream into stream of words.

This PR depends on:
- [ ] #974 

Related: #994

CC @proppy

### DIFF
--- a/xls/modules/AXI4DMA/address_generator.x
+++ b/xls/modules/AXI4DMA/address_generator.x
@@ -1,0 +1,33 @@
+// Copyright 2023 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


proppy:

tmichalak:

mtdudek:
I've removed dependency on #1012, as it's tangential to this PR.
Now this PR only requires #974

proppy:
can you clarify why the muxer is a necessary part of the implementation, either in the PR description and/or in the source file?

mtdudek:
Technically speaking muxing (symbol, counter) pair stream into stream of words isn't strictly necessary to implement RLE. I'll close this PR

PR #1005, opened 2023-06-01
Add simple RLE encoding pipeline

mtdudek:
This PR adds example of simple RLE encoding pipeline.
It uses RLE encoder from #974 and stream muxer form #1004.

Depends on:
- [ ] #1004 

CC @proppy


proppy:

tmichalak:
It seems we are close to merging #974, so that we can start reviewing this soon.

proppy:
can we decouple this from #1012?

mtdudek:
I'm closing this PR, as it's not necessary to have basic RLE implementation.

PR #1006, opened 2023-06-01
Add RLE decoder

mtdudek:
This PR adds a simple implementation of RLE decoder. 
It depends on:
- [ ] #1004 for intra RLE interfaces definitions.

CC @proppy


proppy:

tmichalak:

mtdudek:
I'm removing _draft_ label from this PR

proppy:
can you rebase on top of #974?

xls/modules/rle/rle_interface.x
```diff
@@ -0,0 +1,34 @@
+// Copyright 2023 The XLS Authors
```
proppy, line 1: maybe rename this to `rle_common.x` to match the convention in https://github.com/google/xls/tree/main/xls/modules/aes?

xls/modules/rle/rle_interface.x
```diff
@@ -0,0 +1,34 @@
...
+
+// This file implements a parametric RLE encoder
+//
+// The encoder uses Run Length Encoding (RLE) to compress the input stream
+// with symbols (data to compress). The output contains packets of data with
+// the symbols and the number of their occurrences in the input stream.
+
+// structure containing output data a Data preprocessor
+// and input to an RLE encoder
+pub struct PreData<SYMB_WIDTH: u32> {
```
proppy, line 23: is there a better name for this? i.e: `DecData`?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH>> in,
+        output_s: chan<DecOutData<SYMBOL_WIDTH>> out,
+    ) {(input_r, output_s)}
+
+    next (tok: token, state: RLEDecState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+        let zero_input = DecInData { symbol: bits[SYMBOL_WIDTH]:0, count: bits[COUNT_WIDTH]:0, last: false };
+        let empty = state.count == bits[COUNT_WIDTH]:0;
+        let (input_tok, input) = recv_if(tok, input_r, empty, zero_input);
+        let (next_symbol, next_count, next_last) = if (empty) {
+            let t_count = input.count - bits[COUNT_WIDTH]:1;
```
proppy, line 87: is `t_count` for `token_count`?

proppy, line 87: seems like this could go after the `if` block, since we substract `1` in both case?

mtdudek: it was for `temporary_count`, as it's just an intermediate value.

mtdudek: sounds good to me, I'll change that

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+    )}
+
+    config (
+        input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH>> in,
+        output_s: chan<DecOutData<SYMBOL_WIDTH>> out,
+    ) {(input_r, output_s)}
+
+    next (tok: token, state: RLEDecState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+        let zero_input = DecInData { symbol: bits[SYMBOL_WIDTH]:0, count: bits[COUNT_WIDTH]:0, last: false };
+        let empty = state.count == bits[COUNT_WIDTH]:0;
```
proppy, line 84: should we rename `empty` to `symbol_completed` to be more explicit?

mtdudek: I think empty was good, but this will be more descriptive, so I'll change it

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        let (input_tok, input) = recv_if(tok, input_r, empty, zero_input);
```
proppy, line 85: it might actually be simpler to put the recv inside the if?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        let (next_symbol, next_count, next_last) = if (empty) {
+            let t_count = input.count - bits[COUNT_WIDTH]:1;
+            (input.symbol, t_count, input.last)
+        } else {
+            let t_count = state.count - bits[COUNT_WIDTH]:1;
+            (state.symbol, t_count, state.last)
+        };
+        let send_last = next_last & (next_count == bits[COUNT_WIDTH]:0);
```
proppy, line 93: shouldn't that be `&&` here ?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        let data_tok = send(input_tok, output_s, DecOutData {symbol: next_symbol, last: send_last});
+        RLEDecState {
+            symbol: next_symbol,
+            count: next_count,
+            last: next_last,
```
proppy, line 98: should we still be setting `next_last` and `next_symbol` and `next_count` if we're done sending?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH  = u32:2;
+
+type TestSymbol = bits[TEST_SYMBOL_WIDTH];
+type TestCount  = bits[TEST_COUNT_WIDTH];
+type TestDecInData  = DecInData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+type TestDecOutData = DecOutData<TEST_SYMBOL_WIDTH>;
+
+
+//const TEST_SEND_LOOKUP = TestDecInData[10]:[
```
proppy, line 132: remove commented block?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        (terminator, dec_input_s, dec_output_r)
+    }
+
+    next(tok: token, state: ()) {
+
+        // Simple transaction without repeats
+        let trans_send: (TestSymbol, TestCount)[2] =
+            [(TestSymbol:0xA, TestCount:0x3), (TestSymbol:0xB, TestCount:0x1)];
+        let tok = for ((counter, (symbol_in, count_in)), tok): ((u32, (TestSymbol, TestCount)) , token) in enumerate(trans_send) {
+            let _last = counter == (array_size(trans_send) - u32:1);
```
proppy, line 165: curious why you have `_` here?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+        ];
+        let tok = for ((counter, symbol_out), tok): ((u32, TestSymbol) , token) in enumerate(trans_recv) {
+            let _last = counter == (array_size(trans_recv) - u32:1);
+            let data_out = TestDecOutData{symbol: symbol_out, last: _last};
+            let (tok, dec_output) = recv(tok, dec_output_r);
+            let _ = trace_fmt!(
+                "Received {} transactions, symbol: 0x{:x}, last: {}",
+                counter, dec_output.symbol, dec_output.last
+            );
+            let _ = assert_eq(dec_output, data_out);
```
proppy, line 183: should we assert `last` too?

xls/modules/rle/rle_dec.x
mtdudek: If we are done sending next_count is 0, so will stall at recv_if

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,246 @@
...
+            (tok)
+        }(tok);
+
+        // Transaction with repeating symbols
```
proppy, line 187: is the different thing we're testing here the symbol `0xC` being chunked across two packet? shouldn't that meant that we set `last` on the first one (as we don't overflow the only time it can happen is with two transactions)

xls/modules/rle/rle_dec.x
mtdudek: missed that, removed in next commit

xls/modules/rle/rle_dec.x
mtdudek: It' left over from when I had there name collision.
I can remove it, but it will lead to `last: last`

xls/modules/rle/rle_dec.x
proppy: oh I see, I think `count` is perfectly acceptable then.

xls/modules/rle/rle_dec.x
mtdudek: having `recv` outside `if` reduces returned tuple.
With `recv` in if, `tok` would have to be returned as well.

mtdudek: so moving `recv` to `if` caused `segmentation fault` during DSLX to IR conversion

proppy: did you already file an issue about this?

xls/modules/rle/rle_common.x
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file implements a parametric RLE encoder
+//
+// The encoder uses Run Length Encoding (RLE) to compress the input stream
+// with symbols (data to compress). The output contains packets of data with
+// the symbols and the number of their occurrences in the input stream.
+
+// structure containing output data a Data preprocessor
```
proppy, line 21: can you reword this comment?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,292 @@
...
+        let zero_input = DecInData { symbol: bits[SYMBOL_WIDTH]:0, count: bits[COUNT_WIDTH]:0, last: false };
+        let recv_next_symbol = (state.count == bits[COUNT_WIDTH]:0);
+        let (tok, input) = recv_if(tok, input_r, recv_next_symbol, zero_input);
+        let (next_symbol, count, next_last) = if (recv_next_symbol) {
+            (input.symbol, input.count, input.last)
+        } else {
+            (state.symbol, state.count, state.last)
+        };
+        let next_count = count - bits[COUNT_WIDTH]:1;
+        let send_last = next_last && (next_count == bits[COUNT_WIDTH]:0);
```
proppy, line 92: should we capture `(next_count == bits[COUNT_WIDTH]:0)` condition with a name? i.e: `done_sending`

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,292 @@
...
+        let data_tok = send(tok, output_s, DecOutData {symbol: next_symbol, last: send_last});
+        if (send_last) {
```
proppy, line 94: I wonder if we could be more explicit here by flattening the three possible state (with a match statement):
https://google.github.io/xls/dslx_reference/#match-expression
```
  match (recv_next_symbol, done_sending, send_last) {
    (_, true, true) => zero!<RLEDecState>(),
    (true, _, _) => (input.symbol, input.count, input.last)
    (false, _, _) => (state.symbol, state.count, state.last)
  }
```
wdy?

xls/modules/rle/rle_dec.x
proppy: same commment here, is it valid in the protocol to heave multiple consecutive `(symbol, count)` packet with the same `symbol` without having `last` being set between them?

xls/modules/rle/rle_dec.x
proppy: clarified my comment with the `match` suggestion.

xls/modules/rle/rle_dec.x
mtdudek: Issue #1035

xls/modules/rle/rle_dec.x
mtdudek: I agree it will make this code cleaner.
I'll change it in next commit

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,512 @@
...
+          count: next_count,
+          last: next_last,
+        }
+      }
+  }
+}
+
+
+// RLE decoder specialization for the codegen
+proc RLEDec32 {
```
proppy, line 108: rename to match the state naming convention?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,512 @@
...
+struct RunLengthDecoderState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
+  // symbol to be repeated on output
+  symbol: bits[SYMBOL_WIDTH],
+  // count of symbols that has to be send
+  count: bits[COUNT_WIDTH],
+  // send last when repeat ends
+  last: bool,
+}
+ // RLE decoder implementation
+pub proc RLEDec<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
```
proppy, line 65: rename to match the state naming convention?

xls/modules/rle/rle_dec.x
proppy: I think `last: last` is ok.

xls/modules/rle/rle_dec.x
mtdudek: Names updated

xls/modules/rle/rle_dec.x
mtdudek: I've changed all `_last` to `last`

xls/modules/rle/rle_dec.x
mtdudek: It's possible to receive multiple `(symbol, count)` pairs with the same `symbol` without `last`being set in the first one.
This can happen when either encoder's counter has overflown or some preceding stage has decided to send such packets. Decoder must be able to handle such transaction.

xls/modules/rle/rle_common.x
mtdudek: I missed this comment before, now it's reworded.

xls/modules/rle/rle_dec.x
mtdudek: I've changed code to capture condition as requested

mtdudek:
@proppy can you check if PR is good for merging, or some more work is required?

xls/modules/rle/rle_common.x
```diff
@@ -0,0 +1,34 @@
...
+// It is also used as an output from RLE decoder.
+pub struct PlainData<SYMB_WIDTH: u32> {
+    symbol: bits[SYMB_WIDTH], // symbol
+    last: bool,               // flush RLE
+}
+
+// Structure contains compressed (symbol, counter) pairs.
+// Structure is used as an output from RLE encoder and
+// as an input to RLE decoder.
+pub struct EncData<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
```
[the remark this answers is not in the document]
mtdudek: Sure, I'll change it, and squash all commits into single one, wdyt? @proppy

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,520 @@
...
+
+  next (tok: token, state: RunLengthDecoderState<SYMBOL_WIDTH, COUNT_WIDTH>) {
+    let zero_input = DecInData {
+      symbol: bits[SYMBOL_WIDTH]:0,
+      count: bits[COUNT_WIDTH]:0,
+      last: false
+    };
+    let recv_next_symbol = (state.count == bits[COUNT_WIDTH]:0);
+    let (tok, input) = recv_if(tok, input_r, recv_next_symbol, zero_input);
+    let (next_symbol, count, next_last) = if (recv_next_symbol) {
```
proppy, line 90: I wonder if we could simply assign the `zero_input` so the current state and get rid of this if?

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,520 @@
...
+    ()
+  }
+}
+
+// Tests
+
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH  = u32:32;
+
+type TestCommonSymbol     = bits[TEST_SYMBOL_WIDTH];
```
proppy, line 137: nit: should this be named after the test `SingleSymbol`?

proppy: or simply drop the `Command` if it's shared between multiple tests.

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,520 @@
...
+      );
+      assert_eq(dec_output, output);
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
+  }
+}
+
+// Check RLE decoder on longer transaction
```
proppy, line 451: I think we can simply keep this test + the special cases (and drop the simpler one at the beginning), what do you think?

proppy (approved):
Looks good, just some nits around naming, simplification and testing.

xls/modules/rle/rle_dec.x
mtdudek: Sure thing, I'll see that 2 tests are redundant. I'll remove them in next update

mtdudek: I hope it better now, wdyt @proppy?

xls/modules/rle/rle_dec.x
mtdudek: But it's `Common`, so maybe it can be left as it is, wdyt @proppy?

xls/modules/rle/rle_dec.x
mtdudek: Changed it

xls/modules/rle/rle_dec.x
proppy: yep, I understand it's being shared by the tests but I think the definition appearing at the beginning of the file is enough of a hint to convey that it's being shared (without having to prefix all the name by `Common`)

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,391 @@
...
+#[test_proc]
+proc RunLengthDecoderrTransactionTest {
```
proppy, line 324: s/Decoderr/Decoder/

xls/modules/rle/rle_dec.x
proppy: yep, I don't think we need `RunLengthDecoderSingleSymbolTest` and `RunLengthDecoderSinglePacketTest`, since we have the longer test.

I think test that cover corner cases (`last`, `overflow` packet patterns) would be more relevant.

xls/modules/rle/rle_dec.x
mtdudek: Fixed

xls/modules/rle/rle_dec.x
mtdudek: I've removed `Common` from naming.

proppy:
I wonder if we should decouple this from https://github.com/google/xls/pull/974, in order to import both PRs concurrently (at the cost of a bit of duplication that we could resolve in the other PRs)

xls/modules/rle/rle_dec.x
mtdudek: I've removed these tests. I agree that they were redundant.

proppy:
> mtdudek force-pushed the rle-decoder branch from d28c593 to 633e8da 

it's sometime challenging to review changes that are force-pushed, maybe we could send incremental commit and only force push at the end after rebasing once the review is complete?

mtdudek:
My bad, sorry with these force pushes.

mtdudek:
> I wonder if we should decouple this from #974, in order to import both PRs concurrently (at the cost of a bit of duplication that we could resolve in the other PRs)

I'll change first commit in the chain, so it will add interface file, but it will not change rle, and rle renaming will be removed.
This should be enough to decouple it from #974. I'll create new PR that only updates RLE and changes it's name. 
Is that ok with you  @proppy?

proppy:
> I'll change first commit in the chain, so it will add interface file, but it will not change rle, and rle renaming will be removed.

That would work too, or we could simply duplicate the types and create a new PR that add `rle_common.x` and remove the duplication.

mtdudek:
> > I'll change first commit in the chain, so it will add interface file, but it will not change rle, and rle renaming will be removed.
> 
> That would work too, or we could simply duplicate the types and create a new PR that add `rle_common.x` and remove the duplication.

I'll take this route. I'll have to force push it. @proppy please tell me when I can do it safely.

proppy:
> I'll have to force push it

I think it's always fine to force push after all comments have been resolved and the PR is approved. It's just that it's useful to be able to review incremental commits before resolving pending comments.

mtdudek:
Remove dependency on #974 @proppy

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,274 @@
...
+      );
+      assert_eq(dec_output, output);
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
+  }
+}
+
+// Check RLE decoder on longer transaction
```
proppy, line 205: can we put the general test case first?

proppy, line 205: > "longer transaction"

It's not really "longer" since there is no more simple tests :)

xls/modules/rle/rle_dec.x
```diff
@@ -0,0 +1,274 @@
...
+const TEST_SYMBOL_WIDTH = u32:32;
+const TEST_COUNT_WIDTH  = u32:32;
+
+type TestSymbol     = bits[TEST_SYMBOL_WIDTH];
+type TestCount      = bits[TEST_COUNT_WIDTH];
+type TestStimulus   = (TestSymbol, TestCount);
+type TestDecInData  = DecInData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
+type TestDecOutData = DecOutData<TEST_SYMBOL_WIDTH>;
+
+// Check that RLE decoder will create 2 `last` output packets,
```
proppy, line 145: q: are there other corner case we should test for?

mtdudek: I think there are no more corner cases to be checked.

xls/modules/rle/rle_dec.x
mtdudek: Removed "longer" from the comment.

mtdudek: Changed, so that simple tests is first

xls/modules/rle/rle_dec.x
proppy: what if send a packet with a `count` of `0`?

mtdudek: I assumed that count is never 0, as encoder never emits pairs with count = 0.

proppy: I see, that makes sense if we only care about testing valid input.

proppy:
Feel free to rebase and squash.

proppy:
@mtdudek actually, I just realized that if we want to land this and #974 concurrently we need to decouple the `BUILD` file too. Do you mind if we split this in separate directory and then unify them in a follow up PR (sorry for all the back and forth on this).

mtdudek:
I'll move it to rle_dec folder for now.

xls/modules/rle/rle_dec.x
grebe: I think it is useful to treat count as being offset by 1 so count=0 means 1 symbol, count=1 means 2 symbols, etc. Disallowing count=0 is wasteful when count is narrow, e.g. in the case of count being 1 bit it will always be longer. If you encode count with an offset, it should probably be named something like `offset_count` or `count_minus_one` to avoid confusion.

grebe: Regardless, as-is I think there should be a `fail!()` if count is 0, as otherwise it is interpreted as `2^COUNT_WIDTH` symbols.

mtdudek: If you have nothing against it, for now I will add `fail!()` if count is `0`.

proppy: @mtdudek do you think we should file an issue to capture @grebe's offset by 1 suggestion? I think it would be worth discussing it.

xls/modules/rle_dec/rle_dec.x
```diff
@@ -94,7 +94,11 @@ pub proc RunLengthDecoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
     };
     let recv_next_symbol = (state.count == bits[COUNT_WIDTH]:0);
     let (tok, input) = recv_if(tok, input_r, recv_next_symbol, state_input);
-    let next_count = input.count - bits[COUNT_WIDTH]:1;
+    let next_count = if input.count == bits[COUNT_WIDTH]:0 {
+        fail!("Count_eq_0", input.count)
```
proppy, line 98: nit: rename to `invalid_count_0`?

xls/modules/rle_dec/rle_dec.x
```diff
@@ -94,7 +94,11 @@ pub proc RunLengthDecoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
     };
     let recv_next_symbol = (state.count == bits[COUNT_WIDTH]:0);
     let (tok, input) = recv_if(tok, input_r, recv_next_symbol, state_input);
-    let next_count = input.count - bits[COUNT_WIDTH]:1;
+    let next_count = if input.count == bits[COUNT_WIDTH]:0 {
```
proppy, line 97: q: I don't think we have a way to test for `fail!`, should we file a feature request for `assert_fail!`?

proppy (approved):
just some optional nits, can you rebase and and squash?

xls/modules/rle/rle_dec.x
proppy: filed https://github.com/google/xls/issues/1070

xls/modules/rle_dec/rle_dec.x
mtdudek: Updated @proppy

mtdudek:
Squashed and rebased

PR #1007, opened 2023-06-01
Add simple RLE demuxer

mtdudek:
This PR adds implementation for simple stream demuxer.
It converts incoming stream of words, with pattern symbol, count, symbol, count
into stream of (symbol, count) pairs.
PR includes test, verilog and ir_benchamerk targets.  

It depends on:
- [ ] #1006

CC @proppy


proppy:

tmichalak:

proppy:
Can we decouple this from #1012?

proppy:
can you clarify why the demuxer is a necessary part of the implementation, either in the PR description and/or in the source file?

mtdudek:
This isn't necessary to have RLE decoder working, so I'm closing this PR

PR #1008, opened 2023-06-01
Add simple RLE decoding pipeline

mtdudek:
This PR adds implementation for simple RLE decoding pipeline.
It uses simple RLE decoder and simple stream demuxer to convert
incoming stream of words, with pattern symbol, count, symbol, count
into stream of symbols that should be equal to original stream.
PR includes test, verilog and ir_benchamerk targets.  

It depends on:
- [ ] #1007

CC @proppy


proppy:

mtdudek:
It needs #974 to be merged first. We could start review now, and I will rebase final results on then current  main.

mtdudek:
This PR isn't necessary for RLE decoder

Issue #1037, opened 2023-06-19
DSLX: feature request - support send/recv within if block & match arm


rdob-ant:
Having a possibility to use send/recv within `if` conditional expressions will greatly simplify writing certain types of code in DSLX (e.g. large state machines).

Consider the following code:
```
pub proc foo {
    chn: chan<u32> out;

    init {u32:0}

    config(chn: chan<u32> out) {
        (chn, )
    }

    next(tok: token, state: u32) {
        let a = if (state < u32:1000) {
            send(tok, chn, u32:5);
            true
        } else {
            false
        };

        state + u32:1
    }
}
```

With current XLS it produces following chunks of IR and optimized IR:
```
--IR--
top proc __test__foo_0_next(__token: token, __state: bits[32], init={0}) {
  literal.4: bits[32] = literal(value=1000, id=4, pos=[(0,80,32)])
  literal.6: bits[32] = literal(value=5, id=6, pos=[(0,81,31)])
  ult.5: bits[1] = ult(__state, literal.4, id=5, pos=[(0,80,26)])
  literal.9: bits[1] = literal(value=0, id=9, pos=[(0,84,12)])
  literal.8: bits[1] = literal(value=1, id=8, pos=[(0,82,12)])
  literal.11: bits[32] = literal(value=1, id=11, pos=[(0,87,20)])
  send.7: token = send(__token, literal.6, channel_id=0, id=7)
  literal.3: bits[1] = literal(value=1, id=3)
  a: bits[1] = sel(ult.5, cases=[literal.9, literal.8], id=10, pos=[(0,80,16)])
  add.12: bits[32] = add(__state, literal.11, id=12, pos=[(0,87,14)])
  after_all.13: token = after_all(__token, send.7, id=13)
  next (after_all.13, add.12)
}


--Opt IR--
top proc __test__foo_0_next(__token: token, init={}) {
  literal.6: bits[32] = literal(value=5, id=6, pos=[(0,81,31)])
  send.7: token = send(__token, literal.6, channel_id=0, id=7)
  next (send.7)
}
```

**Observed behavior:**
Although `send()` is called conditionally in the original DSLX code, this conditional nature is not reflected in generated IR, and `send()` is always executed.

**Expected:**
`send()` should remain conditional in the generated IR, that is, it should be converted into send-with-predicate: `send(.., predicate=...)`.

---

[...]

Issue #1035, opened 2023-06-19
DSLX: bug - SIGSEGV in IR generation when token is returned in a tuple


rdob-ant:
Consider the following code:
```
pub proc foo {
    init {()}
    config() {()}

    next(tok: token, state: ()) {
        let tok = if true {
            tok
        } else {
            tok
        };
        ()
    }
}
```

Trying to make IR out of it results in the following, which I presume is expected:
```
$ ir_converter_main ...

F0619 18:16:20.305308 1718188 ir_converter_main.cc:163] Check failed: ::absl::OkStatus() == (status) (OK vs. INVALID_ARGUMENT: Could not build IR: Expected sel.5 to not contain a token in its type; type is token
  File: 0, Line: 20, Col: 18
Stack Trace:
0x5650217e5732: xls::BuilderBase::SetError()
0x5650217e5af7: xls::BuilderBase::CreateBValue()
0x5650217e8846: xls::BuilderBase::Select()
0x5650217e8978: xls::BuilderBase::Select()
```

---

Now change the code as follows:
```
pub proc foo {
    init {()}
    config() {()}

    next(tok: token, state: ()) {
        let (a, tok) = if true {
            (false, tok)
        } else {
            (true, tok)
        };
        ()
    }
}
```

Trying to run IR converter results in a silent segmentation fault. To figure out what happens the user must enable VLOG, scroll up and take a note of that SetError:
```
I0619 18:20:12.926616 1718441 verifier.cc:2014] Verifying node: sel.9: (bits[1], token) = sel(literal.4, cases=[tuple.8, tuple.6], id=9, pos=[(0,37,23)])
I0619 18:20:12.926625 1718441 function_builder.cc:109] BuilderBase::SetError; msg: Expected sel.9 to not contain a token in its type; type is (bits[1], token); loc: [(0,37,23)]
I0619 18:20:12.932035 1718441 function_converter.cc:370] Define node 'if true { (false, tok) } else { (true, tok) }' (Conditional) to be (nil) @ xls/modules/dbe/encoder_lz4.x:38:24-42:10
I0619 18:20:12.932047 1718441 function_converter.cc:419] Setting node 'if true { (false, tok) } else { (true, tok) }' (0x5562c8ed89f0) to IR value (nil).
I0619 18:20:12.932051 1718441 function_converter.cc:408] Using node 'if true { (false, tok) } else { (true, tok) }' (0x5562c8ed89f0) as IR value (nil).
[...]

Issue #143, opened 2020-09-22
Create standard Python unit testing base class for Python-based testbenches


cdleary:
Often people want to bail out to flexible non-synthesizable testbenches, and there's some boilerplate scaffolding currently required to do that, loading up the IR files and executing them and such. We should be able to hide it all inside of a base class, similar to how we did in the C++ test bench / JIT header generation scaffolding.

cdleary:
Since we have our `test_base` library in xls/common that calls `init_xls` appropriately, the adjusted task description here is probably to switch all of our Python tests to use that and make a "METADATA-file" (Goog internal presubmit check description) based warning for use of the raw underlying libraries.

cdleary:
Good use: https://github.com/google/xls/blob/62deead76f19000e143dd3dae47f13c84502815a/xls/ir/python/type_test.py#L24
Bad use: https://github.com/google/xls/blob/c41a2732b3a6918edd497fd118156d1214e5fb57/xls/ir/python/number_parser_test.py#L23

dplassgit:
These tests were removed.

PR #1038, opened 2023-06-20
WIP: LZ4 implementation in DSLX

rdob-ant:
This is a draft PR that contains most of work done on dictionary-based algorithm so far. As a draft PR, this is not supposed to be reviewed in full, but rather serve as a material for discussions.

The status as of now is:
- [x] Generic DBE token type hierarchy in Python and DSLX
- [x] Python: universal decoder
- [x] Python: LZ4 encoder implementation with compression ratio estimator & basic correctness test
- [x] DSLX: universal decoder with simple test. **TBD:** history buffer array must be replaced by RAM for it to be practical.
- [ ] DSLX: LZ4 encoder - basic port done, no tests yet. **TBD:** leads to SIGSEGV in IR converter.
- [ ] DSLX: LZ4 framer and deframer - not started yet.

### DIFF
--- a/xls/modules/dbe/lz4_encoder.x
+++ b/xls/modules/dbe/lz4_encoder.x
@@ -1,0 +1,494 @@
+// Copyright 2023 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import std
+import xls.examples.ram as ram
+import xls.modules.dbe.common as dbe
+
+type TokenKind = dbe::TokenKind;
+type Token = dbe::Token;
+type PlainData = dbe::PlainData;
+type RamReq = ram::SinglePortRamReq;
+type RamResp = ram::SinglePortRamResp;
+
+fn fifo_shift_left<FIFO_SIZE: u32, DATA_WIDTH: u32>(
+    fifo: uN[DATA_WIDTH][FIFO_SIZE],
+    data: uN[DATA_WIDTH]) -> uN[DATA_WIDTH][FIFO_SIZE] {
+   let fifo = for (i, arr): (u32, uN[DATA_WIDTH][FIFO_SIZE])
+               in u32:1..FIFO_SIZE {
+       update(arr, i - u32:1, fifo[i])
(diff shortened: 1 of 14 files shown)


xls/modules/dbe/lz4_encoder.x
```diff
@@ -0,0 +1,494 @@
...
+        //     i_ht_read_req, o_ht_read_resp, i_ht_write_req, o_ht_write_resp);
+
+        (
+            i_data, o_encoded,
+            o_ram_hb_req, i_ram_hb_resp,
+            o_ram_ht_req, i_ram_ht_resp,
+        )
+    }
+
+    next (tok: token, cur: State) {
```
rdob-ant, line 173: @proppy 

I think this is a good example of why #1037 is needed.

What I'm trying to do here:
- In some FSM states (e.g. START_MATCH_0), HB RAM is written (line 194), and then HT RAM is read (line 234).
- In other FSM states (e.g. START_MATCH_1), HT RAM is written (line 245), and then HB RAM is read (line 262).

It's easy to prove that above states are mutually exclusive. However, I still can not use two mutually exclusive send's or recv's to the same channel. Replacing them with `send_if` / `recv_if` manually is not enough: the current XLS limitation seems to be that I must provide a _single_ `send_if` or `recv_if` for each channel. There must not be two mutually exclusive `send_if`'s, there must be at most one `send_if` in the whole `next()` function for each channel. While it is possible to rewrite the code to fulfill that constraint, the code most probably will get quite cumbersome and dirty as a result.. 

I see another option: instead of relying on 1RW memory, I should be able to use 1R1W memories here, and because those have separate channels for read and write requests, I may be able to figure out something that makes XLS happy without much code reshuffling.

Do you know what's the status of 1R1W RAM Verilog codegen support? From XLS soruces it seems that it may be supported, but I haven't found any 1R1W DSLX example that actually generates Verilog.

proppy:
> leads to SIGSEGV in IR converter.

is there a bug already filed about this?

proppy:
Thanks for including the python reference implementation!

It would be interesting to including as a https://bazel.build/reference/be/python#py_test rule that test the its output against the dslx implementation, similar to what we have in C++ for some examples and modules, see:
https://github.com/google/xls/tree/main/xls/modules/aes
https://github.com/google/xls/blob/main/third_party/xls_colors/test.cc

There are existing private python bindings to the DSLX and IR interpreter that could be leveraged for doing this:
- https://github.com/google/xls/tree/main/xls/dslx/python
- https://github.com/google/xls/tree/main/xls/ir/python
- see also https://github.com/google/xls/issues/143
(I don't think we have any binding for the JIT yet, so filed https://github.com/google/xls/issues/1041)

Note:  there is also an embryo of a public python API there https://github.com/google/xls/tree/main/xls/public/python but it's currently very limited and is more targeted toward external usecases, rather than testing module that are maintained in the tree.

tmichalak:
> > leads to SIGSEGV in IR converter.
> 
> is there a bug already filed about this?

Filed in #1035

proppy:
is this superseeded by #1075?

rdob-ant:
@proppy Correct. Closing this one.

PR #1052, opened 2023-07-03
Add multisymbol RLE implementation

mtdudek:
This PR adds more advanced implementation of RunLengthEncoder that is capable of ingesting
multiple symbols at once and produce multiple pairs in output.

This PR depends on #1006 
It updates implementations of simple encoder and simple decoder to work with new
channel description.

Main differentiating factor is that original RLE allows for symbol width to be adjusted.
This implementation adds ability to process multiple symbols at once.
Let's take a base RLE and multisybol RLE with input that takes up to 4 symbols and outputs up to 2 pairs
and following sequence:`AABCCCCB`
* base RLE has to ingest symbols one by one. It takes 8 channel transactions `A, A, B, C, C, C, C, B`.
Generating output takes require 5 transactions: `(2,A), (1,B), (3,C), (1,C), (1,B)`
* multisymbol RLE will ingest symbols in 2 transactions as `AABC, CCCB`.
Generating output takes only 3 transactions: `[(2,A), (1,B)], [(3,C), (1,C)], [(1,B)]`

Multisybol RL encoder should have higher throughput but wt will use more logic to implement.

### DIFF
--- a/xls/modules/rle/rle_dec.x
+++ b/xls/modules/rle/rle_dec.x
@@ -63,8 +63,8 @@ struct RunLengthDecoderState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
 }
  // RLE decoder implementation
 pub proc RunLengthDecoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
-  input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH>> in;
-  output_s: chan<DecOutData<SYMBOL_WIDTH>> out;
+  input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH, 1>> in;
+  output_s: chan<DecOutData<SYMBOL_WIDTH, 1>> out;
 
   init {(
     RunLengthDecoderState<SYMBOL_WIDTH, COUNT_WIDTH> {
@@ -75,34 +75,38 @@ pub proc RunLengthDecoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
   )}
 
   config (
-    input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH>> in,
-    output_s: chan<DecOutData<SYMBOL_WIDTH>> out,
+    input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH, 1>> in,
+    output_s: chan<DecOutData<SYMBOL_WIDTH, 1>> out,
   ) {(input_r, output_s)}
 
   next (tok: token, state: RunLengthDecoderState<SYMBOL_WIDTH, COUNT_WIDTH>) {
     let state_input = DecInData {
-      symbol: state.symbol,
-      count: state.count,
+      symbols: [state.symbol],
+      counts: [state.count],
       last: state.last
     };
     let recv_next_symbol = (state.count == bits[COUNT_WIDTH]:0);
     let (tok, input) = recv_if(tok, input_r, recv_next_symbol, state_input);
-    let next_count = if input.count == bits[COUNT_WIDTH]:0 {
-        fail!("invalid_count_0", input.count)
+    let next_count = if input.counts[0] == bits[COUNT_WIDTH]:0 {
+        input.counts[0]
     } else {
-        input.count - bits[COUNT_WIDTH]:1
+        input.counts[0] - bits[COUNT_WIDTH]:1
     };
     let done_sending = (next_count == bits[COUNT_WIDTH]:0);
     let send_last = input.last && done_sending;
-    let data_tok = send(tok, output_s, DecOutData {
-      symbol: input.symbol,
-      last: send_last
+    let symbol_valid = input.counts[0] > bits[COUNT_WIDTH]:0;
+    let data_tok = send_if(tok, output_s,
+      symbol_valid || send_last,
+      DecOutData {
+        symbols: [input.symbols[0]],
+        symbol_valids: [symbol_valid],
+        last: send_last
     });
     if (send_last) {
       zero!<RunLengthDecoderState>()
     } else {
       RunLengthDecoderState {
-        symbol: input.symbol,
+        symbol: input.symbols[0],
         count: next_count,
         last: input.last,
       }
@@ -116,8 +120,8 @@ proc RunLengthDecoder32 {
   init {()}
 
   config (
-    input_r: chan<DecInData<32, 2>> in,
-    output_s: chan<DecOutData<32>> out,
+    input_r: chan<DecInData<32, 2, 1>> in,
+    output_s: chan<DecOutData<32, 1>> out,
   ) {
     spawn RunLengthDecoder<u32:32, u32:2>(input_r, output_s);
     ()
@@ -136,8 +140,8 @@ const TEST_COUNT_WIDTH  = u32:32;
 type TestSymbol     = bits[TEST_SYMBOL_WIDTH];
 type TestCount      = bits[TEST_COUNT_WIDTH];
 type TestStimulus   = (TestSymbol, TestCount);
-type TestDecInData  = DecInData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>;
-type TestDecOutData = DecOutData<TEST_SYMBOL_WIDTH>;
+type TestDecInData  = DecInData<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH, 1>;
+type TestDecOutData = DecOutData<TEST_SYMBOL_WIDTH, 1>;
 
 // Check RLE decoder on a transaction
 #[test_proc]
@@ -171,13 +175,13 @@ proc RunLengthDecoderTransactionTest {
         in enumerate(TransactionTestStimuli) {
       let last = counter == (array_size(TransactionTestStimuli) - u32:1);
       let data_in = TestDecInData{
-        symbol: stimulus.0,
-        count: stimulus.1,
+        symbols: [stimulus.0],
+        counts: [stimulus.1],
         last: last
       };
       let tok = send(tok, dec_input_s, data_in);
       trace_fmt!("Sent {} stimuli, symbol: 0x{:x}, count:{}, last: {}",
-          counter + u32:1, data_in.symbol, data_in.count, data_in.last);
+          counter + u32:1, data_in.symbols[0], data_in.counts[0], data_in.last);
       (tok)
     }(tok);
     let TransationTestOutputs: TestSymbol[14] = [
@@ -194,18 +198,94 @@ proc RunLengthDecoderTransactionTest {
         in enumerate(TransationTestOutputs) {
       let last = counter == (array_size(TransationTestOutputs) - u32:1);
       let data_out = TestDecOutData{
-        symbol: symbol,
+        symbols: [symbol],
+        symbol_valids: [bits[1]:1],
         last: last
       };
       let (tok, dec_output) = recv(tok, dec_output_r);
       trace_fmt!(
           "Received {} transactions, symbol: 0x{:x}, last: {}",
-          counter, dec_output.symbol, dec_output.last
+          counter, dec_output.symbols[0], dec_output.last
       );
       assert_eq(dec_output, data_out);
       (tok)
     }(tok);
     send(tok, terminator, true);
+  }
+}
+
+// Check that RLE decoder will remove empty pairs, `count == 0`.
+// Check that RLE decoder will set `symbol_valids` to 0 only in
+// the last output packet.
+#[test_proc]
+proc RunLengthDecoderZeroCountTest {
+  terminator: chan<bool> out;            // test termination request
+  dec_input_s: chan<TestDecInData> out;
+  dec_output_r: chan<TestDecOutData> in;
+
+  init {()}
+
+  config(terminator: chan<bool> out) {
+    let (dec_input_s, dec_input_r)   = chan<TestDecInData>;
+    let (dec_output_s, dec_output_r) = chan<TestDecOutData>;
+
+    spawn RunLengthDecoder<TEST_SYMBOL_WIDTH, TEST_COUNT_WIDTH>(
+        dec_input_r, dec_output_s);
+    (terminator, dec_input_s, dec_output_r)
+  }
+
+  next(tok: token, state: ()) {
+    let ZeroCountTestStimuli: TestStimulus[6] =[
+      (TestSymbol:0xB, TestCount:0x2),
+      (TestSymbol:0x1, TestCount:0x0),
+      (TestSymbol:0xC, TestCount:0x1),
+      (TestSymbol:0xC, TestCount:0x0),
+      (TestSymbol:0x3, TestCount:0x3),
+      (TestSymbol:0x2, TestCount:0x0),
+    ];
+    let tok = for ((counter, stimulus), tok):
+        ((u32, (TestSymbol, TestCount)) , token)
+        in enumerate(ZeroCountTestStimuli) {
+      let last = counter == (array_size(ZeroCountTestStimuli) - u32:1);
+      let data_in = TestDecInData{
+        symbols: [stimulus.0],
+        counts: [stimulus.1],
+        last: last
+      };
+      let tok = send(tok, dec_input_s, data_in);
+      trace_fmt!("Sent {} stimuli, symbol: 0x{:x}, count:{}, last: {}",
+          counter + u32:1, data_in.symbols[0], data_in.counts[0], data_in.last);
+      (tok)
+    }(tok);
+    let ZeroCountTestOutputs: TestDecOutData[7] = [
+      TestDecOutData{symbols: [TestSymbol: 0xB],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0xB],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0xC],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0x3],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0x3],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0x3],
+                     symbol_valids: [true], last: false},
+      TestDecOutData{symbols: [TestSymbol: 0x2],
+                     symbol_valids: [false], last: true},
+    ];
+    let tok = for ((counter, output), tok):
+        ((u32, TestDecOutData) , token)
+        in enumerate(ZeroCountTestOutputs) {
+      let (tok, dec_output) = recv(tok, dec_output_r);
+      trace_fmt!(
+          "Received {} transactions, symbols: 0x{:x}, last: {}",
+          counter + u32:1, dec_output.symbols, dec_output.last
+      );
+      assert_eq(dec_output, output);
+      (tok)
+    }(tok);
+    send(tok, terminator, true);
+    ()
   }
 }
 
@@ -231,13 +311,13 @@ proc RunLengthDecoderLastAfterLastTest {
   next(tok: token, state: ()) {
     let LastAfterLastTestStimuli: TestDecInData[2] =[
       TestDecInData {
-        symbol: TestSymbol:0x1,
-        count: TestCount:0x1,
+        symbols: [TestSymbol:0x1],
+        counts: [TestCount:0x1],
         last:true
       },
       TestDecInData {
-        symbol: TestSymbol:0x2,
-        count: TestCount:0x1,
+        symbols: [TestSymbol:0x2],
+        counts: [TestCount:0x1],
         last:true
       },
     ];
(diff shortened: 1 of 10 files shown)


proppy:
can you rebase?

mtdudek:
@proppy I've rebased it on to the newest main

proppy:
> implementation of RunLengthEncoder that is capable of ingesting multiple symbols at once and produce multiple pairs in output

Can you detail a little a bit more in the PR description (or the `.x` header) on how this differ from the original implementation, maybe with examples of input / output; the original implementation did support variable symbol width: so it'd be good to highligh how "multisymbol" differs from "larger symbol".

mtdudek:
Main differentiating factor is that original RLE allows for symbol width to be adjusted.
This implementation adds ability to process multiple symbols at once.
Let's take a base RLE and multisybol RLE with input that takes up to 4 symbols and outputs up to 2 pairs
 and following sequence:
`AABCCCCB`
* base RLE has to ingest symbols one by one. It takes 8 channel transactions `A, A, B, C, C, C, C, B`.
Generating output takes require 5 transactions: `(2,A), (1,B), (3,C), (1,C), (1,B)`
* multisymbol RLE will ingest symbols in 2 transactions as `AABC, CCCB`.
Generating output takes only 3 transactions: `[(2,A), (1,B)], [(3,C), (1,C)], [(1,B)]`

Multisybol RL encoder should have higher throughput but wt will use more logic to implement.

mtdudek:
I'll update `.x` file with similar explanation and waveform.

proppy:
> [(2,A), (1,B)], [(3,C), (1,C)], [(1,B)]

Curious why there is three transaction here, and two separate packet for `C`?

xls/modules/rle/rle_enc.x
```diff
@@ -66,8 +66,8 @@ struct RunLengthEncoderState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
 
 // RLE encoder implementation
```
proppy, line 67: curious why we need the original implementation to change?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// repeating symbols to the output stream that contains the symbols and
+// the number of its consecutive occurrences in the input stream.
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 processes.
+// 1. Reduce stage - this process takes incoming symbols and symbol_valid
+// and reduces them into symbol count pairs. This stage is stateless.
```
proppy, line 27: if the stage is stateless, does it need to be a proc?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+//
+// The encoder uses Run Length Encoding (RLE) to compress the input stream of
+// repeating symbols to the output stream that contains the symbols and
+// the number of its consecutive occurrences in the input stream.
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 processes.
```
proppy, line 25: s/processes/proc/

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// 1. Reduce stage - this process takes incoming symbols and symbol_valid
```
proppy, line 26: It's a bit misleading to call those `stage` as XLS itself might split them in multiple pipeline stages.

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// and reduces them into symbol count pairs. This stage is stateless.
+// 2. Realign stage - this process moves pairs emitted from previous stage
+// so that they are align to the left, it also calculates propagation distance
+// for the first pair.
+// 3. Core stage - this stage is stateful. It takes align pairs,
+// and combines them with its state.It outputs multiple symbol/count pairs.
```
proppy, line 32: nit: space after punctuation.

xls/modules/rle/rle_enc_adv.x
proppy, line 27: isn't this stage equivalent to the current `rle_enc` can the implementation be shared?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 processes.
+// 1. Reduce stage - this process takes incoming symbols and symbol_valid
+// and reduces them into symbol count pairs. This stage is stateless.
+// 2. Realign stage - this process moves pairs emitted from previous stage
+// so that they are align to the left, it also calculates propagation distance
```
proppy, line 29: can you give an example of what 'align to the left` refer to?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// for the first pair.
+// 3. Core stage - this stage is stateful. It takes align pairs,
```
proppy, line 31: s/align/aligned/

proppy, line 31: can you document what the state is?

xls/modules/rle/rle_enc_adv.x
proppy, line 32: can you give an example of the output to show how it different from the received input?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// and combines them with its state.It outputs multiple symbol/count pairs.
+// 4 - Adjust Width stage - this stage takes output from the core stage.
```
proppy, line 33: s/4 - /4./

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// the number of its consecutive occurrences in the input stream.
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 processes.
+// 1. Reduce stage - this process takes incoming symbols and symbol_valid
+// and reduces them into symbol count pairs. This stage is stateless.
+// 2. Realign stage - this process moves pairs emitted from previous stage
```
proppy, line 28: does it takes input from the preceding "stage"?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// so that they are align to the left, it also calculates propagation distance
+// for the first pair.
+// 3. Core stage - this stage is stateful. It takes align pairs,
+// and combines them with its state.It outputs multiple symbol/count pairs.
+// 4 - Adjust Width stage - this stage takes output from the core stage.
+// If output can handle more or equal number of pairs as
```
proppy, line 34: How does the "stage" know that the "output can handle more or equal the number of pairs", can you give an example

xls/modules/rle/rle_enc.x
mtdudek: This PR modifies existing interfaces for symbols and encoded pairs.
It adds parameter for number of symbols/pairs in single transaction.
It also adds symbol_valids.
This is why I had to update existing implementations.

xls/modules/rle/rle_enc_adv.x
mtdudek: As far as I understand DSLX if I want to perform channel operations I have to use proc's.

mtdudek:
> > [(2,A), (1,B)], [(3,C), (1,C)], [(1,B)]
> 
> Curious why there is three transaction here, and two separate packets for `C`?

This was with an assumption that RLE counter is 2 bit wide, so it can only hold at most 3 symbols, while C was repeated 4 times.

xls/modules/rle/rle_enc_adv.x
mtdudek: You are right, that `stage` is too verloaded in this context,
as it can refer to HW pipeline, processing pipeline.
What do you think about about calling it `step`?

xls/modules/rle/rle_enc_adv.x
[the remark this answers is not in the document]
mtdudek: Let's assume that count is 2 bit wide and input can ingest 4 symbols at once.
Let the input be `[A,A,A,A]`, so after the second stage it will look like:
`[(A,3), (A,1), ..., ...]` and propagation distance will be 1.
For input `[A,B,B,B]` second step will emit `[(A,1), (B,3), ..., ...]` with propagation distance 0.

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,472 @@
...
+// input number of symbols. This stage does nothing.
+// If the output is narrower than the input,
+// this stage will serialize symbol counter pairs.
```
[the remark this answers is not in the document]
mtdudek: Let input width be 4 and output with is 2. Core step can emit the following : `[(A,1), (B,1) , (A,1), (B,1)]`.
Adjust Width step will serialize it to `[(A,1(, (B,1)], [(A,1), (B,1)]`

xls/modules/rle/rle_enc_adv.x
mtdudek: Output and input widths are parametrized `INPUT_WIDTH` and `OUTPUT_WIDTH`.
If `OUTPUT_WIDTH` is greater or equal to `INPUT_WIDTH,  adjust width step has nothing to adjust.

xls/modules/rle/rle_enc_adv.x
mtdudek: I can't see any obvious code that can be shared between `reduce step` and `rle_enc`.
Main difference is that `rle_enc` code has to take into account state and update it f necessary,
while reduce step doesn't have one.

xls/modules/rle/rle_enc_adv.x
mtdudek: Fixed

xls/modules/rle/rle_enc_adv.x
mtdudek: Fixed

xls/modules/rle/rle_enc_adv.x
mtdudek: Fixed

xls/modules/rle/rle_enc_adv.x
proppy: proc?

xls/modules/rle/rle_dec.x
```diff
@@ -63,8 +63,8 @@ struct RunLengthDecoderState<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
 }
  // RLE decoder implementation
 pub proc RunLengthDecoder<SYMBOL_WIDTH: u32, COUNT_WIDTH: u32> {
-  input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH>> in;
-  output_s: chan<DecOutData<SYMBOL_WIDTH>> out;
+  input_r: chan<DecInData<SYMBOL_WIDTH, COUNT_WIDTH, 1>> in;
```
proppy, line 66: I wonder if it would be more friendly to parameterize with `max_count` and use https://google.github.io/xls/dslx_std/#stdclog2 + 1 to represent the width?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,485 @@
...
+
+// This encoder is implemented as a net of 4 proc.
+// 1. Reduce proc - this process takes incoming symbols and symbol_valid
+// and reduces them into symbol count pairs. This step is stateless.
+// 2. Realign proc - this process moves pairs emitted from the reduce step
+// so that they are aligned to the left, it also calculates propagation
+// distance for the first pair.
+// Example behaviours:
+// 1)
+//    input:  [.., (A, 2), .., (B, 2)]
```
proppy, line 33: what does `..` means here?

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,485 @@
...
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 proc.
+// 1. Reduce proc - this process takes incoming symbols and symbol_valid
+// and reduces them into symbol count pairs. This step is stateless.
+// 2. Realign proc - this process moves pairs emitted from the reduce step
+// so that they are aligned to the left, it also calculates propagation
+// distance for the first pair.
```
proppy, line 30: can you also explain what "propagation distance" mean here? (it's not clear, at least to me, from just reading the example)

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,485 @@
...
+// Example behaviours:
+// 1)
+//    input:  [.., (A, 2), .., (B, 2)]
+//    output: [(A, 2), (B, 2), .., ..]
+//    propagation distance: 0
+// 2)
+//    input:  [.., .., (A, 3), (A, 1)]
```
proppy, line 37: `(A, 3), (A, 1)` can this actually happen? wouldn't the previous step simply emit `(A, 4)`?

xls/modules/rle/rle_enc_adv.x
proppy: see the comment below can you also explain what `...` refers to?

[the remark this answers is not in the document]
proppy: see the comment below, can you update the comment with a textual explanation of what the propagation distance refer to? (here is seems that it mainly indicate with a packet has been broken or not)

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,485 @@
...
+//    output: [(A, 3), (A, 1), .., ..]
+//    propagation distance: 1
+// 3. Core proc - this step is stateful. It takes align pairs from
+// the realign step, and combines them with its state to create multiple
+// symbol/count pairs output. State is represented by following tuple
+// `<symbol, count, last>`. It contains symbol and count from last pair
```
proppy, line 43: does it effectively means that we're counting again? is that right to assume that if each transaction only include one symbol this does nothing?

xls/modules/rle/rle_enc_adv.x
mtdudek: Here `...` was used for an invalid data symbol, counter pair.

xls/modules/rle/rle_enc_adv.x
```diff
@@ -0,0 +1,485 @@
...
+// The encoder uses Run Length Encoding (RLE) to compress the input stream of
+// repeating symbols to the output stream that contains the symbols and
+// the number of its consecutive occurrences in the input stream.
+// Both the input and the output channels use additional `last` flag
+// that indicates whether the packet ends the transmission. After sending
+// the last packet encoder dumps all the data to the output stream.
+// The behavior of the encoder is presented on the waveform below:
+
+// This encoder is implemented as a net of 4 proc.
+// 1. Reduce proc - this process takes incoming symbols and symbol_valid
```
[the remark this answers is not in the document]
mtdudek: It's not exactly the same thing as existing RLE enc. Existing RLE has state that is combined with incoming data,
this stage doesn't have state, it only works on data visible in a single channel payload and doesn't include stream history.

xls/modules/rle/rle_enc_adv.x
mtdudek: It was to mark invalid data and data, count pairs.

xls/modules/rle/rle_enc_adv.x
mtdudek: In the examples I was assuming counter 2 bit wide, so that's why there is `(A, 3), (A, 1)` and not `(A, 4)`.

xls/modules/rle/rle_enc_adv.x
mtdudek: This is the first proc with a state that holds stream history. If a stream would only have one data value in all symbols.
First processing state would reduce incoming data stream into  `(data_value, input_count)`, this stage would combine this pair with one that preceded it, and only emit output pair when either internal counter overflows or stream ends.

xls/modules/rle/rle_enc_adv.x
mtdudek: Here propagation distance is the number of consecutive pairs, starting form first pair, that have the same symbol value as the first one and have their counter maxed out. This way 'Core proc' can only check "propagation distance"th pair value and counter. 
* If value differs from the internal state, then all pairs from [0:"propagation distance") will differ and core should combine payload with its state.
* If value is equal internal state, then  "Core proc" only checks if internal state counter + "propagation distance"th pair counter overflow.

xls/modules/rle/rle_enc_adv.md
```diff
@@ -0,0 +1,85 @@
...
+### Process
+1. Reduce step - this process takes incoming symbols and symbol_valid
+and reduces them into symbol count pairs. This step is stateless.
+
+Example:
+
+|||
+|-----|-------|
+|input|output |
+|[(A, True), (A, True), (A, True), (A, True)]|[.., .., (A, 3), (A, 1)]|
```
proppy, line 33: is the boolean after the symbol corresponding to `!last`?

xls/modules/rle/rle_enc_adv.md
```diff
@@ -0,0 +1,85 @@
+# Multisymbol Run-length Encoder
+
+The encoder uses Run Length Encoding (RLE) to compress the input stream of
+repeating symbols to the output stream that contains the symbols and
+the number of its consecutive occurrences in the input stream.
+
+Overall, we can break down the data processing into four stages: reduction, alignment, compression, and output generation. The division of responsibility allowed the specialized blocks to efficiently process data and made it possible test each functionality separately.
```
proppy, line 7: did we measure the overhead of splitting the process into different proc in term of additional registers and wires, compared to a single proc implementation?

xls/modules/rle/rle_enc_adv.md
```diff
@@ -0,0 +1,85 @@
...
+## Encoder processing pipeline detailed breakdown.
+
+### Initial conditions
+- input width is 4 symbols wide,
+- output width is 2 pairs wide,
+- symbol counter is 2 bits wide.
+
+### Process
+1. Reduce step - this process takes incoming symbols and symbol_valid
+and reduces them into symbol count pairs. This step is stateless.
```
proppy, line 26: couldn't stateless proc be in theory implemented as `fn`?

xls/modules/rle/rle_enc_adv.md
mtdudek: I haven't checked how much more resource does multi process implementation uses over single process implementation, and how max frequency is affected by this.

xls/modules/rle/rle_enc_adv.md
```diff
@@ -0,0 +1,85 @@
...
+
+Input width is defined using the `INPUT_WIDTH` parameter and output width is defined with the `OUTPUT_WIDTH` parameter.
+Both the input and the output channels use additional `last` flag
+that indicates whether the packet ends the transmission. After sending
+the last packet the encoder dumps all the data to the output stream.
+
+## Encoder processing pipeline detailed breakdown.
+
+### Initial conditions
+- input width is 4 symbols wide,
```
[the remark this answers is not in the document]
mtdudek: Yes, implementation assumes that the input is always x symbols wide.
Handling u9 to u2 conversion should be done by additional proc.
As it will know how to handle that extra one bit,
if that bit is a part of symbol that was split on u9 boundary, or it has some other meaning.

xls/modules/rle/rle_enc_adv.md
mtdudek: All state less processes are effectively a `fn`s, but as far as I know you can't put `recv/send` operations into `fn`s.
Using `fn ` would potentially reduce fmax, as first proc would have to do all the processing before it would be able to update it's state, or we would reduce pipeline throughput when using II>1 to improve fmax.

xls/modules/rle/rle_enc_adv.md
mtdudek: Thanks for catch this. I've forgotten to put last in this example.
This boolean value next to the symbol is data valid value.
I'll update this example to include last value

proppy:
Should we close this in favor of the work on #1038 ?

proppy:
@tmichalak closing this now that the more relevant #1315 has been merged.

PR #1073, opened 2023-07-20
[RLE] Cleanup RLE encoder and decoder

mtdudek:
This PR is a followup to #1006 and #974.
It cleanups:
- `xls/modules` directory;
- renames rle to rle_enc;
- extracts common interfaces and moves them to rle_common.x.


mtdudek:
wdyt? @proppy

proppy (approved):

Issue #995, opened 2023-05-30
add DSLX sample for Dictionary Based Encoding (DBE) encoder/decoder


proppy:
See https://en.wikipedia.org/wiki/Dictionary_coder

This will help demonstrate how to implement streaming algorithms that heavily rely on stateful memory with DSLX procs.

proppy:
Possible algorithms researched:
- [Adaptive Lossless Data Compression Algorithm](https://www.ecma-international.org/wp-content/uploads/ECMA-222_1st_edition_june_1995.pdf) (ALDC) w/ known hardware implementations ([1](https://ieeexplore.ieee.org/document/5389240), [2](https://ieeexplore.ieee.org/document/10019048))
- https://github.com/google/brotli
- https://facebook.github.io/zstd/

rdob-ant:
In addition to the above, we're also considering two other algorithms:
- **DEFLATE** - very commonly used (e.g. in .zip / .gz file formats), based on LZ77-like dictionary coder used together with Huffman encoding.
- **LZ4** - based on pure LZ77-like dictionary coder. Commonly used where especially fast decompression is required (that comes at the expense of reduced compression ratio compared to DEFLATE/Brotli/Zstd).

For me, it is tempting to go with LZ4. Due to its relatively widespread use it would make a much more "useful" DSLX sample, while still being similar to the initially considered ALDC (so most of our ALDC-related ideas are still applicable), and at the same time not resulting in much scope creep (e.g. adding Huffman or entropy coders/decoders on top of the dictionary coder - which would be required for most other algorithms).

proppy:
Also learned about the following today:
- https://sneller.io/blog/decompressing-at-over-10-gigabytes-per-second/
- https://github.com/inikep/lizard
- https://github.com/SnellerInc/sneller/tree/master/ion/zion/iguana

rdob-ant:
Opened draft #1038 which shows current state of matters

PR #1075, opened 2023-07-21
modules/dbe - Lz4 decoder

rdob-ant:
This is an implementation of Lz4 decoder in DSLX and accompanying cocotb testsuite. Since Lz4 encoder reuses much of the code from this PR, I'd like to prepare it after this one is reviewed.

Some notes:
1. This relies on cocotb, so includes commits from #1057 . Once that is merged, this branch will be rebased.
2. DSLX implementation of decoder uses 1R1W RAM with 1 partition. Such configuration triggers a small bug (missing mask expansion) in examples/ram.x that this PR fixes.
3. It also includes reference decoder implementation in Python which is not used by the tests in the current PR, but will be needed for the upcoming Lz4 encoder PR.

### DIFF
--- a/dependency_support/pip_requirements.txt
+++ b/dependency_support/pip_requirements.txt
@@ -8,6 +8,8 @@ termcolor==1.1.0
 psutil==5.7.0
 portpicker==1.3.1
 pyyaml==5.4.1
+cocotb==1.8.0
+cocotb_bus==0.2.1
 
 # Note: numpy and scipy version availability seems to differ between Ubuntu
 # versions that we want to support (e.g. 18.04 vs 20.04), so we accept a
--- a/dependency_support/rules_hdl/workspace.bzl
+++ b/dependency_support/rules_hdl/workspace.bzl
@@ -29,9 +29,9 @@ def repo():
         ],
     )
 
-    # Commit on 2023-06-13, current as of 2023-06-13.
-    git_hash = "e6540a5bccbfb124aec0b19deaa9cf855781b3a5"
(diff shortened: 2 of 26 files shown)


proppy:
Related to #995.

proppy:
> This relies on cocotb, so includes commits from https://github.com/google/xls/pull/1057 . Once that is merged, this branch will be rebased.

Maybe we could move this to a separate PR, as it is not strictly required by the implementation (and not to block this being merged on the resolution of the discussions happening in #1057)

proppy:
> Lz4 encoder reuses much of the code from this PR, I'd like to prepare it after this one is reviewed.

I'd be good to get the encoder and the decoder in the same PR, in order to be able to more effectively cross-review them. If it's challenging to send both at the same time, I think it'd be easier to review the encoder first.

rdob-ant:
> can we get a separate PR for that (so that we can merge it earlier)

Opened #1078

rdob-ant:
> Maybe we could move this to a separate PR, as it is not strictly required by the implementation (and not to block this being merged on the resolution of the discussions happening in https://github.com/google/xls/pull/1057)

We can, but since I wrote most of the tests in cocotb, I'm afraid without them the implementation may look "incomplete".

rdob-ant:
> Do you plan to validate the dslx/ir/codegen versions against the python reference impl?

Yes, at least in one test I'm using Python decoder to decode (and thus validate) the output of DSLX encoder.

rdob-ant:
Superseded by #1079 that includes both encoder and decoder.

PR #1076, opened 2023-07-23
Make boost libbacktrace dependency hermetic.

hzeller:
This uses Ethans [libbacktrace support in bazel_rules_hdl](https://github.com/hdl/bazel_rules_hdl/commit/196706dee7fdfeb0ddf916f8a8bd125e0ed0ab45) (that we already depend on).

Without, XLS compilation will fail if the libbacktrace doesn't happen to be part of the local compiler set-up:

```
external/boost/libs/stacktrace/include/boost/stacktrace/detail/libbacktrace_impls.hpp:23:13: fatal error: backtrace.h: No such file or directory
   23 | #   include <backtrace.h>
      |             ^~~~~~~~~~~~~
compilation terminated.
```


hzeller:
CC @QuantamHD

cdleary (approved):

PR #1078, opened 2023-07-25
bugfix: examples/ram: 1R1W - fix missing mask expansion

rdob-ant:
Current code does not perform mask width expansion when processing read requests. As a result, configurations with `NUM_PARTITIONS != DATA_WIDTH` fail to compile.


proppy:
> As a result, configurations with NUM_PARTITIONS != DATA_WIDTH fail to compile.

Can we add a test that capture this?

rdob-ant:
Added two things:
1. A simple test for masked writes and reads.
2. Bugfix for the initialization check in case of partial masked reads (that bug manifested itself when running the test. It did not in DBE code because we never do partial reads).

PR #1092, opened 2023-08-03
modules/dbe - lz4 encoder and decoder [DSLX code only]

rdob-ant:
This is an implementation of Lz4 encoder and decoder in DSLX without reference Python & cocotb test code that was present in  #1079.

The source branch additionally includes changes from #1078 that are needed to run DSLX tests of procs using 1R1W single-partition RAM. These will be dropped by rebasing when we get that PR merged.


tmichalak:
CC @proppy @hongted

hongted (approved):

xls/examples/ram.x
```diff
@@ -249,9 +249,18 @@ proc RamModel<DATA_WIDTH:u32, SIZE:u32, WORD_PARTITION_SIZE:u32={u32:0},
     // Assert memory being read is initialized by checking that all partitions
     // have been initialized.
     if read_req_valid && ASSERT_VALID_READ {
-      assert_eq(
-        mem_initialized[read_req.addr],
-        std::convert_to_bools_lsb0(read_req.mask))
+      let init_parts = mem_initialized[read_req.addr];
+      let requested_is_init = for (idx, requested_is_init): (u32, bool) in
+      range(u32:0, NUM_PARTITIONS) {
```
hongted, line 254: Add 1 level of indentation.

xls/examples/ram.x
```diff
@@ -249,9 +249,18 @@ proc RamModel<DATA_WIDTH:u32, SIZE:u32, WORD_PARTITION_SIZE:u32={u32:0},
...
+        if read_req.mask[idx+:bool] {
+          if !init_parts[idx] {
+            trace_fmt!("reading partitions {} but part={} is not init: {}",
+              std::convert_to_bools_lsb0(read_req.mask), idx, init_parts);
+          } else { () };
+          requested_is_init && init_parts[idx]
+        } else { requested_is_init }
+      } (true);
```
hongted, line 262: This seems general enough to suggest a separate function and unit test.

hongted: Also it seems that this can be simplified to just 
let mem_initialized_as_bits = std::convert_to_bits_lsb0(mem_initialized[req_req.addr]);

assert_eq(mem_initialized_as_bits, mem_initialized_as_bits | read_req.mask)

hongted (changes requested):

xls/modules/dbe/BUILD
```diff
@@ -0,0 +1,156 @@
...
+
+# ---------------------------------------------------------------------------
+# Common
+# ---------------------------------------------------------------------------
+
+xls_dslx_library(
+    name = "dbe_common_dslx",
+    srcs = [
+        "common.x",
+        "common_test.x"
```
hongted, line 40: Move common_test into a xls_dslx_test rule.

hongted (changes requested):
Can a round-trip test be added?  

i.e.  

Create a round-trip test proc that spawns an encoder and decoder back to back.
Give it some arbitrary plain text and assert that what goes in comes out.

xls/examples/ram.x
rdob-ant: Hi @hongted . Sure, will do that. To keep things clean, I propose to discuss _ram.x_ changes in a PR dedicated to that bugfix: https://github.com/google/xls/pull/1078 . I will address your comments directly there.

This change is present in the current PR only because Lz4 DSLX tests depends on the RAM bugfix, and I'm not aware of a way of submitting Lz4 changes for review without making commits from a dependency PR visible here.

xls/modules/dbe/common.x
```diff
@@ -0,0 +1,75 @@
...
+    // Literal fields
+    lt_sym: uN[SYM_WIDTH],
+    // Copy pointer fields
+    cp_off: uN[PTR_WIDTH],
+    cp_cnt: uN[CNT_WIDTH],
+    // Marker fields
+    mark: Mark
+}
+
+pub fn zero_token<SYM_WIDTH: u32, PTR_WIDTH: u32, CNT_WIDTH: u32>()
```
hongted, line 52: Suggest using the builtin zero!<>.  This is a common comment.

xls/modules/dbe/lz4_decoder.x
```diff
@@ -0,0 +1,618 @@
...
+        } else if is_copying {
+            state.rd_ptr + uN[PTR_WIDTH]:1
+        } else {
+            state.rd_ptr
+        };
+
+        // Check whether rd_ptr points to an unwritten HB cell
+        let rd_ptr_valid = (rd_ptr < state.wr_ptr) || state.hb_all_valid;
+
+        // HB READ - read the symbol from history buffer
```
hongted, line 206: Does this handle the "overlap match" scenario?

hongted, line 206: Does this handle the "overlap match" scenario referenced in https://github.com/lz4/lz4/blob/dev/doc/lz4_Block_format.md when the symbol from the history buffer depends on the match copy?

xls/modules/dbe/lz4_decoder.x
```diff
@@ -0,0 +1,618 @@
...
+            })
+        } else {
+            (false, zero_data)
+        };
+        let tok = send_if(tok, o_data, data_valid, data);
+
+        // HB WRITE - write emitted symbol to the history buffer
+        let do_hb_write = data_valid && !data.is_mark;
+        let tok = send_if(tok, o_ram_hb_wr_req, do_hb_write,
+            ram::WriteWordReq<u32:1>(state.wr_ptr, data.data));
```
hongted, line 272: LZ4 is a byte-oriented compression right?  Should there be some use of masks to handle a WordReq'ed sized write?

xls/modules/dbe/lz4_decoder.x
```diff
@@ -0,0 +1,618 @@
...
+        let (tok, hb_rd) = recv_if(tok, i_ram_hb_rd_resp, do_hb_read,
+            RamZeroReadResp<SYM_WIDTH>());
+
+        // Generate ERROR_INVAL_CP when we should've read from HB but
+        // we couldn't because CP points to an invalid location
+        let err_inval_cp = emit_sym_from_hb && !do_hb_read;
+
+        // Send decoded data symbol or Mark
+        let zero_data = dbe::zero_plain_data<SYM_WIDTH>();
+        let (data_valid, data) = if rx_reset {
```
hongted, line 219: rx_rest is more about clearing the state to handle a new block, right?  If so, consider changing the name to reflect that.

xls/modules/dbe/lz4_decoder.x
```diff
@@ -0,0 +1,618 @@
...
+    PlainData{is_mark: false, data: uN[4]:9, mark: Mark::NONE},
+    PlainData{is_mark: false, data: uN[4]:2, mark: Mark::NONE},
+    PlainData{is_mark: false, data: uN[4]:15, mark: Mark::NONE},
+    PlainData{is_mark: false, data: uN[4]:14, mark: Mark::NONE},
+    PlainData{is_mark: false, data: uN[4]:8, mark: Mark::NONE},
+    PlainData{is_mark: true, data: uN[4]:0, mark: Mark::END}
+];
+
+#[test_proc]
+proc test_simple {
```
hongted, line 586: Suggest adding a couple of more tests to target edge conditions
1. "overlap match"
2. History buffer wraparouond
3. offset zero

xls/modules/dbe/lz4_encoder.x
```diff
@@ -0,0 +1,832 @@
...
+        pold: uN[PTR_WIDTH]:0,
+        match_nconts: u32:0,
+        ht_ptr: uN[HASH_BITS]:0,
+        recycle: false,
+        finalize: false,
+        fsm: FsmSt::RESET
+    }
+}
+
+pub proc encoder_base<
```
hongted, line 151: Add a comment on how matches are found using the hash table (i.e. algorithm).
Comment on how aliases in hashing are found and prevented from causing a bad encoding.

xls/modules/dbe/lz4_encoder.x
```diff
@@ -0,0 +1,832 @@
...
+            upd
+        };
+
+        // State change logic
+        let fsm = cur.fsm;
+        let fsm = if rx_error || rx_unexpected_mark {
+            FsmSt::ERROR
+        } else if rx_reset {
+            FsmSt::RESET
+        } else if fsm == FsmSt::RESET {
```
cdleary, line 520: Nit: usually we prefer a `match` statement for FSMs in the DSL (so you can demonstrate that you exhaustively handled all cases in a declarative fashion).

xls/modules/dbe/BUILD
```diff
@@ -0,0 +1,156 @@
...
+    dslx_top = "decoder",
+    library = "dbe_lz4_decoder_dslx",
+    ir_file = "dbe_lz4_decoder_ir.ir",
+)
+
+xls_ir_opt_ir(
+    name = "dbe_lz4_decoder_opt_ir",
+    src = "dbe_lz4_decoder_ir.ir",
+    top = "__lz4_decoder__decoder__decoder_base_0__16_16_8_next",
+)
```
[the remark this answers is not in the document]
rdob-ant: @hongted Hm, I'm a bit confused here. I see that code that targets XLScc (which I haven't looked into at all till today) indeed represents RAM as an abstract object with array-like semantics, which has to be then rewritten to map onto a real-world 1RW or 1R1W RAM core.

Contrary to that, my DSLX code, as well as existing DSLX example code in e.g. _xls/examples/delay.x_ is written with a specific RAM core in mind from the beginning - there is a protocol for 1R1W core (4 channels: 2x output for rd & wr requests, 2x input for rd & wr responses) and a protocol for 1RW core (2 channels: 1x output for combined rd or wr request, 1x input for combined rd or wr response). 

If there a way to write DSLX code to target 'abstract' RAM and then rewrite it at opt_ir stage like in case of XLScc, do you know if there's example code that shows this, or at least some file with DSLX struct definitions for the abstract protocol?

xls/modules/dbe/BUILD
```diff
@@ -0,0 +1,156 @@
...
+    src = "dbe_lz4_decoder_opt_ir.opt.ir",
+    verilog_file = "dbe_lz4_decoder.v",
+    codegen_args = {
+        "module_name": "dbe_lz4_decoder",
+        "delay_model": "unit",
+        "pipeline_stages": "3",
+        "reset": "rst",
+        "use_system_verilog": "false",
+        "streaming_channel_data_suffix": "_data",
+        "io_constraints": ",".join([
```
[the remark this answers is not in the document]
rdob-ant: From what I understood, I'll first have to modify DSLX proc to use the abstract RAM protocol (see my reply above). Please let me know if got it wrong.

xls/modules/dbe/lz4_decoder.x
rdob-ant: In general it's an equivalent to a complete state reset: it both prepares the proc to handle a new block of data, as well as it clears a sticky error state that the proc (and other procs in the chain) may be in.

xls/modules/dbe/lz4_encoder.x
rdob-ant: Fixed.

xls/modules/dbe/lz4_encoder.x
rdob-ant: I'm going to prepare a document that explains how the algorithm works - this will be explained there.

xls/modules/dbe/common.x
rdob-ant: Fixed.

xls/modules/dbe/common.x
```diff
@@ -0,0 +1,75 @@
...
+    ERROR_BAD_MARK = 8,
+    ERROR_INVAL_CP = 9,
+}
+
+pub fn is_error(mark: Mark) -> bool {
+    (mark as u32) >= (Mark::_ERROR_FIRST as u32)
+}
+
+pub enum TokenKind : u2 {
+    LT = 0, // Literal
```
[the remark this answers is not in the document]
rdob-ant: Fixed.

xls/examples/ram.x
rdob-ant: Fixed in #1078

xls/modules/dbe/BUILD
hongted: Personally I'd factor out the tests into their own dslx file -- that keeps the interface cleaner.

xls/modules/dbe/BUILD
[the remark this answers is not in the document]
hongted: While the example is XLSCC, both XLSCC and DSLX are translated to a common IR. 

The idea here is to use to target an an abstract ram model so that we can easily retarget the implementation to whatever ram is available.  Whether it's 1RW, 1R1W, etc...

xls/modules/dbe/BUILD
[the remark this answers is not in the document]
hongted: Correct -- it should be a minimal change.

xls/modules/dbe/common.x
```diff
@@ -0,0 +1,75 @@
...
+    CP = 1, // Copy pointer
+    MK = 2, // Control marker
+}
+
+pub struct Token<SYM_WIDTH: u32, PTR_WIDTH: u32, CNT_WIDTH: u32> {
```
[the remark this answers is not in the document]
hongted: Understood.  Add a TODO if this the plan is to address the comment in a follow up change.

However, I think the current design should be such that it's possible to implement it in the future with minimal changes.  One thing I don't see here is how to handle the variable literal lengths available in LZ4.

What is the plan to address that?

xls/modules/dbe/lz4_decoder.x
hongted: I'd leave this to @proppy to decide whether or not the DSLX tests should be more comprehensive.

xls/modules/dbe/lz4_decoder.x
hongted: I think a TODO is warranted here then.  I think most RAMs we'll have available for implementation would have a data bus of greater than 8 bits.

xls/modules/dbe/lz4_decoder.x
```diff
@@ -0,0 +1,618 @@
...
+    // Whether we're in the ERROR state
+    err: bool,
+    // Helps to track which cells in HB are written and which ones are not
+    // (if some CP token points to an unwritten cell, we detect that and
+    // generate an error, as otherwise that will make decoder output garbage
+    // data that may leak some stale data from previous blocks)
+    hb_all_valid: bool,
+}
+
+pub proc decoder_base<SYM_WIDTH: u32, PTR_WIDTH: u32, CNT_WIDTH: u32> {
```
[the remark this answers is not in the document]
hongted: See comment above -- if abstract channels are used then RAM architecture comes free.

However, #2 remains.  Would it make sense to break apart the proc into separate communicating procs using the same SRAM?

xls/modules/dbe/lz4_decoder.x
hongted: It seems the common case is to handle the new block of data, no?  I'd suggest naming it that rather than the less common case of resetting the error state.

proppy:
> xls/examples/ram.x

Looks like there is a conflict, can we rebase?

xls/modules/dbe/lz4_encoder.x
proppy: @rdob-ant curious which format do you plan to use? (separate markdown file, comments in the header of the impl, or google docs?)

xls/modules/dbe/docs/README.md
```diff
@@ -0,0 +1,289 @@
+# Dictionary coder implementation in DSLX
```
hongted, line 1: Couple of generic comments
1. Given the long FSM backedge -- what's the expected throughput of this implementation?
2. Is the throughput/area tunable?
3. Perhaps a multi-proc implementation would be preferred to improve this.

xls/modules/dbe/docs/README.md
```diff
@@ -0,0 +1,289 @@
...
+    data: uN[DATA_WIDTH],
+    mark: Mark,
+}
+```
+
+* *is_marker* tells whether this object is a symbol or a marker.
+* *data* communicates a symbol whenever *is_marker* is not set.
+* *mark* communicates a control mark whenever *is_marker* is set.
+
+#### Token
```
hongted, line 179: I'd suggest writing another section on how these tokens can be converted into a stream compatible with the LZ4 block format (https://github.com/lz4/lz4/blob/dev/doc/lz4_Block_format.md)

Notably the following constraints
  1. That there is a minimum match length of 4 bytes
  2. That after each literal section there MUST be a minimum match
  3. The end of the sequence must have 5 literal bytes
  4. The last match must have an offset >= 12

xls/modules/dbe/docs/README.md
```diff
@@ -0,0 +1,289 @@
...
+  * Otherwise:
+    * __(3)__ Push it to the HB (dropping the oldest symbol from the HB).
+    * __(4)__ Push it to the FIFO (dropping the oldest symbol from the FIFO).
+* __(5)__ Designate the current oldest symbol in the FIFO as *current_symbol*,
+  calculate *current_ptr* - the location of this symbol in the HB.
+* __(6)__ If we do not have an existing matching string that we're trying to
+  grow, try to start a new matching string:
+  * __(7)__ Compute hash of the data contained in the FIFO.
+  * __(8)__ Load *candidate_ptr* pointer from the HT.
+  * __(9)__ Store *current_ptr* to the HT.
```
hongted, line 88: Seems the hash table is only updated when starting a new match, so that even assuming no collisions -- not all sequences will be located in the hash table?  Is this intended and matching the reference implementation?

xls/modules/dbe/BUILD
rdob-ant: Will do in the upcoming update

xls/modules/dbe/BUILD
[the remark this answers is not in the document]
rdob-ant: Forgot to reply: this was addressed, now there's a configuration for IR optimizer to rewrite the RAM interface.

xls/modules/dbe/common.x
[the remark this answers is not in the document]
rdob-ant: As for variable literal lengths in LZ4 block format.

A single Token represents either a single literal (symbol), or a single match (copy) operation. If I understand it right, the "variable literal length" that you're referring to is a way to specify the number of literals present in LZ4 sequence, where a single LZ4 sequence is defined as N literals followed by a single match operation (except for the last sequence in the block, which has only literals and no match operation). Important detail of the format is that you have to first specify the number of literals that the sequence contains (using a variable number of bytes that depends on how many literals do you have), and only then specify the literals themselves. The extreme case is of an LZ4 block that contains only literals, without any match operation - so the number of literals in the sequence approx. corresponds to the block size, which iirc can be as large as 4MB.

W.r.t. hardware implementation, LZ4 algorithm itself can be laid out rather nicely as it's stream-oriented and doesn't need any extra buffering besides 64KB history buffer and 8~32KB hash table.

However, if one is to implement an IP core that emits data in a standard LZ4 block format, due to the fact that LZ4 compressor emits tokens sequentially and their type and count is not known beforehand, and the fact that the LZ4 block specifies number of literals before the literals themselves, they'll first need to be buffered (with buffer size limited by the max block size), and only then one will be able to emit an LZ4 sequence header and then emit literals themselves. It all can be done within a single buffer as well, but you'll still need a block buffer :)

There are alternative encoding formats that try to mitigate this issue e.g. [MLZ4](https://pureadmin.qub.ac.uk/ws/files/147101774/Dec._5_Data_Compression_Device_based_on_Modified_LZ4_Algorithm_2_.pdf), but they are not compatible with existing LZ4 software.

In the scope of this PR, the idea was to implement an LZ4 encoder core that will accept a stream of raw bytes on the input, and output a stream of tokens: literals or matches, without targeting a specific block encoding format. Packing of tokens into blocks (using LZ4 or MLZ4 or any other format) is thus offloaded to an external proc, which is out of scope of this PR. However, the encoder core should of course emit enough data to allow such packing - and that's why I've recently added e.g. an "UNMATCHED_SYMBOL" token to allow easily rewriting matches (needed to address some "special" LZ4 requirements w.r.t. minimal match length and minimal literal sequence length at the end of the block).

xls/modules/dbe/lz4_decoder.x
rdob-ant: I can add a TODO. But do we want to make DSLX code aware of the port size of a specific RAM IP used in the hardware?  E.g. in FPGAs you'd frequently rely on RAM block being auto-inferred from the Verilog code, thus the synthesis tool takes care of a mismatch between the data/address bus widths of available hard-IP core and the expectations set by the RTL design.

xls/modules/dbe/lz4_decoder.x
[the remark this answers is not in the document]
rdob-ant: Hm, even if we're just considering splitting the current monolithic LZ4 compressor into multiple procs, having a few procs communicating with the same RAM may make things easier. But is this already supported in DSLX? Can you perhaps point me to such an example? I thought that a single RAM can be used at most by two procs (one reading, one writing) 🤔

xls/modules/dbe/lz4_decoder.x
rdob-ant: For now I've kept the "reset" name, as the idea of "RESET" token was to reset a full network of processing blocks (e.g. LZ4 block encoder or anything else that can be there besides this single core), and it just happened that in the encoder this function coincides with the operation that has to be done when you want to start processing of a new independent block. I can rename it to "restart" or something like that, or even add a new marker code, but that would be redundant (in a sense that if you're starting to encode an independent block of data, you don't want to preserve any previous state, so why not just completely reset the circuit?).

xls/modules/dbe/lz4_encoder.x
rdob-ant: @proppy I wrote a plain markdown file.

xls/modules/dbe/docs/README.md
rdob-ant: 1. As this is an example code, we haven't really considered the throughput while writing this code. However, as far as I remember, when I was simulating the generated Verilog code, I got something in the order of 10~12 clock cycles for a single byte (with 1-cycle single-port RAMs). Even a dual-port RAM will make it somewhat better, the number is still not good enough for a practical implementation that should be targeting something like 1 byte per cycle on average (at most).
2. Not directly. We can tune the area/compression ratio by changing the hash table size, and also using single-port -vs- double-port RAM or RAMs with different latency will result in different throughput/area ratios, but there is no parameter to control this directly in a smooth way (and tbh I can't think of any way this can be done in a scalar LZ4).
3. Probably yes. I don't see how it can be improved without splitting the design.

xls/modules/dbe/docs/README.md
rdob-ant: I will do this. In short, the idea is that you can use data conveyed within "MATCHED_SYMBOLs" to rewrite "MATCHes", e.g. if your encoding format requires a match of at least 4 symbols, and the encoder emits a match of 2 symbols, there will be two MATCHED_SYMBOL tokens accompanying a MATCH, so the postprocessor can remove the match and convert MATCHED_SYMBOLs into UNMATCHES_SYMBOLs (i.e. literals).

This way it is possible to remove or shorten matches, and I think it should be enough to cover all the requirements imposed by LZ4 standard.

xls/modules/dbe/docs/README.md
rdob-ant: I'll check this, but IIRC this behavior matches the reference code 🤔

xls/modules/dbe/common.x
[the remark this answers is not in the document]
hongted: I think my concern here is that it is not possible to output the Lz4 block encoding from this token stream.  As you mention, an implementation would need store the entire token stream, and then process that.

Sounds like what you are saying is that all implementations have that limit?  If so it sounds like LZ4 isn't a suitable compression algorithm for streaming hardware acceleration.  

@proppy -- thoughts?

xls/modules/dbe/lz4_decoder.x
hongted: I think it should be an input parameter.  SRAMS take a non-trivial amount of time to generate collateral and incorporate during an ASIC desgin flow.

xls/modules/dbe/lz4_decoder.x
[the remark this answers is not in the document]
hongted: Yes -- I don't think we've contemplated more than two procs for an SRAM.  But it's possible to write a SRAM-interface proc that will handle the arbitration.

xls/modules/dbe/lz4_decoder.x
hongted: I'd say the issue here is that the reset signal is usually special -- and may be considered a multi-cycle signal.

It sounds like this signal here is expected to be a mission-mode signal that should be timed as such -- so it's more a personal preference?

xls/modules/dbe/lz4_decoder.x
hongted: Ack.

xls/modules/dbe/docs/README.md
hongted: Thanks -- I think the discussion on token format is relevant here as well.  Is it even possible to do such a thing without needing to buffer the entire block?

xls/modules/dbe/common.x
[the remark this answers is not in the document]
rdob-ant: The algorithm itself is stream-oriented and plays rather well with the HW, but the encoding of tokens into byte sequences does not. You can check e.g. the aforementioned MLZ4 in part of encoding - they've modified it a bit so that it's more hardware-friendly and does not require excessive buffering. They've also improved the algorithm itself a bit, but that is not significant in this context.

xls/modules/dbe/BUILD
rdob-ant: I've factored out the tests & added a round-trip test.

xls/modules/dbe/docs/README.md
rdob-ant: I've added a description to the doc that shows how matches can be rewritten to fulfill these requirements.

One note, though, about requirement 2: *That after each literal section there MUST be a minimum match* - a match must be there, except if it's the end of the block. The end of the block must contain only literals and no match. Thus, a typical LZ4 block looks like this:
```
literals - match - literals - match - literals - END
```

One other requirement is about the minimal match size, and that's easy to implement with only 4 (or even 3) bytes of FIFO memory. The rest of requirements discuss how exactly that last sequence of "literals" has to look like. Those requirements can be fulfilled by making sure that there is no match within the last 12 symbols in the block - which is as simple as "buffering up to 12 symbols and if there are any matches in them, replacing those matches with literals". The tricky part is if the block ends in a long match that produces more than those 12 symbols. In this case, the match has to be split - and it's possible to do this without needing more than a 12-symbol buffer (or, how the doc says, 15-token buffer, since in reality we also want to buffer match tokens, there can be up to 3 of those per 12 symbols).

xls/modules/dbe/docs/README.md
rdob-ant: @hongted See this line in Rust Lz4 implementation: https://github.com/PSeitz/lz4_flex/blob/main/src/block/compress.rs#L488

The code above this line looks for the beginning of the match and uses hash table to do that (lines 423~425). When the initial matching sequence of 4 bytes has been found, however, it tries to grow it (line 488) but this time without updating the hash table.

proppy:
Should we close this in favor of the work on #1038 ?

proppy:
@tmichalak closing this now that the more relevant #1315 has been merged.

PR #1093, opened 2023-08-07
Grab-bag of fixes for OS X

grebe:


meheff (approved):

xls/common/subprocess_test.cc
```diff
@@ -42,7 +42,7 @@ TEST(SubprocessTest, EmptyArgvFails) {
 
 TEST(SubprocessTest, NonZeroExitWorks) {
   auto result = InvokeSubprocess(
-      {"/bin/sh", "-c", "echo -n hey && echo -n hello >&2 && exit 10"},
+      {"/bin/bash", "-c", "echo -n hey && echo -n hello >&2 && exit 10"},
```
hzeller, line 45: While `/bin/sh` is pretty universally installed on Unix-like systems, `/bin/bash` is not (and so this test fails now on minimal Unix systems or where `bash` is in a different path).
What were the things that didn't work in OSX, could we re-work it to work with the minimal bourne shell available there ?

grebe: The flags passed to echo weren't supported by `/bin/sh`

Issue #1112, opened 2023-08-23
Move XLS_EXPECT_OK_AND_EQ to xls/common


lpawelcz:
[indexer_test.cc](https://github.com/google/xls/blob/9315c506b14c672e45f0abf774614c66e7ae5099/xls/noc/simulation/indexer_test.cc#L24-L26) defines a handy macro for checking status and comparing the result against the reference if the status is OK. It would be nice to move it to common header like for example to `xls/common/status/matchers.h` where `XLS_EXPECT_OK` and `XLX_ASSERT_OK_AND_ASSIGN` are defined.

CC @proppy

PR #1114, opened 2023-08-23
Move XLS_EXPECT_OK_AND_EQ to matchers

lpawelcz:
Fixes https://github.com/google/xls/issues/1112

This PR moves macro `XLS_EXPECT_OK_AND_EQ` which was previously defined in [xls/noc/simulation/indexer_test.cc](https://github.com/antmicro/xls/blob/main/xls/noc/simulation/indexer_test.cc) to [xls/common/status/matchers.h](https://github.com/antmicro/xls/blob/main/xls/common/status/matchers.h) so that it is now widely available.

CC @proppy


proppy:
also add @hongted since he authored the original TODO.

proppy (approved):

Issue #997, opened 2023-05-30
add RISC-V based simulation of a DSLX examples


proppy:
It would be nice to show interfacing of a RISC-V simulator with the XLS JIT to demonstrate fast simulation from firmware of a DSLX examples (like #994 or #995).

tmichalak:
The idea is that the RISC-V core and SoC will be emulated (with Renode) and special C++ wrappers will be used to run DSLX and IR. To sum up in more detail:
* Renode will be a pre built portable package
* DSLX will use bytecode interpreter.
* IR will use either interpreter or JIT based on user input.
* The user specifies a path to the tested file and libraries (if needed).
* The user provides channel to address mapping using command args. This mapping will be offset based for better portability. Proto file will be used for configuration to keep configuration files type common throughout XLS. Proto will contain mapping for all channels and one extra register which will contain state of input/output FIFOs, it will provide information if FIFO is not empty.
* For channels wider than byte little endian will be used. Writing to oldest byte will trigger send transaction, reading from youngest will trigger recv transaction.

proppy:
> The user provides channel to address mapping using command args. This mapping will be offset based for better portability. Proto file will be used for configuration to keep configuration files type common throughout XLS. Proto will contain mapping for all channels and one extra register which will contain state of input/output FIFOs, it will provide information if FIFO is not empty.
> For channels wider than byte little endian will be used. Writing to oldest byte will trigger send transaction, reading from youngest will trigger recv transaction.

I wonder if we could reuse an existing protocol adapter (ex: AXIS stream) rather than creating a XLS-specific IPC protocol. As discussed with @kgugala offline renode could dynamically load the XLS JIT dll and dispatch packet directly from renode's DMA to an AXIS stream/XLS IR JIT adapter.

wdyt?

mtdudek:
We want to use an existing Renode socket peripheral.
It has a well defined protocol for reads/writes, logging and DMA operations.
This protocol is common for socket and DLL peripherals.
 
Socket based approach allows for easier reusability,
as config and design files will be the program's parameters.
 
DLL/so approach will require template C++ code.
This code would have to be recompiled for every design tested with Renode.
 
Interfacing with XLS internals, DSLX interpreter and IR interpreter/JIT, can't be avoided. 
Currently Renode only supports co-simulation with memory mapped devices.
Configuration file is necessary, as we have to map from memory mapped device (Renode) to XLS streams.

proppy:
[...]

PR #1123, opened 2023-08-31
Renode integration 

mtdudek:
This PR adds an integration layer between XLS and the Renode simulator.
It's accomplished by building a shared library which implements Renode's plugin interface which allows Renode to load & use 3rd-party plugins that emulate various HW peripherals.

This particular implementation provides an XlsPeripheral - a system bus peripheral with a single block of MMIO register space. It translates register accesses into calls to the XLS's ChannelQueues inside ProcRuntime,
It advances the simulation when a Tick message is received from Renode.

This way an application running on the CPU emulated by Renode can communicate with a proc running within XLS ProcRuntime.
To run, XlsPeripheral requires some runtime configuration (e.g. path to DSLX/IR proc file, IR channel names) which the user has to provide via a configuration file in .textproto format.

Current state:
- Integration can be loaded into Renode, log messages will be passed to the Renode, there is stub code to test that register accesses work.
- Exposing IR SingleValue channels as memory mapped registers.
- Exposing IR Stream channels as memory mapped registers.
- Passing context from Renode to the xlsperipheral.
- Reading and building simulation based on config file from context.
- Adds DMA engine to transfer data through IRStream without CPU interaction.
- Adds DMA and wrapper for AXIStream-like channels.

Related to #997


proppy:
Thanks for starting this.

With regarde to the following statement in the PR description:
> It translates register accesses into calls to the XLS's ChannelQueues inside ProcRuntime,
> Add DMA and wrapper for AXIStream-like channels.

Would `AXIStream` depend on the `ChannelQueues` for its implementation? or would those be two different commentation protocols? If the later can you expand on why we need both?

>  CPU emulated by Renode can communicate with a proc running within XLS ProcRuntime.

Can you describe the calling convention (in term of CPU operation)?

proppy:
> It's accomplished by building a shared library which implements Renode's plugin interface

Curious if a common interface could be generalized to reuse this infrastructure with other simulator? Would it makes sense to isolate the renode specific parts in a separate library?

proppy:
> AXIStream-like uses ChannelQueue to access transfer data to/form simulation.

I wonder if exposing only the `AXIStream-like` protocol could simplify and generalize the interface between the host simulator and the XLS IR simulation?

We could require that block that need to be simulated have to implement AXI-Stream, and the host simulator could dispatch the signal thru a C-API to the XLS IR simulation without any knowledge of the channels.

rdob-ant:
@proppy 
>Curious if a common interface could be generalized to reuse this infrastructure with other simulator? Would it makes sense to isolate the renode specific parts in a separate library?

I think yes, although the difficulty of that depends on which simulator you have in mind and the specifics of the interface that it uses to communicate with peripheral "drivers". But in general I think it should be possible to reuse much of current logic since most of our MMIO peripheral code is rather generic. Current code will have some references to the renode-specific communication protocol here and there, but it should be possible to factor them out later on.

>I wonder if exposing only the AXIStream-like protocol could simplify and generalize the interface between the host simulator and the XLS IR simulation? We could require that block that need to be simulated have to implement AXI-Stream, and the host simulator could dispatch the signal thru a C-API to the XLS IR simulation without any knowledge of the channels.

I don't think it matters much, since in the end all our classes, both AXI and non-AXI, are just peripherals that are controlled via MMIO regs, make level-sensitive IRQ and DMA requests to the CPU system (acting as DMA bus master), and should be periodically ticked to advance their state. So as long as we can make generic versions of all those interfaces (some of them already are generic) and implement them for a certain simulator, any simulator can be integrated this way.

proppy:
> Current code will have some references to the renode-specific communication protocol here and there, but it should be possible to factor them out later on.

Good to know!

> as long as we can make generic versions of all those interfaces (some of them already are generic) and implement them for a certain simulator, any simulator can be integrated this way.

But wouldn't a hard requirement on the streaming protocol implementation make the middle interfacing layer simpler? It could only deal with an high level procol without having to deal with all of the XLS channel semantics. Or are you saying that the implementation XLS channel and AXI stream communication primtive are similar enough in complexity not to create additional burden?

dependency_support/load_external.bzl
```diff
@@ -294,3 +294,18 @@ def load_external_repositories():
             "https://github.com/nlohmann/json/archive/refs/tags/v3.10.2.tar.gz",
         ],
     )
+
+    # 2023-06-09
+    http_archive(
+        name = "renode",
+        sha256 = "6ea7445344d56772bf527e935383ed90d8a3e5e8bef8649acb1b56e8edcba234",
+        strip_prefix = "renode_1.14.0+20230829git57762908_portable",
+        urls = ["https://builds.renode.io/renode-1.14.0+20230829git57762908.linux-portable.tar.gz"],
```
proppy, line 303: do you already have a bazel workspace for renode? it would be nice to be able to compile from source.

mtdudek:
I think this PR was to large to be done as a single unit.
So it is now split into 4 PRs:

- [ ] #1176 
- [ ] #1177 
- [ ] #1178 

and this one.

New PRs add generic code that can be used by other integrations,
while this PR should only add Renode specific classes.

proppy:
should we close this and favor of reviewing the demonstrator at https://github.com/antmicro/xls-cosimulation-demonstrator ?

we can then re-open a new PR once we figure out how to integrate the simulator build w/ the bazel workspace?

mtdudek:
Closing this PR in the favour of the demonstrator repo.

PR #1166, opened 2023-10-26
modules/zstd: build libzstd

rdob-ant:
This PR adds zstd library to the XLS framework. It exposes the following targets:

    @com_github_facebok_zstd//:libzstd - zstd library itself
    @com_github_facebok_zstd//:zstd - binary for performing zstd compression/decompression
    @com_github_facebok_zstd//:decodecorpus - binary for creating test data for a zstd decoder

Additionally, we remove zstd from llvm build to avoid linking conflicts with newly-added zstd library. It seems that llvm_zstd is not necessary for correct llvm operation.

### DIFF
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -42,12 +42,8 @@ cc_library(
         "lib/zstd.h",
     ],
     strip_include_prefix = "lib",
-    linkopts = [
-        "-pthread",
-    ],
     local_defines = [
         "ZSTD_LEGACY_SUPPORT=4",
-        "ZSTD_MULTITHREAD",
         "XXH_NAMESPACE=ZSTD_",
     ],
     visibility = ["//visibility:public"],
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -1,0 +1,54 @@
+# Copyright 2023 The XLS Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


proppy:
> @com_github_facebok_zstd//:zstd - binary for performing zstd compression/decompression
> @com_github_facebok_zstd//:decodecorpus - binary for creating test data for a zstd decoder

Do we need to those? It sounds like we would motsly need to drive (software)-zstd from C++ tests to compare data with DSLX/IR simulation (and not from the shell or bazel rules).

rw1nkler:
They were added for our convenience during development but we can remove those two targets from the build file

proppy (changes requested):
can you squash the commits, and rename the PR (without the part 1)

proppy:
also if you rebase, that should get rid of the macosx continuous integration failure (it only runs nightly now)

dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
```diff
@@ -0,0 +1,54 @@
...
+        "lib/dictBuilder/*.c",
+        "lib/legacy/*.h",
+        "lib/legacy/*.c",
+    ]),
+    hdrs = [
+        "lib/zstd.h",
+    ],
+    strip_include_prefix = "lib",
+    linkopts = [
+        "-pthread",
```
grebe, line 46: Is this necessary? I'd have thought not

dependency_support/load_external.bzl
```diff
@@ -286,3 +286,11 @@ def load_external_repositories():
             "https://github.com/nlohmann/json/archive/refs/tags/v3.10.2.tar.gz",
         ],
     )
+
+    http_archive(
```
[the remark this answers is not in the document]
lpawelcz: Added comment with release info and dependency update date

dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
lpawelcz: It looks like disabling multithread builds does not affect our tests which use zstd library. I've updated the build rule to build library without multithread support.

PR #1168, opened 2023-10-26
modules/zstd: add frame header parser

rdob-ant:
This PR adds Zstd frame header parsing machinery written in DSLX + tests for it.

NOTE: this is based on #1167 , please ignore commits from that branch when reviewing.

### DIFF
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -49,3 +49,42 @@ cc_library(
     ],
     visibility = ["//visibility:public"],
 )
+
+# NOTE: Required because of direct zstd_compress.c include in decodecorpus sources
+cc_library(
+    name = "decodecorpus_includes",
+    hdrs = [
+        "lib/compress/zstd_compress.c",
+    ],
+)
+
+cc_binary(
+    name = "decodecorpus",
+    srcs = [
+        "tests/decodecorpus.c",
+    ] + glob(
+        [
+            "programs/*.c",
+            "programs/*.h",
+        ],
+        exclude = [
+            "programs/zstdcli.c",
+        ],
+    ),
+    deps = [
+        ":zstd",
+        ":decodecorpus_includes",
+    ],
+    includes = [
+        "lib/",
+        "lib/common/",
+        "lib/compress/",
+        "lib/dictBuilder/",
+        "lib/deprecated/",
+        "programs/",
+    ],
+    local_defines = [
+        "XXH_NAMESPACE=ZSTD_",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -40,6 +40,7 @@ cc_library(
     ]),
     hdrs = [
         "lib/zstd.h",
+        "lib/common/zstd_errors.h",
     ],
     strip_include_prefix = "lib",
     local_defines = [
--- a/dependency_support/load_external.bzl
+++ b/dependency_support/load_external.bzl
@@ -297,4 +297,6 @@ def load_external_repositories():
         strip_prefix = "zstd-1.4.7",
         urls = ["https://github.com/facebook/zstd/releases/download/v1.4.7/zstd-1.4.7.tar.gz"],
         build_file = "@//dependency_support/com_github_facebook_zstd:bundled.BUILD.bazel",
+        # Modify decodecorpus to allow generation of zstd frame headers only
+        patches = ["@com_google_xls//dependency_support/com_github_facebook_zstd:decodecorpus.patch"],
     )
--- a/xls/modules/zstd/data_generator.cc
+++ b/xls/modules/zstd/data_generator.cc
@@ -1,0 +1,110 @@
+// Copyright 2023 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "xls/modules/zstd/data_generator.h"
+
+#include <algorithm>
+#include <filesystem>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <iterator>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "absl/time/time.h"
+#include "absl/types/span.h"
+#include "xls/common/file/filesystem.h"
+#include "xls/common/file/get_runfile_path.h"
+#include "xls/common/status/status_macros.h"
+#include "xls/common/subprocess.h"
+#include "xls/ir/value.h"
+
+namespace xls::zstd {
+
+static void PrintRawDataVector(std::vector<uint8_t> vector) {
+  for (int i = 0; i < vector.size(); ++i) {
+    std::cout << std::setfill('0') << std::setw(8) << std::hex << i << ": "
+              << "0x" << std::setw(2) << std::hex << int(vector[i])
+              << std::endl;
+  }
+}
+
+static absl::StatusOr<std::vector<uint8_t>> ReadFileAsRawData(
+    const std::filesystem::path& path) {
+  std::ifstream file(path, std::ios::binary);
+  if (!file.is_open()) {
+    return absl::NotFoundError("Unable to open a test file");
+  }
+
+  std::vector<uint8_t> raw_data((std::istreambuf_iterator<char>(file)),
+                                (std::istreambuf_iterator<char>()));
+  return raw_data;
+}
+
+static std::string CreateNameForGeneratedFile(
+    absl::Span<std::string> args, std::string_view ext,
+    std::optional<std::string_view> prefix) {
+  std::string output;
+
+  if (prefix.has_value()) {
+    output += prefix.value();
+    output += "_";
+  }
+
+  for (auto const& x : args) {
+    output += x;
+  }
+  auto nospace_output = std::remove(output.begin(), output.end(), ' ');
+  output.erase(nospace_output, output.end());
+  std::replace(output.begin(), output.end(), '-', '_');
+
+  output += ext;
+
+  return output;
+}
+
+static absl::StatusOr<SubprocessResult> CallDecodecorpus(
+    absl::Span<const std::string> args,
+    std::optional<std::filesystem::path> cwd = std::nullopt,
+    std::optional<absl::Duration> timeout = std::nullopt) {
+  XLS_ASSIGN_OR_RETURN(
+      std::filesystem::path path,
+      xls::GetXlsRunfilePath("external/com_github_facebook_zstd/decodecorpus"));
+
+  std::vector<std::string> cmd = {path};
+  cmd.insert(cmd.end(), args.begin(), args.end());
+  return SubprocessErrorAsStatus(xls::InvokeSubprocess(cmd));
+}
+
+absl::StatusOr<std::vector<uint8_t>> GenerateFrameHeader(int seed, bool magic) {
+  std::array<std::string, 4> args;
+  args[0] = "-s" + std::to_string(seed);
+  args[1] = (magic) ? "" : "--no-magic";
+  args[2] = "--frame-header-only";
+  std::filesystem::path output_path =
+      std::filesystem::temp_directory_path() /
+      std::filesystem::path(
+          CreateNameForGeneratedFile(absl::MakeSpan(args), ".zstd", "fh"));
+  args[3] = "-p" + std::string(output_path);
+
+  XLS_ASSIGN_OR_RETURN(auto result, CallDecodecorpus(args));
+  auto raw_data = ReadFileAsRawData(output_path);
+  std::remove(output_path.c_str());
+  return raw_data;
+}
+
+}  // namespace xls::zstd
--- a/dependency_support/com_github_facebook_zstd/decodecorpus.patch
+++ b/dependency_support/com_github_facebook_zstd/decodecorpus.patch
@@ -1,0 +1,152 @@
+diff --git tests/decodecorpus.c tests/decodecorpus.c
+index 50935d31..522b3769 100644
+--- tests/decodecorpus.c
++++ tests/decodecorpus.c
+@@ -240,6 +240,12 @@ typedef enum {
+   gt_block,      /* generate compressed blocks without block/frame headers */
+ } genType_e;
+
++typedef enum {
++  lt_raw,
++  lt_rle,
++  lt_compressed,
++} literalType_e;
++
+ /*-*******************************************************
+ *  Global variables (set from command line)
+ *********************************************************/
+@@ -252,7 +258,11 @@ U32 g_maxBlockSize = ZSTD_BLOCKSIZE_MAX;                       /* <= 128 KB */
+
+ struct {
+     int contentSize; /* force the content size to be present */
+-} opts; /* advanced options on generation */
++    blockType_e *blockType; /* force specific block type */
++    literalType_e *literalType; /* force specific literals type */
++    int frame_header_only; /* generate only frame header */
++    int no_magic; /* do not generate magic number */
++} opts;
+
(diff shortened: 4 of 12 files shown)


xls/modules/zstd/zstd_frame_header.x
```diff
@@ -0,0 +1,648 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains utilities related to ZSTD Frame Header parsing.
+// More information about the ZSTD Frame Header can be found in:
+// https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.1
```
proppy, line 17: thanks for linking to the spec!

xls/modules/zstd/zstd_frame_header.x
```diff
@@ -0,0 +1,648 @@
...
+        BufferStatus::OK => {
+            let frame_header_desc = extract_frame_header_descriptor(data);
+            (result, frame_header_desc)
+        },
+        _ => (result, zero!<FrameHeaderDescriptor>())
+    }
+}
+
+#[test]
+fn test_parse_frame_header_descriptor() {
```
proppy, line 123: should we add c++ tests that create/parse header with the zstd c++ library and assert them against the dslx/ir simulation?

rw1nkler: Sure! I will provide those tests

lpawelcz:
@proppy we updated the code, please take a look:
* rebased onto newest changes from https://github.com/google/xls/pull/1167
* picked changes from https://github.com/google/xls/pull/1166
* Updated FrameHeader parser to be compatible with changes from https://github.com/google/xls/pull/1167
* Patched `decodecorpus` from `@com_github_facebook_zstd` to allow generating random FrameHeaders
* Added `data_generator` library for generating random zstd-encoded data with `decodecorpus` (generating FrameHeaders)
* Added FrameHeader parser tests

C++ tests generate 250 test cases with pseudorandom valid FrameHeaders. Test cases run in parallel through `sharding` mechanism.
In a single tests case a frame header in form of a vector of `uint8_t` is generated. It is then passed through `libzstd` to convert it into `ZSTD_frameHeader` struct and also to validate the data. After that, generated vector is converted to DSLX simulation input values and `ZSTD_frameHeader` structure is used to form expected return values to compare against the results of DSLX simulation.

Please do note that this branch is based on https://github.com/google/xls/pull/1167 and it also includes commits cherry-picked from https://github.com/google/xls/pull/1166. Please ignore those when reviewing this PR.

dependency_support/load_external.bzl
```diff
@@ -286,3 +286,15 @@ def load_external_repositories():
...
+    # Version 1.4.7 released on 17.12.2020
+    # https://github.com/facebook/zstd/releases/tag/v1.4.7
+    # Updated 23.11.2023
+    http_archive(
+        name = "com_github_facebook_zstd",
+        sha256 = "192cbb1274a9672cbcceaf47b5c4e9e59691ca60a357f1d4a8b2dfa2c365d757",
+        strip_prefix = "zstd-1.4.7",
+        urls = ["https://github.com/facebook/zstd/releases/download/v1.4.7/zstd-1.4.7.tar.gz"],
+        build_file = "@//dependency_support/com_github_facebook_zstd:bundled.BUILD.bazel",
+        patches = ["@com_google_xls//dependency_support/com_github_facebook_zstd:decodecorpus.patch"],
```
proppy, line 299: can you add a comment about the patch? has it been submitted upstream?

dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
```diff
@@ -0,0 +1,89 @@
...
+
+# NOTE: Required because of direct zstd_compress.c include in decodecorpus sources
+cc_library(
+    name = "decodecorpus_includes",
+    hdrs = [
+        "lib/compress/zstd_compress.c",
+    ],
+)
+
+cc_binary(
```
proppy, line 60: do we actually need the binary rule?

xls/modules/zstd/data_generator.cc
```diff
@@ -0,0 +1,110 @@
...
+    absl::Span<const std::string> args,
+    std::optional<std::filesystem::path> cwd = std::nullopt,
+    std::optional<absl::Duration> timeout = std::nullopt) {
+  XLS_ASSIGN_OR_RETURN(
+      std::filesystem::path path,
+      xls::GetXlsRunfilePath("external/com_github_facebook_zstd/decodecorpus"));
+
+  std::vector<std::string> cmd = {path};
+  cmd.insert(cmd.end(), args.begin(), args.end());
+  return SubprocessErrorAsStatus(xls::InvokeSubprocess(cmd));
```
proppy, line 90: any reason we shell out rather than calling into libstd natively?

xls/modules/zstd/data_generator.cc
```diff
@@ -0,0 +1,110 @@
...
+  args[1] = (magic) ? "" : "--no-magic";
+  args[2] = "--frame-header-only";
+  std::filesystem::path output_path =
+      std::filesystem::temp_directory_path() /
+      std::filesystem::path(
+          CreateNameForGeneratedFile(absl::MakeSpan(args), ".zstd", "fh"));
+  args[3] = "-p" + std::string(output_path);
+
+  XLS_ASSIGN_OR_RETURN(auto result, CallDecodecorpus(args));
+  auto raw_data = ReadFileAsRawData(output_path);
```
proppy, line 105: any reason we need to use a real file, rather than handling in memory buffer returned by libstd?

xls/modules/zstd/zstd_frame_header_test.cc
```diff
@@ -0,0 +1,263 @@
...
+class ZSTDFrameHeaderSeededTest
+    : public ZSTDFrameHeaderTest,
+      public ::testing::WithParamInterface<uint32_t> {
+ public:
+  static const uint32_t random_headers_count = 250;
+};
+
+TEST_P(ZSTDFrameHeaderSeededTest, ParseMultipleFrameHeaders) {
+  auto seed = GetParam();
+  auto frame_header = zstd::GenerateFrameHeader(seed, false);
```
proppy, line 247: is this redundant to what we're doing in `GenerateRandomFrameHeader` or does it have more coverage?

xls/modules/zstd/zstd_frame_header_test.cc
```diff
@@ -0,0 +1,263 @@
...
+    absl::flat_hash_map<std::string, Value> hashed_input = {{"buffer", input}};
+
+    auto expected_frame_header_result = CreateExpectedFrameHeaderResult(
+        &zstd_fh, input, buffer, expected_status);
+
+    RunAndExpectEq(hashed_input, expected_frame_header_result, this->converted,
+                   true, true);
+  }
+
+  std::vector<uint8_t> GenerateRandomFrameHeader() {
```
proppy, line 127: I wonder if https://github.com/google/xls/issues/476 would help doing this kind of things?

dependency_support/load_external.bzl
lpawelcz: It was not submitted upstream yet. The patch modifies `decodecorpus` utility from zstd library. Unmodified tool allows generating valid zstd frames with randomized contents and size. With our modifications it is possibile to generate only some parts of the zstd frame. In this case we modify `decodecorpus` so that it allows generating only frame headers.

dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
lpawelcz: Decodecorpus in zstd is used as a standalone binary for generating test data. We decided to use it similarly here.

xls/modules/zstd/data_generator.cc
lpawelcz: That's because decodecorpus was introduced to zstd as a standalone binary. We decided to make the smallest possible changes to this tool in order to use it in our testing procedure so we just call it in a separate subprocess.

xls/modules/zstd/data_generator.cc
lpawelcz: This was not possible in original `decodecorpus` and we wanted to provide the minimal changes required for generating only frame headers.

xls/modules/zstd/zstd_frame_header_test.cc
lpawelcz: `GenerateFrameHeader()` calls decodecorpus in order to generate frame header which is valid in case of zstd frame header specification. The contents and length of those frame headers are randomized but the header is always valid - it should always be correctly parsed both by libzstd and by our decoder (frame header parser).

`GenerateRandomFrameHeader()` generates completely random vector of bytes of randomly picked size from range 0 to arbitrary max size of 32 bytes. With this generator we are able to test more 'negative' test cases, meaning:
* Not enough data in buffer for finishing parsing
* Corrupted header - e.g. `reserved` bit in `Frame_Header_Descriptor` is set
* Handling of buffers larger than bare minimum for parsing frame header

xls/modules/zstd/zstd_frame_header_test.cc
lpawelcz: This kind of testing approach would be great if we didn't have `decodecorpus` which always generate valid zstd frames/frame headers. We could say that decodecorpus has embedded in itself the properties that describe valid frame header.
Without it we would have to manually specify those properties in form of multiple `RC_ASSERT()` statements in rapidcheck test cases in order to generate valid test data. I think that when it comes to 'positive' tests (comparing against valid frame headers) it is better to use decodecorpus.
However, I'm not sure how would it look like for 'negative' tests (generating potentially invalid frame header data and comparing parsing through libzstd against simulation of our decoder).
Frame Header parsing should fail basically only in 2 situations:
* there was not enough data to finish the parsing
* reserved bit in frame header descriptor was set

Currently we generate random test vectors in order to check if decoder fails when those situations occur.

dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
proppy: Make sense.

xls/modules/zstd/data_generator.cc
proppy: Ditto.

xls/modules/zstd/zstd_frame_header_test.cc
ericastor: I agree that using GoogleFuzzTest here would probably result in better coverage - though it might end up discovering some valid headers!

xls/modules/zstd/zstd_frame_header_test.cc
lpawelcz: Added more comments

xls/modules/zstd/zstd_frame_header_test.cc
lpawelcz: Thanks for the suggestion, I've replaced generating random frame headers with `FuzzTest`. It is true that it can discover valid headers. We can work on constraining the tested domain further to precisely test only invalid frame headers but I believe it is better to do a general check with valid and invalid headers.

xls/modules/zstd/data_generator.cc
lpawelcz: I believe this was discussed with you on one of the meetings. We settled on leaving `decodecorpus` as it is.

lpawelcz:
Fixed errors in fuzz tests caused by differences in priority of reporting errors between `libzstd` and our implementation. `libzstd` prioritizes reporting error about not enough data available in the buffer for parsing the whole frame header over error caused by reserved bit in frame header descriptor being set.
I've changed our implementation to act the same way as the library. This required changing the contents of the `Buffer` when returning with `CORRUPTED` error. We treat this error as critical one which requires resetting the whole decoder because after occurring corrupted frame header it is impossible to recover from this condition. Because of that we can safely return empty buffer in such case.

lpawelcz:
I've noticed a failure in one of the CI runs caused by `frame_header_cc_test` and specifically by the FuzzTest case:
https://github.com/google/xls/actions/runs/7410803786/job/20163885832?pr=1214

Here is the failure log:
```
2024-01-04T14:55:33.3138622Z [----------] 1 test from FrameHeaderFuzzTest
2024-01-04T14:55:33.3139507Z [ RUN      ] FrameHeaderFuzzTest.ParseMultipleRandomFrameHeaders
2024-01-04T14:55:33.3140674Z FUZZTEST_PRNG_SEED=NuIZziOUrkbi_BmDtWnqoy7SbNkrX-lwmJlXdfEMCyQ
2024-01-04T14:55:33.3141491Z xls/ir/ir_test_base.cc:264: Failure
2024-01-04T14:55:33.3142131Z Value of: ValuesEqual(expected, result.value)
2024-01-04T14:55:33.3143777Z   Actual: false ((bits[2]:1, (bits[64]:0, bits[64]:0, bits[32]:0, bits[1]:0), (bits[128]:0x0, bits[32]:0)) != (bits[2]:0, (bits[64]:94489280512, bits[64]:18446744073709551615, bits[32]:0, bits[1]:0), (bits[128]:0x0, bits[32]:0)))
2024-01-04T14:55:33.3145422Z Expected: true
2024-01-04T14:55:33.3145803Z (interpreted unoptimized IR)
2024-01-04T14:55:33.3146240Z Google Test trace:
2024-01-04T14:55:33.3146874Z xls/modules/zstd/frame_header_test.cc:137: RunAndExpectEq failed
2024-01-04T14:55:33.3147656Z Stack trace:
2024-01-04T14:55:33.3148169Z   0x7f34a8ec8185: xls::IrTestBase::RunAndExpectEq()
2024-01-04T14:55:33.3148941Z   0x7f34a8eca1db: xls::IrTestBase::RunAndExpectEq()
2024-01-04T14:55:33.3149994Z   0x563f0f26374d: xls::(anonymous namespace)::FrameHeaderTest::ParseAndCompareWithZstd()
2024-01-04T14:55:33.3151137Z   0x563f0f276f37: fuzztest::internal::FixtureDriver<>::Test()
2024-01-04T14:55:33.3152162Z   0x7f349fbaf317: fuzztest::internal::FuzzTestFuzzerImpl::RunOneInput()
2024-01-04T14:55:33.3153286Z   0x7f349fbb19de: fuzztest::internal::FuzzTestFuzzerImpl::RunInUnitTestMode()
2024-01-04T14:55:33.3154354Z   0x7f34ac0abbb3: fuzztest::internal::GTest_TestAdaptor::TestBody()
2024-01-04T14:55:33.3155403Z   0x7f349f443ffd: testing::internal::HandleExceptionsInMethodIfSupported<>()
2024-01-04T14:55:33.3156707Z   0x7f349f443ebe: testing::Test::Run()
2024-01-04T14:55:33.3157594Z   0x7f349f445171: testing::TestInfo::Run()
2024-01-04T14:55:33.3158212Z ... Google Test internal frames ...
2024-01-04T14:55:33.3158607Z 
2024-01-04T14:55:33.3158614Z 
2024-01-04T14:55:33.3158848Z =================================================================
2024-01-04T14:55:33.3159429Z === BUG FOUND!
2024-01-04T14:55:33.3159679Z 
2024-01-04T14:55:33.3160606Z xls/modules/zstd/frame_header_test.cc:298: Counterexample found for FrameHeaderFuzzTest.ParseMultipleRandomFrameHeaders.
2024-01-04T14:55:33.3161924Z The test fails with input:
2024-01-04T14:55:33.3162413Z argument 0: {16, 211}
2024-01-04T14:55:33.3162996Z 
2024-01-04T14:55:33.3163238Z =================================================================
2024-01-04T14:55:33.3164180Z [  FAILED  ] FrameHeaderFuzzTest.ParseMultipleRandomFrameHeaders (837 ms)
2024-01-04T14:55:33.3165326Z [----------] 1 test from FrameHeaderFuzzTest (837 ms total)
2024-01-04T14:55:33.3165859Z 
2024-01-04T14:55:33.3166176Z [----------] Global test environment tear-down
2024-01-04T14:55:33.3166902Z [==========] 2 tests from 2 test suites ran. (1568 ms total)
2024-01-04T14:55:33.3167559Z [  PASSED  ] 1 test.
2024-01-04T14:55:33.3168029Z [  FAILED  ] 1 test, listed below:
2024-01-04T14:55:33.3168813Z [  FAILED  ] FrameHeaderFuzzTest.ParseMultipleRandomFrameHeaders
2024-01-04T14:55:33.3169478Z 
2024-01-04T14:55:33.3169632Z  1 FAILED TEST
```

I've already investigated the issue and managed to fix it. It was caused by not discarding ZSTD frames with `window_size` being larger than specified limit. PR will soon be updated with the fix for that

proppy:
should we close this and focus on reviewing https://github.com/google/xls/pull/1315 ?

lpawelcz:
The review will take place in https://github.com/google/xls/pull/1315

@proppy this PR can be closed

PR #1169, opened 2023-10-26
modules/zstd: add Zstd decoder skeleton

rdob-ant:
This PR adds a skeleton of Zstd decoder that for now only parses a frame header (using DSLX modules added by other Zstd PRs). The module itself is in DSLX, while tests are in C++ as they need to use file IO and will also call into libzstd in the future. As this is a skeleton only, it'll most probably undergo substantial changes when other Zstd parser modules are added.

NOTE: this is based on #1168 , please ignore commits from that branch when reviewing.


proppy:
> while tests are in C++ as they need to use file IO and will also call into libzstd in the future

👍 for writing test in C++ that compare against libzstd.

xls/modules/zstd/zstd_dec.x
```diff
@@ -0,0 +1,103 @@
...
+const BUFFER_WIDTH = u32:128;
+const ZERO_FRAME_HEADER = frame_header::ZERO_FRAME_HEADER;
+
+enum ZstdDecoderStatus : u8 {
+  IDLE = 0,
+  DECODE_FRAME_HEADER = 1,
+  DECODE_BLOCKS = 2,
+}
+
+struct ZstdDecoderState {
```
proppy, line 36: can we add doc string for each member of the state?

xls/modules/zstd/zstd_dec.x
```diff
@@ -0,0 +1,103 @@
...
+                        buffer: magic_result.buffer,
+                        frame_header: ZERO_FRAME_HEADER,
+                    },
+                    _ => state
+                }
+            },
+            _ => state
+        };
+
+        let state = match state.status {
```
proppy, line 85: are you keeping the match separte so that you can process the frame header in the same tick? (if yes, is this state even necessary?)

xls/modules/zstd/zstd_dec_test.cc
```diff
@@ -0,0 +1,108 @@
...
+    }
+    this->interpreter->ClearInterpreterEvents();
+  }
+
+  const char *proc_name = "__zstd_dec__ZstdDecoder_0_next";
+  const char *recv_channel_name = "zstd_dec__output_s";
+  const char *send_channel_name = "zstd_dec__input_r";
+
+  const char *ir_file = "xls/modules/zstd/zstd_dec_opt_ir.opt.ir";
+  const char *zstd_file = "xls/modules/zstd/data.zst";
```
proppy, line 64: maybe we could generate this from the text fixture?

rw1nkler: Yes! We will add the ability to generate test data

xls/modules/zstd/zstd_dec.x
rw1nkler: With the current shape of the skeleton, the state is not necessary. However, soon we will extend it with the decoding capabilities that would make use of the internal state

proppy:
should we close this and focus on reviewing #1315 ?

rw1nkler:
Closing in favor of #1315

tmichalak:
The review will happen in https://github.com/google/xls/pull/1315,  so this one can be closed.

Issue #1174, opened 2023-10-28
Build downloads two different external python binaries


hzeller:
The build downloads two different external binary versions of Python
  * external/python39_x86_64-unknown-linux-gnu/bin/python3.9
  * external/rules_hdl_cpython/install/bin/python3.8
 
Maybe not too much of an issue in reality but probably something to be aware of. Would of course be cool if the different parts could settle one one version.

The escape hatch to use the system Python does now work anymore [`--repo_env=PYTHON_BIN_PATH`](https://github.com/google/xls/blob/main/README.md?plain=1#L112-L114), so probably worthwhile removing that documentation.

cdleary:
Is this because we need to update rules_hdl to be consistent with the XLS repo? cc @mithro who had been looking at that

PR #1175, opened 2023-10-28
Update README to not include how to use system python.

hzeller:
The build downloads external python binaries and this override does not work anymore.

Issues: #1174


ericastor:
For some reason the import is failing on this PR - possibly because I accidentally let it get this stale. My apologies.

We'll make this change internally.

PR #1176, opened 2023-10-31
Generic integration interfaces

mtdudek:
Generic integration creates memory mapped device.

This is the first PR in the series of 3.
It adds interfaces that are used in the generic integration.

PR #1177, opened 2023-10-31
Generic integration channels

mtdudek:
Generic integration creates memory mapped device.

This is the second PR  in the series of 3.
It requires #1176 

It adds channel wrapper for IR single value and IR stream channel types.
It also adds AXIStreamLike wrapper.
It is design to work with channels that have control signals similar to the AXIStream.

This PR also changes parts of the RLE implementation,
so it works with the ir_axistreamlike wrapper.


proppy:
should we close this and favor of reviewing the demonstrator at https://github.com/antmicro/xls-cosimulation-demonstrator ?

we can then re-open a new PR once we figure out how to integration gem5 w/ the bazel workspace?

PR #1183, opened 2023-11-06
Add example of DSLX design parameterization based on proto files

lpawelcz:
This example showcases the usage of [proto_to_dslx](https://github.com/google/xls/blob/main/xls/tools/proto_to_dslx.cc) tool in DSLX design parameterization. It defines a protobuf message containing design parameters:

* input width,
* output width

Two textproto files with different variants of the design are provided. Each textproto file can be converted to DSLX file which is then imported by top-level design in order to use the parameters.

The design itself has 2 inputs and one output. The widths of those are controlled by parameters.

In order to provide single import path in top DSLX file custom configuration setting 'design_variant' was provided. It is used to select the design variant from bazel command line.

Verilog files for the variants can be generated with:

```
bazel build //xls/examples/parameterized:design_verilog --//xls/examples/parameterized:design_variant=narrow
```

and:

```
bazel build //xls/examples/parameterized:design_verilog --//xls/examples/parameterized:design_variant=wide
```

@proppy In this example I wanted to explore your suggestion from https://github.com/google/xls/pull/1160#issuecomment-1781331646, please take a look.


xls/examples/parameterized/BUILD
```diff
@@ -0,0 +1,99 @@
...
+    "xls_dslx_ir",
+    "xls_dslx_library",
+    "xls_ir_opt_ir",
+    "xls_ir_verilog",
+)
+
+# First variant of the design
+proto_to_dslx(
+    name = "parameters_add",
+    proto_schema_path = ":param.proto",
```
proppy, line 14: would we gain anything by pointing to a `proto_library` here? (so that we can get hooked to a `NativeProtoInfo` provider)

xls/examples/parameterized/BUILD
```diff
@@ -0,0 +1,99 @@
...
+
+config_setting(
+    name = "design_sub",
+    flag_values = {
+        ":design_variant": "sub",
+    },
+)
+
+# Move selected variant to location matching import statement in DSLX design
+genrule(
```
proppy, line 50: curious why we need the intermediate genrules? If we emit a `DslxInfo` from the `proto_to_dslx` rule, we should be able to plug it directly into a `dslx_library`.

xls/examples/parameterized/design.x
```diff
@@ -0,0 +1,32 @@
...
+
+fn top(x: u64, y: u64) -> u64 {
+  let variant: DesignParameters = parameters::parameters.variant[0];
+  let variant_type: DesignType = variant.design_type;
+  let variant_input_width: u32 = variant.input_width;
+  let variant_output_width: u32 = variant.output_width;
+  let arg_x = x as uN[variant_input_width];
+  let arg_y = y as uN[variant_input_width];
+
+  let result = if (variant_type == DesignType::DESIGN_TYPE_ADD) {
```
proppy, line 26: maybe we should stick to paramertizing parameter of a single function (I would argue that switching between function is less common and not necessarly what `proto_to_dslx` seems designed for)

xls/build_rules/xls_utilities.bzl
```diff
@@ -306,3 +306,76 @@ Examples:
...
+              "file represents.",
+        mandatory = True,
+    ),
+    "textproto": attr.label(
+        doc = "The source textproto file to convert to DSLX",
+        allow_single_file = True,
+        mandatory = True,
+    ),
+    "variable_name": attr.string(
+        doc = "The name of the struct variable with contents of the textproto file",
```
proppy, line 363: should we have a default here? (i.e: `parameters`)

xls/build_rules/xls_utilities.bzl
```diff
@@ -306,3 +306,76 @@ Examples:
...
+        mandatory = True,
+    ),
+    "dslx_file": attr.output(
+        doc = "The name of the output file to write DSLX output to.",
```
proppy, line 367: should we document the default?

xls/build_rules/xls_utilities.bzl
```diff
@@ -306,3 +306,76 @@ Examples:
...
+        command = "{} --proto_def_path {} --proto_name {} --textproto_path {} --var_name {} --output_path {}".format(
+            proto2dslx.path,
+            schema.path,
+            ctx.attr.proto_name,
+            ctx.file.textproto.path,
+            ctx.attr.variable_name,
+            dslx_file.path
+        ),
+        use_default_shell_env = True,
+        mnemonic = "Proto2DSLX",
```
proppy, line 342: curious what the mnemonic is for?

proppy (changes requested):
thanks for adding this, left some preliminary comments.

@hongted, I don't think we have an equivalent of this yet? (afaik we mainly rely on `genrules` for this internally)

xls/examples/parameterized/BUILD
lpawelcz: Using existing bazel constructs to abstract this out looks like a good idea

xls/examples/parameterized/BUILD
lpawelcz: This is only for the purpose of allowing easier builds for two different variants of the same design. The problem is that in top-level DSLX file ([here](https://github.com/google/xls/pull/1183/files#diff-c1e7fa11e9f9c66b0c4be19423d8c4937677ed1cd56a171712cdc431a52fa612R1)) we need to provide import path to our DSLX file with parameters. In this example I wanted to provide two different variants so that it would be visible that the implemented design is indeed changing depending on which textproto file was used for the build.
The import line must point to a file with specific name. Bazel does not allow specifying two rules (`proto_to_dslx`) that would generate the same file from two different textproto configurations.
My workaround for this is to generate files with different names and then pick one based on custom build flag and `select()` and then copy the file to generic DSLX file used in the import statement in top-level DSLX

xls/examples/parameterized/design.x
lpawelcz: Here I just wanted to show a different kind of parameter. Something different than I/O width, something that would change the behaviour of the whole module. I'm not that much experienced in DSLX and that seemed like the easiest way to achieve that effect. We should probably change that parameter to something else if we want a real-life example or leave it like that in order to keep it simple and only showcase the parameterization mechanism on minimal example.

xls/build_rules/xls_utilities.bzl
lpawelcz: Certainly!

xls/build_rules/xls_utilities.bzl
lpawelcz: This is not required, it is a one-word description of the action

xls/examples/parameterized/BUILD
lpawelcz: Changed the rule implementation to accept `proto_schema` attribute which provides `ProtoInfo`

xls/build_rules/xls_utilities.bzl
lpawelcz: Added and documented default

xls/build_rules/xls_utilities.bzl
lpawelcz: Done

xls/examples/parameterized/design.x
proppy: It'd focus on a minimal example for now, dslx paremetric behavior is not yet very advanced and is not really meant to swap behavior afaik.

lpawelcz: Alright, so in order to keep this example minimal I will remove `design_type` parameter and leave only IO width.

lpawelcz:
Removed `DesignType` parameter to keep the example minimal, rebased the code and cleaned git history. @proppy please take another look.

xls/examples/parameterized/BUILD
proppy: rather than using select, I think as a minimal example it would be simpler to have two `xls_dlsx_library` variant "top" that depends on the `proto_to_dslx` targets (to show the direct connection between the two rules)

lpawelcz: I was thinking about that solution before opening the PR. The thing is that I thought that having 2 top level DSLX files defeats the purpose of having parameterized design in the first place because of doubling the code. It was important for me to show that you can use a single DSLX top level design and achieve different final results depending on which textproto was used to get design parameters.
Having said that, I don't have a strong opinion here. For the sake of simplicity we can switch to example with 2 top-level designs and without select statements.

lpawelcz: >  I think as a minimal example it would be simpler to have two xls_dlsx_library variant "top" that depends on the proto_to_dslx targets

I've reworked the example to have 2 separate top-level designs. Please take a look @proppy

xls/examples/parameterized/BUILD
```diff
@@ -0,0 +1,105 @@
...
+    textproto = ":wide.textproto",
+)
+
+xls_dslx_library(
+    name = "parameters_wide_lib",
+    srcs = [":parameters_wide"],
+)
+
+# Design implementation rules - narrow variant
+xls_dslx_library(
```
proppy, line 43: I think we can have the parameters be included directly in the library.

xls/build_rules/xls_utilities.bzl
```diff
@@ -306,3 +307,80 @@ Examples:
...
+      DefaultInfo provider
+    """
+    schema_sources = ctx.attr.proto_schema[ProtoInfo].direct_sources
+    schema = schema_sources[0]
+
+    proto2dslx = ctx.executable._proto_to_dslx_tool
+    dslx_filename = get_output_filename_value(
+        ctx,
+        "dslx_file",
+        ctx.attr.name + ".x",
```
proppy, line 327: should we allow to overwrite the output path?

xls/examples/parameterized/BUILD
```diff
@@ -0,0 +1,105 @@
...
+)
+
+proto_library(
+    name = "proto_schema_lib",
+    srcs = [":param.proto"],
+)
+
+# Parameters for the narrow variant of the design
+proto_to_dslx(
+    name = "parameters_narrow",
```
proppy, line 18: if we used the same output path for both would it create a conflict with bazel?

xls/examples/parameterized/design_wide.x
```diff
@@ -0,0 +1,19 @@
+import xls.examples.parameterized.parameters_wide as parameters
```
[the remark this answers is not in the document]
lpawelcz: Provided only one variant of the design

xls/examples/parameterized/BUILD
lpawelcz: Yes, that's correct

xls/build_rules/xls_utilities.bzl
lpawelcz: Isn't the `dslx_file` attribute enough? It's used when it's specified in the rule, otherwise it defaults to `ctx.attr.name + ".x"`

xls/examples/parameterized/BUILD
```diff
@@ -0,0 +1,105 @@
...
+        "use_system_verilog": "false",
+    },
+    verilog_file = "design_narrow.v",
+)
+
+# Design implementation rules - wide variant
+xls_dslx_library(
+    name = "design_wide_lib",
+    srcs = ["design_wide.x"],
+    deps = [":parameters_wide_lib"],
```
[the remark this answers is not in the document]
lpawelcz: Attempting to put results of `proto_to_dslx` target straight in `xls_dslx_library` in `src` attribute beside `design.x` results in [error](https://github.com/google/xls/blob/919e7d8e61465973259b3287d703bcdc3ff2d12f/xls/build_rules/xls_ir_rules.bzl#L596):
```
Error in fail: A single source file must be specified.
```

If I understand correctly, this would probably require keeping rules for building IRs from each source DSLX file and then merging those in codegen.

xls/examples/parameterized/BUILD
lpawelcz: Please, see comment below.

xls/build_rules/xls_utilities.bzl
cdleary: Resolution SGTM we can expand as we find a need.

xls/examples/parameterized/BUILD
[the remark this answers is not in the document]
cdleary: This doesn't seem great for usability but no need for it to block landing this example.

xls/examples/parameterized/BUILD
cdleary, line 1: License header please.

xls/examples/parameterized/param.proto
cdleary, line 1: License header please (on all source files).

xls/examples/parameterized/parameters.textproto
cdleary, line 1: License header please -- this one less clear it's needed but I think we do it anyway. :-)

Also would be good to add a schema annotation comment like so: 

```
# proto-file: xls/noc/config/network_config_builder_options.proto
# proto-message: NetworkConfigBuilderOptionsProto
```

https://github.com/google/xls/blob/c006e19fa83253261fe06849ca042d550faedb5c/xls/noc/config/examples/grid_with_row_loopback_network_config_builder_options.textproto#L1

(It helps proto visualization tools.)

xls/examples/parameterized/BUILD
lpawelcz: Done

xls/examples/parameterized/design.x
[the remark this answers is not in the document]
lpawelcz: Done

xls/examples/parameterized/param.proto
lpawelcz: Done

xls/examples/parameterized/parameters.textproto
lpawelcz: Done

PR #1212, opened 2023-11-29
modules/zstd: add raw block decoder

rw1nkler:
This PR adds RawBlockDecoder Proc which is responsible for decoding ZSTD Raw Blocks. More information about Raw Block's format can be found in [Section 3.1.1.2.2 of RFC8788](https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.2.2). The Proc acts as a simple passthrough for the incoming data. It will be a part of the larger decoding pipeline.

NOTE: this is based on https://github.com/google/xls/pull/1168 , please ignore commits from that branch when reviewing.

CC @proppy


rw1nkler:
Closing, as the review will happen in #1315

Issue #1211, opened 2023-11-29
Add DSLX Zstandard Codec


rw1nkler:
This is an umbrella issue to track the implementation of the DSLX Zstandard Decoder and Encoder.

The created modules should be compatible with [RFC8878](https://datatracker.ietf.org/doc/html/rfc8878). The encoder, when fed with data should provide an output encoded in Zstandard format. The same data decompressed with the original [zstd library](https://github.com/facebook/zstd) should restore the data to its initial form. Similarly, the created decoder should be able to decompress data encoded in Zstandard format and provide the original data on its output (if the memory requirements necessary for decompressing the stream are fulfilled by the decoder).

PR #1213, opened 2023-11-30
modules/zstd: add rle block decoder

lpawelcz:
This PR adds `RleBlockDecoder` Proc which is responsible for decoding ZSTD RLE Blocks. More information about RLE Block's format can be found in [Section 3.1.1.2.2 of RFC8788](https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.2.2).
`RleBlockDecoder` communicates through `BlockDataPacket` channels.
It reuses existing `RunLengthDecoder` block which is interfaced through
two seprate procs:
 * `RleDataPacker`
 * `BatchPacker`
 
Those are responsible for converting input data into format accepted by
RLE decoder and for gathering RLE decoder output symbols into batches
which are then send out through `BlockDataPacket`.

NOTE: this is based on https://github.com/google/xls/pull/1212 , please ignore commits from that branch when reviewing.

This is part of https://github.com/google/xls/issues/1211

CC @proppy

### DIFF
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -49,3 +49,42 @@ cc_library(
     ],
     visibility = ["//visibility:public"],
 )
+
+# NOTE: Required because of direct zstd_compress.c include in decodecorpus sources
+cc_library(
+    name = "decodecorpus_includes",
+    hdrs = [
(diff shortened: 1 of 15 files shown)


proppy:
should we close this and focus on reviewing https://github.com/google/xls/pull/1315 ?

lpawelcz:
Closing. The review will take place in https://github.com/google/xls/pull/1315

Issue #1127, opened 2023-09-14
Emit actual value along with enum name in cpptranspiler


meheff:
The cpptranspiler generates C++ types from DSLX types. The C++ types include some methods on them included `ToString`. For enum types the name is printed or `<unknown>` for invalid values. We should also print the numeric value. Example:

```
pub enum MyEnum : u32 {
  kA = 0,
  kB = 1,
  kC = 42,
}
```
Expected results:

```
MyEnumToString(0) => "MyEnum::kA (0)"
MyEnumToString(42) => "MyEnum::kC (42)"
MyEnumToString(123) => "<unknown> (123)"
```

PR #1219, opened 2023-12-09
Emit actual value along with enum name

fredmon3:
This pull request adds enum values adjacent to enum names in their string representations.

Fixes issue #1127.


meheff:
Nice! Thanks for the contribution.

Issue #931, opened 2023-04-13
Migrate XLS to bzlmod


cdleary:
Bazel is moving to a new system for managing external dependencies called bzlmod. To conform to this future we'll need to migrate from using our WORKSPACE file to using a MODULE.bazel file, and presumably registering ourselves in the Bazel Central Registry: https://github.com/bazelbuild/bazel-central-registry

See also the bzlmod migration guide: https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit

I'm not sure how challenging XLS is to migrate in the grand scheme of things -- I imagine we have at least a few complicating factors:

* "Large" dependencies
  * rules_hdl
  * grpc
  * llvm, z3, ortools etc
  * we also want to start taking a dep on Verible for LSP utilities
 * pybind extensions
 * pip_install environment
 * clang-format pulled down via toolchain
 * an optional config that allows use of hermetic toolchain (clang/llvm)

If I'm understanding correctly we probably need to convert to bzlmod from leaves (of all transitive deps) up to the "root" of XLS's workspace. Because converting everything to bzlmod is not going to happen atomically it seems like generally folks would support two concurrent systems in the intermediate state, and the bzlmod would have a "more stripped down" sort of build configuration.

This new system will also have compatibility specifications, see https://docs.google.com/document/d/1ReuBBp4EHnsuvcpfXM6ITDmP2lrOu8DGlePMUKvDnXM/edit#heading=h.4lwo83a0dksg

This came up because of the discussion in #865 Henner pointed out we don't declare/pin our (direct) dependencies. The team had a discussion on the tradeoffs -- ISTM that specifying your leaf dependencies up front doesn't mitigate risk in the current WORKSPACE flow and can make more work to debug/update when you bump your dependencies vs just letting them float. If you assume the code you control can flex more easily than code in dependencies, or your requirements for the dependencies are looser, then you prefer that you do not pin. It may come up that you need to pin sometimes, but ISTM from the discussion that should be done only when necessary because of the work savings advantages of letting the dependencies float.

By contrast, with bzlmod we'd have explicit specifications of compatibility that were properly checked against transitive dependencies, so it seems like The Solution, but the migration work seems like it could be significant given the wrinkles listed above. This issue serves as the reminder we want to figure out the right time to do this / work through the wrinkles noted above.

cc @ted-xie who was helping me think through the tradeoffs / potential roadmap here

cdleary:
[...]

Issue #865, opened 2023-02-17
@com_google_protobuf is used, but never declared as dependency in any of the workspace set-up


hzeller:
The com_google_protobuf rule is used as a direct dependency in various rules within xls:

```bash
find . -name "BUILD" | xargs grep @com_google_protobuf
```

Yet, it is never declared as such in any of the workspace set-up, but apparently it is only initialized there, assuming it is magically there:

```bash
$ grep com_google_protobuf WORKSPACE $(find dependency_support/ -type f)
dependency_support/initialize_external.bzl:load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
```

It is only initialized, so it looks like it is 'accidentally' imported by some of the other dependencies that this happens to work ?

It would be good to be able to reason what dependencies come into the project., by either explicit way to either explicitly adding them to the dependency files, or declare them otherwise (don't know if there is an alias of sort, but at least  a comment in the `WORKSPACE` file would help `# @com_google_protobuf provided from from @mystery_import`).

That way it is possible to read and understand and track down weird issues (recent issue: the magically imported google protobuf python rule that was there, but that didn't work).

While at it, maybe there are more rules that are used but never declared ? 
```bash
find xls -name "BUILD" | xargs grep '"@' | sed 's|.*"@\([^"/]*\).*|\1|p' | sort | uniq
```

Essentially, we need an `IWYU` for bazel rules :)

cdleary:
Is there a particular thing for the XLS project's setup to do/fix here? I may be missing the actionable part. Or is this more of a Bazel bug?

In the project as Bazel users we generally assume that we can use constructs like `proto_library`/`proto_cc_library`/the C++ proto libraries/etc and that those do whatever they need to do in the workspace to get set up under an abstraction like `protobuf_deps()` here: https://github.com/google/xls/blob/0dd5beeb63d60a5a730a5e204a19743e0674df4b/dependency_support/initialize_external.bzl#L29

hzeller:
The actionable part would be to add all the dependencies needed in the WORKSPACE or load_external.bzl part to not rely on dependencies (such as `@com_google_protobuf`) that are there 'accidentally'.

The com_google_protobuf is not something that comes with bazel by default, so it maybe is inherit it with something else like grpc (don't know which is is, this is why I stumbled upon it)

cdleary:
[...]

PR #1225, opened 2023-12-12
github/workflows: pin bazel version

proppy:
- bump bazelisk
- pin bazel version in `.bazelversion`

rational: bazel 7.0 is a major update that require some coordinated update of some transitive external deps (see google/xls#865) and switching to bzlmod (see google/xls#931); let's pin it first to un-break the CI.


mithro (approved):
LGTM

mithro:
I think this needs to be landed internally?

proppy:
dropping this as this kind of change needs to be done internally to account for https://github.com/google/copybara conversion logic.

PR #1238, opened 2023-12-22
dslx/stdlib: Add round_up_to_nearest_pow2? functions

rw1nkler:
This PR adds `round_up_to_nearest_pow2?` functions that can be used rounding up to a multiple
of a known power of 2. The function is a special case of the existing `std::round_up_to_nearest()` function which can be implemented more efficiently when we round up to a power of two.

CC @proppy

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -396,6 +396,43 @@ fn round_up_to_nearest_test() {
     assert_eq(u32:4, round_up_to_nearest(u32:4, u32:2));
 }
 
+// Returns `x` rounded up to the nearest multiple of `y`, where `y` is a known positive power of 2.
+// This functionality is the same as `round_up_to_nearest` but optimized when `y` is a power of 2.
+fn round_up_to_nearest_pow2_unsigned<N: u32>(x: uN[N], y: uN[N]) -> uN[N] { (x + y - uN[N]:1) & -y }
+
+#[test]
+fn test_round_up_to_nearest_pow2_unsigned() {
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:0, u16:8), u16:0);
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:1, u16:8), u16:8);
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:7, u16:8), u16:8);
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:8, u16:8), u16:8);
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:9, u16:8), u16:16);
+
+    assert_eq(round_up_to_nearest_pow2_unsigned(u16:9, u16:16), u16:16);
+}
+
+// Returns `x` rounded up to the nearest multiple of `y`, where `y` is a known positive power of 2.
+// This functionality is the same as `round_up_to_nearest` but optimized when `y` is a power of 2.
+fn round_up_to_nearest_pow2_signed<N: u32>(x: sN[N], y: uN[N]) -> sN[N] {
+    (x + y as sN[N] - sN[N]:1) & -(y as sN[N])
+}
+
+#[test]
+fn test_round_up_to_pow2_signed() {
+    assert_eq(round_up_to_nearest_pow2_signed(s16:9, u16:8), s16:16);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:8, u16:8), s16:8);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:7, u16:8), s16:8);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:1, u16:8), s16:8);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:0, u16:8), s16:0);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:-1, u16:8), s16:0);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:-7, u16:8), s16:0);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:-8, u16:8), s16:-8);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:-9, u16:8), s16:-8);
+
+    assert_eq(round_up_to_nearest_pow2_signed(s16:9, u16:16), s16:16);
+    assert_eq(round_up_to_nearest_pow2_signed(s16:-9, u16:16), s16:0);
+}
+
 // Rotate `x` right by `y` bits.
 pub fn rrot<N: u32>(x: bits[N], y: bits[N]) -> bits[N] { (x >> y) | (x << ((N as bits[N]) - y)) }
 
--- a/docs_src/dslx_std.md
+++ b/docs_src/dslx_std.md
@@ -801,6 +801,12 @@ pub fn round_up_to_nearest(x: u32, y: u32) -> u32
 
 Returns `x` rounded up to the nearest multiple of `y`.
 
+#### std::round_up_to_nearest_pow2_?
+
+Returns `x` rounded up to the nearest multiple of `y`, where `y` is a known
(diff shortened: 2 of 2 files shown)


xls/dslx/stdlib/std.x
```diff
@@ -396,6 +396,41 @@ fn round_up_to_nearest_test() {
     assert_eq(u32:4, round_up_to_nearest(u32:4, u32:2));
 }
 
+// Returns `x` rounded up to `y`, where `y` is a positive power of 2
+fn round_up_to_pow2_unsigned<N: u32>(x: uN[N], y: uN[N]) -> uN[N] { (x + y - uN[N]:1) & -y }
```
proppy, line 400: can you try to run https://github.com/google/xls/tree/main/xls/dslx/fmt on this?

xls/dslx/stdlib/std.x
```diff
@@ -396,6 +396,41 @@ fn round_up_to_nearest_test() {
     assert_eq(u32:4, round_up_to_nearest(u32:4, u32:2));
 }
 
+// Returns `x` rounded up to `y`, where `y` is a positive power of 2
```
proppy, line 399: curious why `y` is necessary? wouldn't a general "next power of two" rounding be useful as it? (similar to https://en.cppreference.com/w/cpp/numeric/bit_ceil in c++)

xls/dslx/stdlib/std.x
proppy, line 400: would something like https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2 be more efficient? (I think it should only result in bit slicing and +1 increment)

rw1nkler: This code has already been formatted with `dslx_fmt`

rw1nkler: Bit slices are possible only with a constant start/size. I cannot think of constant boundaries if we have `y` as an argument. Do you have any specific solution in mind? We can also move `y` to a generic parameter to make the slice start/size a compile constant. In that case, we can remove the need for one adder:
```rust
fn round_up_to_pow2_unsigned<Y: u32, N: u32, Y_CLOG2: u32 = { std::clog2(Y) }>(x: uN[N]) -> uN[N] {
    let base = x[Y_CLOG2 as s32:];
    let reminder = x[0:Y_CLOG2 as s32] != bits[Y_CLOG2]:0;
    (base as uN[N] + reminder as uN[N]) << Y_CLOG2
}
```

xls/dslx/stdlib/std.x
rw1nkler: I needed this particular version of a function in calculations of offsets in the ZSTD history buffer. I hope, that soon we will open a public PR that will justify the usage of that function or the function proposed in https://github.com/google/xls/pull/1238#discussion_r1439347935.

xls/dslx/stdlib/std.x
```diff
@@ -396,6 +396,41 @@ fn round_up_to_nearest_test() {
...
+}
+
+// Returns `x` rounded up to `y`, where `y` is a positive power of 2
+fn round_up_to_pow2_signed<N: u32>(x: sN[N], y: uN[N]) -> sN[N] {
+    (x + y as sN[N] - sN[N]:1) & -(y as sN[N])
+}
+
+#[test]
+fn test_round_up_to_pow2_signed() {
+    assert_eq(round_up_to_pow2_signed(s16:9, u16:8), s16:16);
```
proppy, line 420: should this be rounded to `8`?

xls/dslx/stdlib/std.x
```diff
@@ -396,6 +396,41 @@ fn round_up_to_nearest_test() {
...
+// Returns `x` rounded up to `y`, where `y` is a positive power of 2
+fn round_up_to_pow2_unsigned<N: u32>(x: uN[N], y: uN[N]) -> uN[N] { (x + y - uN[N]:1) & -y }
+
+#[test]
+fn test_round_up_to_pow2_unsigned() {
+    assert_eq(round_up_to_pow2_unsigned(u16:0, u16:8), u16:0);
+    assert_eq(round_up_to_pow2_unsigned(u16:1, u16:8), u16:8);
+    assert_eq(round_up_to_pow2_unsigned(u16:7, u16:8), u16:8);
+    assert_eq(round_up_to_pow2_unsigned(u16:8, u16:8), u16:8);
+    assert_eq(round_up_to_pow2_unsigned(u16:9, u16:8), u16:16);
```
proppy, line 408: shouldn't this be rounded to `8`?

xls/dslx/stdlib/std.x
proppy: I was thinking of something like this assuming you just need pow2 w/o the rounding:
```
fn round_up_to_pow2<N: u32, N_CLOG2_PLUS_1: u32 = { std::clog2(N) + u32:1 }>(v: uN[N]) -> uN[N] {
  let v = for (i, v) in u32:0..N_CLOG2_PLUS_1 {
    v | (v >> (uN[N]:1 << i))
  } (v-uN[N]:1);
  v+uN[N]:1
}
```

xls/dslx/stdlib/std.x
proppy, line 399: can you detail a little bit more what "rounded up to `y`" implies for each cases (when x < y, when x == y, when x > y) from reading the test I'm not sure I have a good understanding of it

xls/dslx/stdlib/std.x
rw1nkler: Sorry for the confusion. The function should do the same as the existing `std::round_up_to_nearest()`, but generate much simpler logic.

xls/dslx/stdlib/std.x
rw1nkler: Sorry for the confusion. The function should do the same as the existing std::round_up_to_nearest(), but generate much simpler logic.

xls/dslx/stdlib/std.x
rw1nkler: Yes, see https://github.com/google/xls/pull/1238#discussion_r1466484406 for more details.

xls/dslx/stdlib/std.x
```diff
@@ -396,6 +396,41 @@ fn round_up_to_nearest_test() {
     assert_eq(u32:4, round_up_to_nearest(u32:4, u32:2));
 }
 
+// Returns `x` rounded up to the nearest multiple of `y`, where `y` is a positive power of 2
+fn round_up_to_nearest_pow2_unsigned<N: u32>(x: uN[N], y: uN[N]) -> uN[N] { (x + y - uN[N]:1) & -y }
```
cdleary, line 400: Let's try to drop unsigned as it's already given in the type signature.

(I know right now user defined types can't be parameterized on signedness which is a wart, but I think going with the more convenient name probably makes sense here.)

xls/dslx/stdlib/std.x
cdleary: I think the current bit twiddling form makes sense. When there's dynamic offsets as we've said you effectively need to either bit twiddle or call helper functions in the stdlib.

xls/dslx/stdlib/std.x
cdleary: Can you note this in the comment? i.e. in the second line

// This functionality is the same as `round_up_to_nearest` but more optimized when `y` is a known power of 2.

xls/dslx/stdlib/std.x
cdleary: I guess there's precedent in the sizeof_signed and sizeof_unsigned, so nevermind, this LGTM.

cdleary:
Can you also add to the stdlib reference markdown? i.e.

https://github.com/google/xls/blob/54a102480f0f6969197152856f79949cbe1a2e62/docs_src/dslx_std.md#stdsizeof_

rw1nkler:
> Can you also add to the stdlib reference markdown? i.e.
> 
> https://github.com/google/xls/blob/54a102480f0f6969197152856f79949cbe1a2e62/docs_src/dslx_std.md#stdsizeof_

Added

PR #1294, opened 2024-02-07
Try installing libtinfo5

grebe:
DO_NOT_SUBMIT

Testing a fix to actions


grebe:
Looks like this is sufficient, closing here

PR #1297, opened 2024-02-08
Try partially clearing out the cache.

grebe:


cdleary:
Did we want to merge this, did it work for reducing cache size?

cdleary:
I think with @proppy 's recent changes this is obsolete.

PR #1295, opened 2024-02-08
modules/zstd: Add Sequence Executor

rw1nkler:
This PR adds Proc that performs ZSTD Sequence Execution as described in [rfc8878#section-3.1.1.5](https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.5). The Proc is responsible for interpreting packets with Literals and Sequences obtained from the Procs that decode RAW, RLE, or Compressed blocks of data.

NOTE: this is based on https://github.com/google/xls/pull/1214 , please ignore commits from that branch when reviewing.
This is part of https://github.com/google/xls/issues/1211

PR #1314, opened 2024-02-21
modules/zstd: Add Repacketizer

lpawelcz:
This PR adds `Repacketizer` proc. It is used at the end of the processing flow in the ZSTD decoder. It gathers the output of `SequenceExecutor` proc, removes the bits that are not marked as valid and creates new packets with payloads having only valid bits. Such packets are then send out to the output of the whole ZSTD decoder.

NOTE: this is based on https://github.com/google/xls/pull/1295 , please ignore commits from that branch when reviewing.
This is part of https://github.com/google/xls/issues/1211.

### DIFF
--- a/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
+++ b/dependency_support/com_github_facebook_zstd/bundled.BUILD.bazel
@@ -49,3 +49,42 @@ cc_library(
     ],
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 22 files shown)


proppy:
should we close this and focus on reviewing #1315 ?

lpawelcz:
Closing. The review will take place in https://github.com/google/xls/pull/1315

PR #1327, opened 2024-03-03
replace `polyfill.io` to reduce supply chain risk

seiry:
replace `polyfill.io` to reduce supply chain risk.

ref: https://blog.cloudflare.com/polyfill-io-now-available-on-cdnjs-reduce-your-supply-chain-risk
polyfillpolyfill/polyfill-service#2834

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -91,7 +91,7 @@ markdown_extensions:
   - footnotes
   - attr_list
 extra_javascript:
-  - https://polyfill.io/v3/polyfill.min.js?features=es6
+  - https://cdnjs.cloudflare.com/polyfill/v3/polyfill.min.js?features=es6
   - https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js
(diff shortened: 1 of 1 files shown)


cdleary:
Thanks for the PR -- these docs are generated so we'd want to alter the mkdocs use of polyfill.io -- we don't want to changing generated content because it'll be clobbered by the next HTML generation; e.g. I imagine it comes from configuration like the following: https://github.com/squidfunk/mkdocs-material/blob/15f15068d958506734d3544108a1e6971d7e511b/docs/reference/math.md?plain=1#L67

cdleary:
Oh! I see you did that in the second commit. That's great.

If you can a) revert all the HTML changes and b) sign the CLA then we should be able to c) merge the URL change in the YML and d) regenerate, thanks.

cdleary:
Oh, one last request -- can you squash and force-push the single squashed commit? This is how we take PRs as documented here: https://google.github.io/xls/contributing/#pull-request-style

PR #1214, opened 2023-11-30
modules/zstd: add BlockDecoder

lpawelcz:
This PR adds:

* Block Header parsing functions
* `DecoderMux` Proc, which collects data from specialized Raw, RLE, and Compressed Block decoders and re-sends them in the correct order
*  `DecoderDemux` Proc, which is responsible for parsing Block Header and sending the obtained data to the Raw, RLE or Compressed Block decoders
* `BlockDecoder` Proc, which aggregates the created Decoders together with Muxer and Demuxer. The block is responsible for decoding complete ZSTD Blocks. Currently, it supports RAW and RLE Block types.

NOTE: this is based on https://github.com/google/xls/pull/1213 , please ignore commits from that branch when reviewing.

This is part of https://github.com/google/xls/issues/1211

CC @proppy

PR #1315, opened 2024-02-21
modules/zstd: Add ZstdDecoder top level proc

lpawelcz:
This PR supersedes https://github.com/google/xls/pull/1169
It is a part of https://github.com/google/xls/issues/1211.

This PR adds a top level proc for ZSTD Decoder which integrates all its components in order to allow decoding of ZSTD frames. It includes C++ tests which:
* generate with `decodecorpus` tool valid ZSTD frames 
* decode the frames with `libzstd` and `ZstdDecoder`
* compare the results

NOTE: currently it is possible to decode frames with simple block types as RAW and RLE.

PR #1362, opened 2024-03-28
modules/zstd: Add sequence decoding

rw1nkler:
This PR is meant to provide a proc implementing sequence section decoding as described in:
https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.3.2

NOTE: this is based on https://github.com/google/xls/pull/1214, please ignore commits from that branch when reviewing.


rw1nkler:
> is it independent from #1315 or should they be reviewed together?

We agreed that this should be reviewed as an independent PR after merging #1315

proppy:
should this be rebased now that #1315 has been merged?

PR #1370, opened 2024-04-05
xls/passes/query_engine: fix missing macro

proppy:
This should fix https://github.com/google/xls/actions/runs/8573290500/job/23497685241


ericastor (approved):
Thanks! I agree this should fix it

PR #1378, opened 2024-04-11
Do not register hermetic toolchain for use on OS X.

cdleary:
Currently there is lots of associated breakage, whereas not using the hermetic toolchain and using the platform provided/discovered toolchains instead seems to allow the build to succeed.

This seems like a better state and may get our best-effort nightly builds passing.

### DIFF
--- a/docs_src/BUILD
+++ b/docs_src/BUILD
@@ -30,8 +30,13 @@ exports_files(
 )
 
 filegroup(
-    name = "dslx_reference",
+    name = "dslx_reference_md",
     srcs = ["dslx_reference.md"],
+)
+
+filegroup(
+    name = "dslx_type_system_md",
+    srcs = ["dslx_type_system.md"],
 )
 
 filegroup(
--- a/docs_src/dslx_type_system.md
+++ b/docs_src/dslx_type_system.md
@@ -1,0 +1,207 @@
+# DSLX: Type System
+
+The DSL (frontend) performs a few primary tasks:
+
+1.  Parsing text files to an AST representation.
+2.  Typechecking the AST representation.
+3.  Conversion of the AST to bytecode that can be interpreted.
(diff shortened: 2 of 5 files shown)


dependency_support/llvm_register_toolchains_if_not_macos.bzl
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@llvm_toolchain//:toolchains.bzl", "llvm_register_toolchains")
+
+def _register_toolchain_impl(repository_ctx):
```
proppy, line 17: I wonder if there is a way to declare a system toolchain that point to the macosx runner sysroot similar to what's being done here:
https://github.com/bazel-contrib/toolchains_llvm/blob/3e94f956b6de1c3c6d8c55ccd6f7819925f9770c/tests/WORKSPACE#L90-L95

And then use per-platform toolchain resolution as documented here:
https://github.com/bazel-contrib/toolchains_llvm/tree/3e94f956b6de1c3c6d8c55ccd6f7819925f9770c?tab=readme-ov-file#per-host-architecture-llvm-version

Might be easier than introducing a new rule?

grebe: It might be useful to have a setting that says to use the platform-provided toolchain (unsupported *nixes also relevant here). How do we feel about [bool_flag](https://github.com/bazelbuild/bazel-skylib/blob/main/rules/common_settings.bzl#L82)s?

PR #1390, opened 2024-04-20
Start of C API, amenable to FFI binding.

cdleary:
One step towards a possible `libxls.so` public-and-released artifact -- this starts to expose just the DSLX to IR APIs as canaries to see if we can do what we need to e.g. via Python CFFI.


cdleary:
Superceded by https://github.com/google/xls/pull/1435

PR #1393, opened 2024-05-02
Avoid presenting const char* to XmlNode implicit constructor, was cau…

cdleary:
…sing issues on clang 15 (OS X).


ericastor:
Hi Chris - afraid we'll need the CLA check to pass before we can move forward with this! Please ping us again when you've got that sorted.

PR #1419, opened 2024-05-17
Fix misc-include-cleaner complaints in xls/dslx and xls/passes

hzeller:
fixes ~400ish warnings.


xls/dslx/lsp/lsp_type_utils.cc
```diff
@@ -17,6 +17,9 @@
 #include <string>
 #include <string_view>
 
+#include "common/lsp/lsp-protocol.h"
```
hongted, line 20: This should be #include "external/verible/common/lsp/lsp-protocol.h"

hzeller: Technically, `clang-tidy` is correct, as this is the path it is visible at (but it is _also_ visible as external/verible).
Ideally we can have the set-up in a way that `verible/common/lsp/lsp-protocol.h` would be a good path. Maybe somewhere in the bazel WORKSPACE set-up ? Will have a look at it next week.

PR #1420, opened 2024-05-18
CI: run bazel with --noshow_progress to be less noisy in output.

hzeller:
The github action terminal output can't deal with the in-place update of the Compiling status updates and spams endless lines with the regular refresh, hiding actual potential issues in a wall of text.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -37,8 +37,8 @@ jobs:
 
       - name: Bazel Build Tools (opt)
         run: |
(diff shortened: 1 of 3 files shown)


proppy:
This also seems to swallow warnings? is that intentional?

Before:
https://github.com/google/xls/actions/runs/9198220454/job/25300448344
After:
https://github.com/google/xls/actions/runs/9135704943/job/25123465056

hzeller:
No, warnings should be shown. But since this is a mostly cached build, they won't show up.

proppy (approved):

PR #1488, opened 2024-06-18
Fix tags propagation in bazel macros

### LINKED ISSUE #1487: Bazel macros don't propagate tags correctly
**Describe the bug**
Bazel macros `xls_dslx_verilog` and `xls_benchmark_ir` seem to not propagate `tags` attribute correctly to underlying rules and other macros.

**To Reproduce**
1. Tag the instances of `xls_dslx_verilog` and `xls_benchmark_ir` macros that are in the same bazel package with the `manual` tag. You can use the following patch: [reproduce_tags_propagation_bug_diff.txt](https://github.com/user-attachments/files/15888003/reproduce_tags_propagation_bug_diff.txt) or the [branch](https://github.com/antmicro/xls/tree/61129-tag-propagation-reproduce) on the fork repository (please do not mind the DSLX code, it's just an example to base the work on bazel macros upon). 
2. Run `bazel test` with the wildcard pattern pointing to the package that defines both macro targets:
```
bazel test //xls/modules/zstd/test/...
```
3. After the test completes, observe the build artifacts from `*_tagged` targets under `bazel-bin/xls/modules/zstd/test`  directory confirming that rules/macros tagged as `manual` are indeed being executed.

 

**Expected behavior**
Every target tagged as `manual` should be skipped when wildcard pattern is used for the build/test.

**Screenshots**
N/A

N/A

**Additional context**
N/A

lpawelcz:
Fixes #1487 

Propagate tags in `xls_dslx_verilog` and `xls_benchmark_ir` bazel macros.

Tags were not propagated correctly which resulted in targets tagged as 'manual' still being executed when running bazel with `//...` or `//:all` patterns

### DIFF
--- a/xls/build_rules/xls_macros.bzl
+++ b/xls/build_rules/xls_macros.bzl
@@ -275,6 +275,7 @@ def xls_dslx_verilog_build_and_test(
     build_test(
         name = "__" + name,
         targets = [":" + name],
+        tags = kwargs["tags"] if "tags" in kwargs else [],
     )
 
 def xls_dslx_opt_ir_macro(
@@ -628,6 +629,7 @@ Examples:
         name = opt_ir_target,
         src = src,
         opt_ir_args = full_opt_args,
+        tags = tags,
     )
 
     codegen_target = "{}.default_{}.codegen".format(name, delay_model)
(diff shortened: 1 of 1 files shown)


xls/build_rules/xls_macros.bzl
```diff
@@ -275,6 +275,7 @@ def xls_dslx_verilog_build_and_test(
     build_test(
         name = "__" + name,
         targets = [":" + name],
+        **kwargs,
```
ericastor, line 278: It seems maybe worth grabbing the `tags` argument specifically; I'm not sure we want to pass all kwargs through to the build_test.

lpawelcz: Yeah, you are right. I've changed this, thanks for the suggestion!

PR #1501, opened 2024-06-28
[DSLX:TS] Fully qualify struct type in error message when strings would otherwise be identical

### LINKED ISSUE #1500: Confusing `XlsTypeError` when identical structs are defined in two separate files
**Describe the bug**
When two identical structs are defined in separate files and one is used as the other, the resulting `XlsTypeError` is confusing.

**To Reproduce**
1. Write the following code:

`foo.x`
```rust
import bar;

struct Bar {
    a: u1,
}

fn fn_foo() {
    let b = Bar { a: u1:0 };
    bar::fn_bar(b);
}
```
`bar.x`
```rust
struct Bar {
    a: u1,
}

pub fn fn_bar(arg: Bar) {}
```
2. Build DSLX target.
```bazel
xls_dslx_library(
    name = "bar_dslx",
    srcs = ["bar.x"],
)

xls_dslx_library(
    name = "foo_dslx",
    srcs = ["foo.x"],
    deps = ["bar_dslx"],
)
```
```bash
bazel build //:foo_dslx
```
3. Observe the error message:
```none
foo.x:9:17-9:18
0007: fn fn_foo() {
0008:     let b = Bar { a: u1:0 };
0009:     bar::fn_bar(b);
~~~~~~~~~~~~~~~~~~~~~~^ XlsTypeError: Mismatch between parameter and argument types.
Type mismatch:
   Bar { a: uN[1] }
vs Bar { a: uN[1] }
0010: }
Error parsing and type checking DSLX source file: foo.x
Target //:foo_dslx failed to build
```

**Expected behavior**
The error should contain information why these types differ.

 - OS: Arch Linux x86_64
 - Kernel version: 6.9.6-arch1-1
 - XLS hash: `def000aaa4ec5745397a21c4eeb051760f08e855`

cdleary:
Fixes #1500 

* Adds `ToStringInternal(FullyQualify)` on types.
* `ToString` defaults to `FullyQualify::kNo`
* Switch a few spots that were using ostreams from using `ToString()` explicitly to using the `operator<<` which is defined to be `ToString()`
* In the case where we get a type mismatch where the "want" and "got" are identical, instead we do `ToStringFullyQualified()` and use that instead.

### DIFF
--- a/xls/dslx/tests/errors/error_modules_test.py
+++ b/xls/dslx/tests/errors/error_modules_test.py
@@ -1193,10 +1193,11 @@ class ImportModuleWithTypeErrorTest(test_base.TestCase):
     # virtue of being defined in two different modules, we fully qualify the
     # module where the struct came from in the error message.
     stderr = self._run('xls/dslx/tests/errors/nominal_type_mismatch.x')
-    self.assertIn(
+    self.assertRegex(
+        stderr,
         'Type mismatch:\n'
-        '   xls/dslx/tests/errors/mod_simple_struct.x:MyStruct {}\n'
-        'vs xls/dslx/tests/errors/mod_simple_struct_duplicate.x:MyStruct {}\n', stderr
+        '   .*/dslx/tests/errors/mod_simple_struct.x:MyStruct {}\n'
+        'vs .*/dslx/tests/errors/mod_simple_struct_duplicate.x:MyStruct {}\n'
     )
 
 
--- a/xls/dslx/tests/errors/error_modules_test.py
+++ b/xls/dslx/tests/errors/error_modules_test.py
@@ -1188,6 +1188,17 @@ class ImportModuleWithTypeErrorTest(test_base.TestCase):
         ' which is not present in the parametric environment', stderr
     )
 
+  def test_nominal_type_mismatch(self):
+    # When the types are structurally identical and also have the same name by
+    # virtue of being defined in two different modules, we fully qualify the
+    # module where the struct came from in the error message.
+    stderr = self._run('xls/dslx/tests/errors/nominal_type_mismatch.x')
+    self.assertIn(
+        'Type mismatch:\n'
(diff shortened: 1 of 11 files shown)


hzeller (approved):
Nice!

xls/dslx/tests/errors/error_modules_test.py
```diff
@@ -1188,6 +1188,17 @@ def test_gh_1473(self):
         ' which is not present in the parametric environment', stderr
     )
 
+  def test_nominal_type_mismatch(self):
+    # When the types are structurally identical and also have the same name by
+    # virtue of being defined in two different modules, we fully qualify the
+    # module where the struct came from in the error message.
+    stderr = self._run('xls/dslx/tests/errors/nominal_type_mismatch.x')
```
hzeller, line 1195: This string ...

xls/dslx/tests/errors/error_modules_test.py
```diff
@@ -1188,6 +1188,17 @@ def test_gh_1473(self):
...
+    self.assertIn(
+        'Type mismatch:\n'
+        '   xls/dslx/tests/errors/mod_simple_struct.x:MyStruct {}\n'
+        'vs xls/dslx/tests/errors/mod_simple_struct_duplicate.x:MyStruct {}\n', stderr
```
hzeller, line 1199: and this
might get a bit long if someone was to prepend some prefix, say, `third_party/` to `xls/dslx/` :)

Are there Python-ways to break these ? Backslash or other concatenation ?

cdleary: Ah good point, let me change it to assertMatchesRegexp so prefixes can be there.

allight:
Ok I've updated copybara so in the future it should just do a format-run before importing so hopefully issues with lines too long won't cause issues in the future.

allight (approved):

PR #1503, opened 2024-07-01
Fix missing and superfluous includes in *.cc files.

hzeller:
Using clang-tidy --checks="-*,misc-include-cleaner" --fix

The resulting dependencies have been build cleaned with

  bant dwyu ...

(ignoring the existing independent 6 findings due to protobuf).


hzeller:
Note, the c++ files have been explicitly _not_ cleaned with clang-format (clang-tidy duplicates the includes for each finding) as I want to see how they are cleaned up on import after Alex' recent auto clang-format on import.

PR #1505, opened 2024-07-02
Fix missing and superfluous includes in *.h files.

hzeller:
Using clang-tidy --checks="-*,misc-include-cleaner" --fix

The resulting dependencies have been build cleaned with

 bant dwyu ...

(ignoring the existing independent 6 findings due to protobuf).

Include cleaning with clang-tidy --fix, subsequent file dependency update with `bant dw ...`


hzeller:
Note: deliberately not `clang-format`'ed, as this is happening on import.

PR #1529, opened 2024-08-08
De-pythonify the rendering of op specification, now in C++

cdleary:
Get rid of `op_specification.py` and `render_specification_from_template.py`

Try to help enable more build environments more easily, and continuing our theme of attempted hard-Python-requirement independence.

(Expecting things haven't changed on "Python is good for scripts, and way better than bash scripts, but ideally wouldn't be a required part of the env to use XLS in fundamental ways".)


allight:
To be honest I kind of prefer the python version. I guess that might be partly just familiarity though.

It's nice to be able to see something like:

```
{% if op_class.has_data_members() %}
bool {{ op_class.name }}::IsDefinitelyEqualTo(const Node* other) const {
  if (this == other) {
    return true;
  }
  if (!Node::IsDefinitelyEqualTo(other)) {
    return false;
  }

  return {{ op_class.equal_to_expr() }};
}
{% endif %}
```

instead of having to track down the reference through multiple files.

Are there really any build environments where a python that's at most 2-3 years old is an issue?

ericastor:
I thought the reason we wanted to remove the earlier pybind dependency was that it was intimately dependent on *exactly* which Python the user had in their environment. Using Python for access to a template language seems within scope of "Python is good for scripts"?

meheff (approved):
This seems fine to me. Python *is* better suited for generating and manipulating text, however, I'm not sure the additional complications this causes in the build process for use outside of bazel justifies its use. C++ for this purpose is a bit more complicated but works fine.

Though this is ok to land, I would actually go a step further and just check in the generated c++ sources for ops and nodes perhaps after a bit of cleanup. Generating these sources has always been a mixed bag. For some common attritubes/methods it is great, but so many ops have some unique thing you'd like to do and you either have to hack the python to make it more complicated to support the special case or handle it on the side via functions if possible. Given it's downsides I'd prefer to go with the simple thing and just use plain C++.

I've always been on the fence about generating the op and node sources rather than just checking them in

FWIW, I'm not against the use of python in general. I think it's fine to use for scripts outside of generating C++.

meheff:
On second thought, let's  just go ahead and check in the generated op/node files which obviates this PR. I'll work on that internally.

proppy:
@meheff 

> I would actually go a step further and just check in the generated c++ sources for ops and nodes perhaps after a bit of cleanup. 

👍  @meheff: would that mean that we'd use the generated code as source of truth for ops specification from now on?

meheff:
> 👍 @meheff: would that mean that we'd use the generated code as source of truth for ops specification from now on?

Yes.  We can just check in the generated code as is, then build and improve it over time. Writing the C++ directly gives us more flexibility (admittedly at the cost of some invariant enforcement that the python-generated code provided). I have a change out for review internally.

cdleary:
Closing as it's obsoleted by https://github.com/google/xls/commit/1578b44caac9cc63fc96924f09197d1d03dcd64c -- thanks for helping make XLS easier to build in more environments!

Issue #1476, opened 2024-06-12
DSLX formatter removes comments around proc channel declarations


mikex-oss:
**Describe the bug**
After running through dslx autofmt, the comments describing a channel declaration are gone.

**To Reproduce**
1. Take the example [adder proc](https://google.github.io/xls/tutorials/what_is_a_proc/#channels:~:text=to%20a%20third%3A-,pub%20proc%20adder%20%7B,-A%3A%20chan) in the proc tutorial and add comments above each channel declaration.
```
pub proc adder {
  // foo
  A: chan<u32> in;
  // bar
  B: chan<u32> in;
  // baz
  C: chan<u32> out;
```

2. Run the formatter or dslx_fmt_test.
3. You will see the comments removed; test diff shown below:

```
<   // foo
<   A: chan<u32> in;
<   // bar
<   B: chan<u32> in;
<   // baz
<   C: chan<u32> out;
---
>     A: chan<u32> in;
>     B: chan<u32> in;
>     C: chan<u32> out;
```

**Expected behavior**
Comments should be preserved.

mikex-oss:
Note inline comments and blank lines (between comment and declaration, or between declarations) doesn't seem to workaround this issue.

cdleary:
We think this should be fixed by #1535

PR #1535, opened 2024-08-08
autoformat comments on proc members and functions

### LINKED ISSUE #1531: DSLX autoformatter erases comments within `proc` but outside of next/config/init
**Describe the bug**
The autoformatter will remove comments where it should not.

**To Reproduce**
```
proc f {
  config() {}
  init { }
  // TODO(jmolloy): implement
  next(state: ()) { }
}
```

Autoformatter will remove the comment.

 - Google internal Cider-V.

cdleary:
Fixes #1531 

Leverage the existing `EmitCommentsBetween` helper.

Procs are slightly interesting because we also reorder the proc functions into their canonical order.

As a result, during the pre-pass we emit members with their associated comments, but gather up spans that are relevant to the proc member functions.

Once the pre-pass is complete we can emit the member functions with comments.

Added a "body span" to the `ProcLike` type so we know where the brace open/close are for the `proc` body.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ compile_commands.json
 xls_clang-tidy.out
 user.bazelrc
 .vscode
+.cache
--- a/xls/dslx/fmt/ast_fmt.cc
+++ b/xls/dslx/fmt/ast_fmt.cc
@@ -51,6 +51,67 @@
 namespace xls::dslx {
 namespace {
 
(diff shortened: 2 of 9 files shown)


cdleary:
Sorry forgot the fuzzer constructed these manually and had to be updated, redness should be fixed shortly.

richmckeever (approved):

mikex-oss:
Also fixes https://github.com/google/xls/issues/1476?

PR #1544, opened 2024-08-17
Add API to free a C string via the XLS C API.

cdleary:
If the allocate is not the same between the XLS library and the caller environment it's convenient to say "hey XLS library free this string you gave me ownership of", then even if XLS is compiled with e.g. tcmalloc and a Rust binary is compiled with mimalloc everything works out ok.


meheff (approved):

xls/public/c_api.cc
```diff
@@ -212,6 +212,13 @@ void xls_package_free(struct xls_package* p) {
   delete reinterpret_cast<xls::Package*>(p);
 }
 
+void xls_c_str_free(char* c_str) {
+  if (c_str == nullptr) {
```
hzeller, line 216: FYI: No need to compare against nullptr. `free()` is guaranteed to handle the case.

PR #1545, opened 2024-08-18
Correct the type for bytecode evaluation of enum value (via ColonRef)

### LINKED ISSUE #1541: DSLX bytecode interpreter should create values of the same shape with ZeroMacro or Literal.
Consider the following: 
```
enum MyEnum : s3 {
  VAL_0 = 0,
}

struct EnumInStruct {
  e: MyEnum,
}

fn cast_enum_to_bits() -> s32 {
  let foo = EnumInStruct { e: MyEnum::VAL_0 };
  let a = (foo.e as s32) - s32:1;
  a
}
```
The bytecode interpreter creates foo as a tuple of u32:0 -- already evaluating the enum.

https://github.com/google/xls/blob/aae7ee2025f6dd33e1eaee490a24c918c36ac0ec/xls/dslx/bytecode/bytecode_emitter.cc#L1460

```
bytecode_interpreter.cc:137] Bytecode: literal s3:0 @ test.x:10:38-10:45
bytecode_interpreter.cc:138] PC: 0 : literal s3:0 @ test.x:10:38-10:45
bytecode_interpreter.cc:140]  - stack depth 0 []
interpreter_stack.h:75] PushFormattedValue(s3:0)
bytecode_interpreter.cc:144]  - stack depth 1 [s3:0]
bytecode_interpreter.cc:137] Bytecode: create_tuple 1 @ test.x:10:26-10:48
bytecode_interpreter.cc:138] PC: 0x1 : create_tuple 1 @ test.x:10:26-10:48
bytecode_interpreter.cc:140]  - stack depth 1 [s3:0]
interpreter_stack.h:70] Push((s3:0))
bytecode_interpreter.cc:144]  - stack depth 1 [(s3:0)]
bytecode_interpreter.cc:137] Bytecode: store 0 @ test.x:10:7-10:10
```
In contrast use of the ZeroMacro with the following

```
fn cast_enum_to_bits() -> s32 {
  let foo = zero!<EnumInStruct>();
  let a = (foo.e as s32) - s32:1;
  a
}
```

Results in the bytecode interpreter created a tuple with an enum typed value inside.
```
typecheck_function.cc:137] Typechecking fn: cast_enum_to_bits
constexpr_evaluator.cc:240] ConstexprEvaluator::HandleBinop : (foo.e as s32) - s32:1
constexpr_evaluator.cc:278] ConstexprEvaluator::HandleCast : (foo.e as s32)
bytecode_interpreter.cc:137] Bytecode: literal (MyEnum:0) @ test.x:11:15-11:17
bytecode_interpreter.cc:138] PC: 0 : literal (MyEnum:0) @ test.x:11:15-11:17
bytecode_interpreter.cc:140]  - stack depth 0 []
interpreter_stack.h:75] PushFormattedValue((MyEnum:0))
bytecode_interpreter.cc:144]  - stack depth 1 [(MyEnum:0)]
```

This should be made consistent

_Originally posted by @hongted in https://github.com/google/xls/issues/1536#issuecomment-2290106284_

cdleary:
Fixes #1541

- Makes the type of the constexpr value annotated on enum members the type of the enum itself -- this causes bytecode emission to pick up the correctly typed value in the constexpr map
- Display the Bits value underlying the enum more uniformly
- Removes dead PrettyPrintValue

Note that this change found an issue only present in downstream code samples that was then surfaced as a new example. This subsequent issue was for constexpr evaluation of arrays with ellipsis.

There were cases where we were confusing:
- the number of members in the array literal with
- the array size, even though there were ellipses (in which case the number of members is determined by the type annotation)

Also removed accidental complexity in the course of making it consistent -- now `ConstexprEvaluator` calls into the bytecode intepreter directly instead of reimplementing another interpretation routine for `HandleArray`.

### DIFF
--- a/xls/dslx/type_system/type_info.cc
+++ b/xls/dslx/type_system/type_info.cc
@@ -155,7 +155,7 @@ absl::StatusOr<TypeInfo*> TypeInfoOwner::GetRootTypeInfo(const Module* module) {
 // -- class TypeInfo
 
 void TypeInfo::NoteConstExpr(const AstNode* const_expr, InterpValue value) {
-  VLOG(1) << absl::StreamFormat(
+  VLOG(5) << absl::StreamFormat(
       "noting node: `%s` (%p) has constexpr value: `%s`",
       const_expr->ToString(), const_expr, value.ToString());
   const_exprs_.insert_or_assign(const_expr, std::move(value));
--- a/xls/dslx/type_system/type_info.cc
+++ b/xls/dslx/type_system/type_info.cc
@@ -163,7 +163,10 @@ void TypeInfo::NoteConstExpr(const AstNode* const_expr, InterpValue value) {
   if (it != const_exprs_.end() && it->second.has_value()) {
     // If the tags are the same, assert they're the same value.
     if (value.tag() == it->second->tag()) {
-      CHECK(value.Eq(it->second.value())) << "previous value: " << it->second->ToString() << " inserting value: " << value.ToString();;
+      CHECK(value.Eq(it->second.value()))
+          << "previous value: " << it->second->ToString()
+          << " inserting value: " << value.ToString();
+      ;
     }
   }
 
--- a/xls/dslx/type_system/type_info.cc
+++ b/xls/dslx/type_system/type_info.cc
@@ -159,16 +159,19 @@ void TypeInfo::NoteConstExpr(const AstNode* const_expr, InterpValue value) {
       "noting node: `%s` (%p) has constexpr value: `%s`",
       const_expr->ToString(), const_expr, value.ToString());
 
-  auto it = const_exprs_.find(const_expr);
-  if (it != const_exprs_.end() && it->second.has_value()) {
-    // If the tags are the same, assert they're the same value.
-    if (value.tag() == it->second->tag()) {
-      CHECK(value.Eq(it->second.value()))
-          << "previous value: " << it->second->ToString()
-          << " inserting value: " << value.ToString();
-      ;
-    }
-  }
+  // Note: this assertion will generally hold as of 2024-08-23, except in the
+  // case of `UnrollFor` nodes, which https://github.com/richmckeever is
+  // actively working on alternative strategies for.
+  //
+  // auto it = const_exprs_.find(const_expr);
+  // if (it != const_exprs_.end() && it->second.has_value()) {
+  //   // If the tags are the same, assert they're the same value.
+  //   if (value.tag() == it->second->tag()) {
+  //     CHECK(value.Eq(it->second.value()))
+  //         << "previous value: " << it->second->ToString()
+  //         << " inserting value: " << value.ToString();
+  //   }
+  // }
 
   const_exprs_.insert_or_assign(const_expr, std::move(value));
 }
--- a/xls/dslx/type_system/type_info.cc
+++ b/xls/dslx/type_system/type_info.cc
@@ -158,6 +158,15 @@ void TypeInfo::NoteConstExpr(const AstNode* const_expr, InterpValue value) {
   VLOG(5) << absl::StreamFormat(
       "noting node: `%s` (%p) has constexpr value: `%s`",
       const_expr->ToString(), const_expr, value.ToString());
+
+  auto it = const_exprs_.find(const_expr);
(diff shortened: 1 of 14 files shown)


meheff (approved):

xls/dslx/type_system/type_info.cc
```diff
@@ -155,7 +155,10 @@ absl::StatusOr<TypeInfo*> TypeInfoOwner::GetRootTypeInfo(const Module* module) {
 // -- class TypeInfo
 
 void TypeInfo::NoteConstExpr(const AstNode* const_expr, InterpValue value) {
-  const_exprs_.insert({const_expr, value});
+  VLOG(5) << absl::StreamFormat(
+      "noting node: `%s` (%p) has constexpr value: `%s`",
+      const_expr->ToString(), const_expr, value.ToString());
+  const_exprs_.insert_or_assign(const_expr, std::move(value));
```
hzeller, line 161: FYI, re review delay: The replacement of the const expr value sounds benign, however it breaks some other code outside this repository which I have yet to narrow down (time-slicing myself this week).

Highlevel: apparently in some for-loop in a proc, `NoteConstExpr()` is called with some array as value, and later for the same `AstNode` with a shorter array ... then things go south when some place attempts to access a non-existent index on the shorter array...
I've yet to determine the details. Will write a reproduction and fix when time permits.

hzeller: Alright. The new https://github.com/google/xls/blob/main/xls/examples/memory_proc.x serves as a useful example for beginners, but it will also be a repo.

Essentially, the state, which is an array representing the 'memory' is replaced with a single value. Which might actually might be different problem somewhere else.

cdleary: Ok found the issue, it was for constexpr evaluation of arrays with ellipsis.

There were cases where we were confusing the number of members in the array literal with the array size even though there were ellipses.

Also removed accidental complexity in the course of making it consistent -- now ConstexprEvaluator calls into the bytecode intepreter directly instead of reimplementing another interpretation routine.

cdleary: Forgot to tag you @hzeller -- thanks for the reproducer!

xls/examples/ram.x
```diff
@@ -79,8 +79,14 @@ fn expand_mask<DATA_WIDTH:u32, NUM_PARTITIONS:u32,
...
   partition_mask:uN[NUM_PARTITIONS]) -> uN[DATA_WIDTH] {
   for (idx, data_mask): (u32, uN[DATA_WIDTH]) in range(u32:0, NUM_PARTITIONS) {
-    let data_mask_segment =
-      flatten(u1[EXPANSION_FACTOR]: [partition_mask[idx +: u1], ...]);
+    let mask_bit: u1 = partition_mask[idx +: u1];
+    let data_mask_segment: uN[EXPANSION_FACTOR] =
+      flatten(u1[EXPANSION_FACTOR]:[mask_bit, ...]);
+
+    let to_flatten: u1[EXPANSION_FACTOR] = u1[EXPANSION_FACTOR]:[partition_mask[idx +: u1], ...];
+    let data_mask_segment': uN[EXPANSION_FACTOR] = flatten(to_flatten);
```
grebe, line 87: What about this change interacted with this? I'm not seeing how the enum stuff applies here.

cdleary: This was exhibiting a bug after I did the constexpr fix. I'll hoist the additional commentary from https://github.com/google/xls/pull/1545#discussion_r1729307861 into the PR description.

The type based on elements was a u1[1] but that was ignoring the ellipsis when EXPANSION_FACTOR was 2.

I made these two variations just to check they produced the same answer & typechecked. It's a little grosser but I figured it wouldn't hurt anything since optimizer will know they're the same -- but I can revert this hunk, now that the bug is fixed w/regression test it shouldn't come back.

grebe: Ah, makes sense. Sorry, should have looked at the other comment for more context.

I think the example Henner added serves a similar purpose, right? Seems simpler to revert this hunk.

cdleary: PR description is updated and this hunk is reverted.

hzeller (approved):

PR #1546, opened 2024-08-18
Respect lack of public annotations for colon-refs

### LINKED ISSUE #1507: stdlib functions seem to bypass visibility check
**Describe the bug**
As part of fixing google/xls#1490, I started seeing a visibility failure for mapped functions like `std::is_unsigned_msb_set`(at https://github.com/google/xls/blob/main/xls/examples/protobuf/varint_streaming_decode.x#L98). 

This led me to realize that we have a number of non-`pub` functions in there, and some are being used outside of `map` context without error. 

**To Reproduce**
1. Use the following DSLX code snippet:
```
import std;

fn foo(a: u32) -> u2 {
  std::vslice<u32:1, u32:0>(a)
}
```

Note that vslice is not public: https://github.com/google/xls/blob/main/xls/dslx/stdlib/std.x#L1002
2. Build the DSLX library.
3. No error is generated.

**Expected behavior**
It should report a build time error:

> TypeInferenceError: Attempted to refer to module member fn vslice ... that is not public.

cdleary:
Adds an error test and fixes the resulting fallout in examples.

Note that synthetic proc init functions need to inherit their visibility from the proc they are defined inside.

Fixes #1507

### DIFF
--- a/xls/dslx/frontend/ast.h
+++ b/xls/dslx/frontend/ast.h
@@ -35,7 +35,6 @@
 #include "absl/strings/str_format.h"
 #include "absl/types/span.h"
 #include "absl/types/variant.h"
-#include "xls/common/casts.h"
 #include "xls/common/status/status_macros.h"
 #include "xls/dslx/channel_direction.h"
 #include "xls/dslx/frontend/ast_node.h"  // IWYU pragma: export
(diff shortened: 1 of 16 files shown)


cdleary:
Note the failures are unrelated to this change and reproduce at HEAD (without this PR) -- see https://github.com/google/xls/issues/1547

meheff (approved):

Issue #1547, opened 2024-08-18
`//xls/modules/zstd:frame_header_cc_test` is failing at HEAD


cdleary:
**Describe the bug**

`//xls/modules/zstd:frame_header_cc_test` is failing at HEAD

cc @proppy @hongted 

**To Reproduce**

```
$ bazel test -c opt --test_output=errors -- $(bazel query 'filter(".*_cc_test", kind(rule, //xls/modules/zstd/...))')
[ RUN      ] FrameHeaderTest.FailUnsupportedVeryLargeFrameContentSizeThroughSingleSegment
xls/ir/ir_test_base.cc:269: Failure
Value of: ValuesEqual(expected, result.value)                                                           
  Actual: false ((bits[2]:3, (bits[64]:0, bits[64]:0, bits[32]:0, bits[1]:0), (bits[128]:0x0, bits[32]:0)) != (bits[2]:0, (bits[64]:9216, bits[64]:46596, bits[32]:0, bits[1]:0), (bits[128]:0xe_0000_7ffd, bits
[32]:80)))
Expected: true                                                                                          
(interpreted unoptimized IR)                                                                            
Google Test trace:                                                                                      
xls/modules/zstd/frame_header_test.cc:161: RunAndExpectEq failed                   
Stack trace:
  0x7162133e9289: xls::IrTestBase::RunAndExpectEq()
  0x7162133eb31c: xls::IrTestBase::RunAndExpectEq()                                                     
  0x63103adfcf50: xls::(anonymous namespace)::FrameHeaderTest::RunAndExpectFrameHeader()
  0x63103adffe56: xls::(anonymous namespace)::FrameHeaderTest_FailUnsupportedVeryLargeFrameContentSizeThroughSingleSegment_Test::TestBody()
  0x716208dba8fd: testing::internal::HandleExceptionsInMethodIfSupported<>()
  0x716208dba7be: testing::Test::Run()
  0x716208dbbbc1: testing::TestInfo::Run()
... Google Test internal frames ... 
//xls/modules/zstd:frame_header_cc_test                                  FAILED in 5 out of 50 in 2.7s
  Stats over 50 runs: max = 2.7s, min = 1.1s, avg = 1.4s, dev = 0.3s
  /media/big/_bazel_cdleary/e3482f322694a8d1a6845914a1a3d787/execroot/com_google_xls/bazel-out/k8-opt/testlogs/xls/modules/zstd/frame_header_cc_test/shard_2_of_50/test.log
[...]

PR #1550, opened 2024-08-19
Fix frame header CC tests

lpawelcz:
Fixes https://github.com/google/xls/issues/1547

CC @proppy

This PR fixes the handling of `absl:Span` with test data input vectors for the `frame_header` tests.
It looks like we must copy the vectors over to the `ZSTDFrameHeader` struct and keep those there for the time of the test case execution and use `absl::Span` based on those copies.
Without this modification we observed that few of the test cases will always fail due to incorrect data under the `input_buffer` span. The interesting thing is that when we commented out all test cases except one of the failing ones (e.g. `FrameHeaderTest.Success`) we noticed that this particular test case started passing indicating that the issue can be triggered by execution of multiple test cases at the same time.

This PR also fixes some formatting errors that were picked up with `clang-format version 18.1.8 `

### DIFF
--- a/xls/modules/zstd/BUILD
+++ b/xls/modules/zstd/BUILD
@@ -210,7 +210,6 @@ cc_test(
         ":frame_header_test_dslx",
     ],
     shard_count = 50,
-    tags = ["manual"],
     deps = [
         ":data_generator",
         "//xls/common:xls_gunit_main",
@@ -999,7 +998,6 @@ cc_test(
         ":zstd_dec_test.ir",
     ],
     shard_count = 50,
-    tags = ["manual"],
     deps = [
         ":data_generator",
         "//xls/common:xls_gunit_main",
--- a/xls/modules/zstd/BUILD
+++ b/xls/modules/zstd/BUILD
@@ -158,7 +158,6 @@ cc_library(
         "//xls/common/file:get_runfile_path",
         "//xls/common/status:status_macros",
         "@com_google_absl//absl/algorithm:container",
-        "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/time",
--- a/xls/modules/zstd/BUILD
+++ b/xls/modules/zstd/BUILD
@@ -153,16 +153,16 @@ cc_library(
         "@zstd//:decodecorpus",
     ],
     deps = [
+        "//xls/common:subprocess",
+        "//xls/common/file:filesystem",
+        "//xls/common/file:get_runfile_path",
+        "//xls/common/status:status_macros",
         "@com_google_absl//absl/algorithm:container",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:span",
-        "//xls/common:subprocess",
-        "//xls/common/file:filesystem",
-        "//xls/common/file:get_runfile_path",
-        "//xls/common/status:status_macros",
     ],
 )
 
@@ -213,8 +213,6 @@ cc_test(
     tags = ["manual"],
     deps = [
         ":data_generator",
-        "@com_google_absl//absl/container:flat_hash_map",
-        "@com_google_absl//absl/types:span",
         "//xls/common:xls_gunit_main",
         "//xls/common/file:filesystem",
         "//xls/common/file:get_runfile_path",
@@ -230,8 +228,10 @@ cc_test(
         "//xls/ir:bits",
         "//xls/ir:ir_test_base",
         "//xls/ir:value",
-        "@zstd",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/types:span",
         "@com_google_googletest//:gtest",
+        "@zstd",
     ],
 )
 
@@ -1002,10 +1002,6 @@ cc_test(
     tags = ["manual"],
     deps = [
         ":data_generator",
-        "@com_google_absl//absl/container:flat_hash_map",
-        "@com_google_absl//absl/log",
-        "@com_google_absl//absl/status:statusor",
-        "@com_google_absl//absl/types:span",
         "//xls/common:xls_gunit_main",
         "//xls/common/file:filesystem",
         "//xls/common/file:get_runfile_path",
@@ -1020,8 +1016,12 @@ cc_test(
         "//xls/ir:ir_parser",
         "//xls/ir:value",
         "//xls/jit:jit_proc_runtime",
-        "@zstd",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/log",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/types:span",
         "@com_google_googletest//:gtest",
+        "@zstd",
     ],
 )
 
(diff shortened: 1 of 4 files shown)


xls/modules/zstd/frame_header_test.cc
```diff
@@ -58,9 +58,19 @@ enum FrameHeaderStatus : uint8_t {
 };
 
 struct ZstdFrameHeader {
-  absl::Span<const uint8_t> kBuffer;
   ZSTD_frameHeader kHeader;
   size_t kResult;
+
+  absl::Span<const uint8_t> kBuffer() const {
```
cdleary, line 64: google style is just buffer() since the member is buffer_

xls/modules/zstd/frame_header_test.cc
```diff
@@ -58,9 +58,19 @@ enum FrameHeaderStatus : uint8_t {
...
+    return absl::MakeConstSpan(buffer_);
+  }
+
+  ZstdFrameHeader(absl::Span<const uint8_t> buffer, ZSTD_frameHeader& h,
```
cdleary, line 68: google style says since it has a constructor it should be a class and not a struct

also passing mutable references is less preferrable here than passing by value -- for any large type (which the frameHeader may be) you can std::move the value

xls/modules/zstd/frame_header_test.cc
```diff
@@ -113,7 +126,7 @@ class FrameHeaderTest : public xls::IrTestBase {
   // expected values.
   void RunAndExpectFrameHeader(const ZstdFrameHeader& zstd_frame_header) {
     // Extend buffer contents to 128 bits if necessary.
-    const absl::Span<const uint8_t>& buffer = zstd_frame_header.kBuffer;
+    const absl::Span<const uint8_t>& buffer = zstd_frame_header.kBuffer();
```
cdleary, line 129: Don't take refs to spans, spans are easily copied so we pass them and hold them by value

cdleary (changes requested):

xls/modules/zstd/BUILD
```diff
@@ -213,8 +213,6 @@ cc_test(
     tags = ["manual"],
```
hzeller, line 213: Is there a reason why this test is marked `"manual"` ? Given that this was hiding an issue, it is probably good to have that enabled.
(Only if it is taking too many resources I'd consider it tagging manual. But then with a comment as explanation)

hzeller:
Can you also run 
```
bant dwyu //xls/modules/...
```
 
? Looks like there is an unnecessary `@com_google_absl//absl/status` dependency in `//xls/modules/zstd:data_generator`

xls/modules/zstd/frame_header_test.cc
lpawelcz: Changed from `kBuffer()` to `buffer()`

xls/modules/zstd/frame_header_test.cc
lpawelcz: Changed `ZstdFrameHeader` to class and dropped passing by reference in favor of passing by value + `std::move`

xls/modules/zstd/frame_header_test.cc
lpawelcz: Fixed

lpawelcz:
> Can you also run
> 
> ```
> bant dwyu //xls/modules/...
> ```
> 
> ? Looks like there is an unnecessary `@com_google_absl//absl/status` dependency in `//xls/modules/zstd:data_generator`

I checked the code with `bant` (current `main` at https://github.com/hzeller/bant/commit/d033e1af257141cc545e311a10ea67a56ca03e85) and it doesn't report anything.

xls/modules/zstd/BUILD
lpawelcz: We marked all DSLX and C++ tests as `manual` so that those would be called only in the separate [ZSTD-specific CI workflow](https://github.com/google/xls/blob/main/.github/workflows/modules-zstd.yml). It explicitly calls all of those rules tagged as `manual` and the workflow is triggered when there is a PR that changes the `xls/modules/zstd` directory or there is a push to `main` branch that changes it.

This was first requested in https://github.com/google/xls/pull/1315#issuecomment-2117654785.
After that we also decided to tag some of the `build` rules as `manual` because e.g. `place_and_route` rule is time consuming and it makes sense to not run it in every CI run (see https://github.com/google/xls/pull/1315#discussion_r1661866743).

xls/modules/zstd/frame_header_test.cc
```diff
@@ -57,10 +57,21 @@ enum FrameHeaderStatus : uint8_t {
   UNSUPPORTED_WINDOW_SIZE
 };
 
-struct ZstdFrameHeader {
-  absl::Span<const uint8_t> kBuffer;
+class ZstdFrameHeader {
+ public:
   ZSTD_frameHeader kHeader;
   size_t kResult;
```
cdleary, line 63: These members should be private with accessors and shouldn't be named kSomething because they're not constexpr.

cdleary (approved):
Looks good with the member changes.

Would be good to not mark the c++ tests as manual just the backend targets if possible.

xls/modules/zstd/BUILD
hzeller: @proppy do the reasons for manually running tests still hold ? I think it is more a danger to not run tests all the time, because then hidden issues will go undetected.

hzeller:
> I checked the code with `bant` (current `main` at [hzeller/bant@d033e1a](https://github.com/hzeller/bant/commit/d033e1af257141cc545e311a10ea67a56ca03e85)) and it doesn't report anything.

@lpawelcz 

Interesting. Since off-topic here, can you open a bug in [bant](https://github.com/hzeller/bant), reference this PR and show what you see as output of 

```bash
# -vvv for extra verbose explaining what it is doing internally
bant dwyu -vvv //xls/modules/zstd:data_generator
```

I do get a
```
buildozer 'remove deps @com_google_absl//absl/status' //xls/modules/zstd:data_generator
```

because there is no `absl/status/status.h` included (`grep status.h xls/modules/zstd/data_generator.{h,cc}`) in any of the source files.

xls/modules/zstd/BUILD
proppy: @hzeller the initial reason was that there were wip tests capturing current limitations of the implementation, but I don't believe that's the case anymore is it @lpawelcz ?

cdleary: @proppy Note you can disable them in gtest by calling them `TEST(MyTest, DISABLED_DoStuffThatOnlyCompilesButFailAtRuntime) { ... }`

cdleary: https://google.github.io/googletest/advanced.html#temporarily-disabling-tests

xls/modules/zstd/frame_header_test.cc
lpawelcz: Fixed

xls/modules/zstd/BUILD
lpawelcz: > @hzeller the initial reason was that there were wip tests capturing current limitations of the implementation, but I don't believe that's the case anymore is it @lpawelcz ?

Yes, that was the initial reason that no longer holds.
There is also a second reason which is the length of the CI workflow. We wanted to avoid running all of our tests each time because both the C++ tests and the benchmarks (build rules for synthesis, place and route; run rules for IR benchmarks) are very time consuming.

With the suggestion from @cdleary I removed the `manual` tags for the C++ tests so that at least those will be executed in the main CI workflow.

hzeller (approved):

Issue #1540, opened 2024-08-15
Use of imported type/enum crashes the front end


meheff:
Repro:

my_enum.x:
```
pub enum MyEnum : bits[1] {
  kA = 0,
  kB = 1,
}
```

my_lib.x:
```
import my_enum;

pub type MyEnumAlias = my_enum::MyEnum;
```

my_top.x:
```
import my_lib;

fn Foo () -> my_lib::MyEnumAlias {
  my_lib::MyEnumAlias::kA
}
```

Building my_top:
```
$ bazel build my_top
ERROR: /google/src/cloud/meheff/verilog/google3/experimental/users/meheff/xls/dslx_errors/BUILD:19:17: Parsing and type checking DSLX source files of target my_top failed: (Exit 255) bash failed: error executing ParseAndTypeCheckDSLXSourceFile command (from target //experimental/users/meheff/xls/dslx_errors:my_top) /usr/buildtools/buildhelpers/v4/bin/bash -c ... (remaining 1 argument skipped).  [forge_remote_host=javk34]
F0814 17:22:54.693965    9610 type_info.cc:534] Check failed: import->owner() == module_ (0x13a7bfe20138 vs. 0x13a7bfe20270) Import node from: google3.experimental.users.meheff.xls.dslx_errors.my_lib vs TypeInfo for my_top
*** Check failure stack trace: ***
    @     0x558602707904  absl::log_internal::LogMessage::SendToLog()
    @     0x558602707436  absl::log_internal::LogMessage::Flush()
    @     0x558602707d29  absl::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x5585ff7c8432  xls::dslx::TypeInfo::GetImported()
    @     0x5585ff664f0a  xls::dslx::ResolveColonRefSubjectForTypeChecking()
    @     0x5585ff666d0b  xls::dslx::ResolveTypeAliasToDirectColonRefSubject()
    @     0x5585ff66726a  std::__u::__variant_detail::__visitation::__base::__dispatcher<>::__dispatch<>()
    @     0x5585ff665061  xls::dslx::ResolveColonRefSubjectForTypeChecking()
    @     0x5585ff5d1431  xls::dslx::(anonymous namespace)::DeduceVisitor::HandleColonRef()
    @     0x5585ff5c6341  xls::dslx::Deduce()
    @     0x5585ff59a7d4  std::__u::__function::__policy_invoker<>::__call_impl<>()
[...]

PR #1578, opened 2024-09-01
[DSLX:TS] Fix for ColonRef to enum value within proc parametrics.

cdleary:
Towards google/xls#1540 -- in particular fixes the example given in this comment:

https://github.com/google/xls/issues/1540#issuecomment-2297711953

When deducing a ColonRef we get the root type information for the subject, create a new context to deduce the entity being referred to, and we should accordingly push a top level entry onto the function stack for that referred-to module.

Previously we were copying the most recent entry in the function stack, which for a parametric proc indicated we were typechecking within a proc function, which caused the wrong things to happen downstream in the code.

Tightened up some of the invariant checks as well for when we expect there to be a single top-level entry on the function stack.

### DIFF
--- a/xls/dslx/type_system/scoped_fn_stack_entry.cc
+++ b/xls/dslx/type_system/scoped_fn_stack_entry.cc
@@ -28,11 +28,6 @@ ScopedFnStackEntry::ScopedFnStackEntry(DeduceCtx* ctx, Module* module)
   ctx->AddFnStackEntry(FnStackEntry::MakeTop(module));
 }
 
-// Args:
-//  expect_popped: Indicates that we expect, in the destructor for this scope,
-//    that the entry will have already been popped. Generally this is `false`
-//    since we expect the entry to be on the top of the fn stack in the
-//    destructor, in which case we automatically pop it.
 ScopedFnStackEntry::ScopedFnStackEntry(Function& fn, DeduceCtx* ctx,
                                        WithinProc within_proc,
                                        bool expect_popped)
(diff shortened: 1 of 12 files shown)


xls/dslx/type_system/scoped_fn_stack_entry.cc
```diff
@@ -28,11 +28,6 @@ ScopedFnStackEntry::ScopedFnStackEntry(DeduceCtx* ctx, Module* module)
   ctx->AddFnStackEntry(FnStackEntry::MakeTop(module));
 }
 
-// Args:
```
proppy, line 31: q: any reasons those get removed?

xls/dslx/type_system/scoped_fn_stack_entry.cc
```diff
@@ -42,10 +37,6 @@ ScopedFnStackEntry::ScopedFnStackEntry(Function& fn, DeduceCtx* ctx,
   ctx->AddFnStackEntry(FnStackEntry::Make(fn, ParametricEnv(), within_proc));
 }
 
-// Called when we close out a scope. We can't use this object as a scope
```
proppy, line 45: q: any reasons those get removed?

xls/dslx/type_system/scoped_fn_stack_entry.cc
cdleary: They're in the .cc file comments should be on the declaration in the .h

xls/dslx/type_system/scoped_fn_stack_entry.cc
cdleary: Ditto. Already present on the .h so this was redundant and in the wrong spot.

richmckeever (approved):

Issue #1591, opened 2024-09-05
Storing dslx::Pos with filename strings is pretty expensive


hzeller:
Currently, we're storing the full filename as a string in [`dslx::Pos`](https://github.com/google/xls/blob/7eb7d07b756df405398755734ff773c0061d881a/xls/dslx/frontend/pos.h#L89) (and two `dslx::Pos` in a [`dslx::Span`](https://github.com/google/xls/blob/7eb7d07b756df405398755734ff773c0061d881a/xls/dslx/frontend/pos.h#L156-L157)).

This is a lot of unnecessary memory that has to be moved around all over.

A quick test running one of the code examples

```
bazel-bin/xls/dslx/interpreter_main xls/dslx/stdlib/apfloat.x
```

This churns through about 6'000'000 Pos objects with > 100MiB of strings. Just for `dslx::Pos`.

In addition, all `dxls::Token`s [contain a string](https://github.com/google/xls/blob/7eb7d07b756df405398755734ff773c0061d881a/xls/dslx/frontend/token.h#L188), instead of a `std::string_view into` the original file. Which means it looses all context and requires for the token to store an additional span and pass around that as expensive tuple.

## To Improve

I suggest to have an object that represents a file, a tuple `(filename, content, content_offset_to_linenumber_index)`, and hold on to this for lifetime of a dslx parse. Now, everything becomes much simpler:

Tokens can just be string-views into the original content, Positions and Spans are not needed anymore and can be implicitly extracted from the `std::string_views` and can be mapped back to filename, line, column when needed on error reporting or similar (by consulting the one `(filename, content, index)` tuple) without having to keep one or multiple `(filename, line, column)` tuple at every token, Node, etc. 

In addition, the formatter can be simplified a lot, as it does not have to keep comments 'on the side', but can just point to the place where they actually are.

This will help dramatically to reduce unnecessary memory churn.

Category: cleanup and simplification. I don't have time to work on this currently.
If someone wants to work on it, I have some detailed thoughts and working framework that can be used as an example.

hongted:
Or in general, consider arena allocation for strings and other objects like nodes that are pseudo-static and used often.

cdleary:
I can for sure take on making a file table since this is my tech debt.

[...]

PR #1594, opened 2024-09-09
[DSLX] Plumb FileTable as a broad-scope context-like object

cdleary:
Fixes primary issue described in #1591 

Thanks in advance for your help to drive this in quickly, or it’s gonna rot "quicker than leftovers that I swore I’d eat".

- This ends up plumbing a context-like object (either `FileTable`, or a thing that can give transitive access to a `FileTable`) into a bunch of spots.
- It’s good to dependency inject a context object like this, or you end up with with confusing duplication (or assumption you can be the origin point) of `FileTable`s at various levels of the hierarchy. It’s annoying to make the change plumb it around because it’s lots of lines, but that’s why I just ground through this PR as fast as I could. :-)
- Notably some things (like the proto to DSLX helpers) didn’t really a have a notion of a global context dependency injected from the outside world, so they just create a FileTable for their internal purposes.
- The most meaningful “tradeoff oriented” decision in this PR was that `Pos` just has an index into a `FileTable`, and no pointer to a `FileTable`. This causes us to dependency inject a file table into various spots where we want to stringify filenames; however, this seemed better than “spooky action at a distance” for a borrowed reference to a `FileTable` coming from lots of `Pos` objects sprinkled around the universe. Since `FileTable` is effectively part of the global context (i.e. it hangs off of `ImportData`) it should always be available when we want to stringify `Pos`/`Span` objects.
    - It’s not hard to change this if people felt strongly, because it’s done the “more explicit / harder way” already in this PR and that alternative is just “more implicit”, but I would recommend that it be done in a followon PR, or maybe only if we determined “yeah, it’s actually annoying to get access to file tables to stringify `Pos`/`Span` objects”.

I made a small LD_PRELOAD interposer library to count malloc calls since it was hard to get tcmalloc working as a bazel dependency for HEAPPROFILE, and this is the effect:

before
after


xls/dslx/bytecode/bytecode_interpreter.cc
```diff
@@ -149,7 +150,9 @@ constexpr int64_t kChannelTraceIndentation = 2;
 
 BytecodeInterpreter::BytecodeInterpreter(
     ImportData* import_data, const BytecodeInterpreterOptions& options)
-    : import_data_(import_data), options_(options) {}
+    : import_data_(import_data),
```
hzeller, line 153: Since we're now dereferencing the line below, I suggest to make this `import_data_(ABSL_DIE_IF_NULL(import_data))`

 ... which then will find two places in `bytecode_interpreter_test.cc` that hit this issue.

(Don't worry for now, will do that while importing)

cdleary: Thanks Henner!

(I guess it's just not exercised? I ran on my machine xls/... with `--config=asan` and it was clean, I imagined it'd show if a null deref happened.)

hzeller: Look for `/*import_data=*/nullptr` in `xls/dslx/bytecode/bytecode_interpreter_test.cc`. It might work by accident as nothing is called on it.

cdleary: Ok let me know if you need me to revise, sounded like you were planning to fix it on import but LMK if I misinterpreted.

hzeller: Yep, already fixed that in my local import copy.
Just in the process to get everything else outside xls lined up. To be submitted soon.

hzeller:
Alright, merged.
If there are little things here and there that need updating/improved/cleaned/prettified later, it is now possible to iterate on the main repo.

PR #1606, opened 2024-09-14
[DSLX] Fixes for casting to/from `xN` (bits constructor) based type, adds docs and more tests

### LINKED ISSUE #1572: Parametric signedness bit type doesn't work
**Describe the bug**
The bytecode emitter errors out if you try to use a return type like `xN[RESULT_SIGNED][RESULT_SZ]`.

```
: INVALID_ARGUMENT: Provided status is not in recognized error form: INTERNAL: Bits-to-array cast bit counts must match. Saw 32 vs 0.; while evaluating: to_int<u32:8, u32:23, u32:32>(APFloat<u32:8, u32:23> { sign: u1:0, bexp: u8:0x0, fraction: u23:0x0 })
...
Error: INTERNAL: Bits-to-array cast bit counts must match. Saw 32 vs 0.; while evaluating: ...
=== Source Location Trace: ===
```

**To Reproduce**
Change this line:

https://github.com/google/xls/blob/69f00f93eb09c0912185ac8a95b8075cb62a7463/xls/dslx/stdlib/apfloat.x#L1301

to `xN[RESULT_SIGNED][RESULT_SZ]` (updating the casts at https://github.com/google/xls/blob/69f00f93eb09c0912185ac8a95b8075cb62a7463/xls/dslx/stdlib/apfloat.x#L1352 and https://github.com/google/xls/blob/69f00f93eb09c0912185ac8a95b8075cb62a7463/xls/dslx/stdlib/apfloat.x#L1355)

**Expected behavior**
This should properly include parameterized signedness and remove the need to cast the result in the wrapper at https://github.com/google/xls/blob/69f00f93eb09c0912185ac8a95b8075cb62a7463/xls/dslx/stdlib/apfloat.x#L1370

cdleary:
Fixes #1572 

Development details -- note that `xN` is the first place where the concept of "BitsLike" became load bearing, any place that assumed `ArrayType` in the hierarchy now has to consider that `ArrayType` could be indicating something that's effectively a bits type; i.e. `BitsLike`. Going forward we always want to use `BitsLike` as the test for a type and whether it's effectively a bits type, there were a few spots here that were missed when the `xN` bits type constructor was introduced.

### DIFF
--- a/xls/dslx/bytecode/bytecode_interpreter.cc
+++ b/xls/dslx/bytecode/bytecode_interpreter.cc
@@ -83,11 +83,15 @@ absl::StatusOr<std::string> ToStringMaybeFormatted(
 // Casts an InterpValue representable as Bits to a new InterpValue
 // with the given BitsType.
 absl::StatusOr<InterpValue> ResizeBitsValue(const InterpValue& from,
-                                            BitsType* to_bits, bool is_checked,
+                                            const BitsLikeProperties& to_bits,
+                                            const Type& to_type,
+                                            bool is_checked,
                                             const Span& source_span,
                                             const FileTable& file_table) {
-  XLS_ASSIGN_OR_RETURN(int64_t to_bit_count,
-                       to_bits->GetTotalBitCount().value().GetAsInt64());
+  VLOG(3) << "ResizeBitsValue; from: " << from << " to: " << to_type << " @ "
+          << source_span.ToString(file_table);
+  XLS_ASSIGN_OR_RETURN(int64_t to_bit_count, to_bits.size.GetAsInt64());
+  XLS_ASSIGN_OR_RETURN(bool is_signed, to_bits.is_signed.GetAsBool());
 
   // Check if it fits.
   // Case A: to unsigned of N-bits
@@ -97,16 +101,16 @@ absl::StatusOr<InterpValue> ResizeBitsValue(const InterpValue& from,
   if (is_checked) {
     bool does_fit = false;
 
-    if (!to_bits->is_signed()) {
+    if (!is_signed) {
       does_fit = !from.IsNegative() && from.FitsInNBitsUnsigned(to_bit_count);
-    } else if (to_bits->is_signed() && !from.IsNegative()) {
+    } else if (is_signed && !from.IsNegative()) {
       does_fit = from.FitsInNBitsUnsigned(to_bit_count - 1);
     } else {  // to_bits->is_signed() && from.IsNegative()
       does_fit = from.FitsInNBitsSigned(to_bit_count);
     }
 
     if (!does_fit) {
-      return CheckedCastErrorStatus(source_span, from, to_bits, file_table);
+      return CheckedCastErrorStatus(source_span, from, &to_type, file_table);
     }
   }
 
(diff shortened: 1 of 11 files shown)


xls/dslx/stdlib/apfloat.x
```diff
@@ -1481,14 +1481,15 @@ fn to_signed_or_unsigned_int<RESULT_SZ: u32, RESULT_SIGNED: bool, EXP_SZ: u32, F
         }
     };
 
+    type ReturnT = xN[RESULT_SIGNED][RESULT_SZ];
```
mikex-oss, line 1484: I think the result is already the correct signedness.

cdleary: I thought this was the change you were trying to make in #1572? Maybe it was just for demonstration purposes -- I can revert this.

mikex-oss: Sorry for the confusion. I was trying to remove the cast on line 1507.

mikex-oss: The RESULT_SIGNED branch now returns sN and the other branch returns uN, which is made possible by the xN change. Thanks btw!

cdleary: No worries! Reverted the apfloat change here. In retrospect, separating that out probably makes sense regardless. I think this'll probably fix the root cause though. 👍

hzeller:
@mikex-oss can you LGTM and tag reviewing internally when ready here ?

mikex-oss:
Hey @richmckeever this looks reasonable to me, can you take a look and approve?

xls/dslx/stdlib/apfloat.x
mikex-oss: Thanks, I'll fix up apfloat as a follow up.

xls/dslx/bytecode/bytecode_interpreter.cc
```diff
@@ -602,79 +606,95 @@ absl::Status BytecodeInterpreter::EvalCast(const Bytecode& bytecode,
...
     return absl::InvalidArgumentError(
         "Bytecode interpreter only supports casts from arrays, enums, and "
         "bits; got: " +
-        from.ToString());
+        from_value.ToString());
+  }
+
+  int64_t from_bit_count = from_value.GetBits().value().bit_count();
+
+  // If the thing we're casing from is bits like, and the thing we're casting
```
richmckeever, line 658: Nit: "casting"

richmckeever (approved):

xls/dslx/bytecode/bytecode_interpreter.cc
cdleary: Done.

xls/dslx/stdlib/apfloat.x
mikex-oss: @cdleary where we probably got crossed is that I forgot DSLX doesn't have constexpr if's, so the types in both branches still need to match, while you probably knew that :)

My initial comment was thinking that we could write:

```
    if RESULT_SIGNED {
        // Reduce to the target size, preserving signedness.
        let result = result as sN[RESULT_SZ];
        if !x.sign { result } else { -result }
    } else {
        // Already unsigned, just size correctly.
        result as uN[RESULT_SZ]
    }
```

Anyway, submitted basically what you had with the additional cast in `to_int` removed in https://github.com/google/xls/commit/17d0d09be1f0c895adfd83478e91e728def009b9.

cdleary: You're giving me too much credit, was kind of blindly trying to do what you had described without thinking much. :-) Glad you were able to make the change based off this one though!

PR #1612, opened 2024-09-18
[DSLX] Add support for `MIN` attribute on builtin bits types.

### LINKED ISSUE #1556: [enhancement] Add `MIN` constant

DSLX has `MAX` and `ZERO` constants, but not `MIN`. https://google.github.io/xls/dslx_reference/#bit-type-attributes

`MIN` seems way more helpful than `ZERO`, given I think you can just write `s32:0` or `u32:0` in less characters than with the constant but need to do a bit of math for the min signed value.

Bit shift a 1 for signed ints.

Provide `MIN` constant.

cdleary:
In similar fashion to how `ZERO` and `MAX` already existed.

Fixes #1556

### DIFF
--- a/docs_src/dslx_reference.md
+++ b/docs_src/dslx_reference.md
@@ -323,6 +323,9 @@ s3::MAX   // s3:0b011 the "maximum signed" value
 
 u3::ZERO  // u3:0b000 the "fill with zeros" value
 s3::ZERO  // s3:0b000 the "fill with zeros" value
+
+u3::MIN   // u3:0b000 the minimum u3 value
+s3::MIN   // s3:0b100 AKA s3:-4 the minimum s3 value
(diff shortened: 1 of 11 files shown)


mikex-oss:
Thanks! Mind also updating this documentation?

https://github.com/google/xls/blob/a593591f61484b6c9da807f0be9aa41f72ab2a90/docs_src/floating_point.md?plain=1#L382-L388

https://github.com/google/xls/blob/a593591f61484b6c9da807f0be9aa41f72ab2a90/xls/dslx/stdlib/apfloat.x#L1497-L1503

cdleary:
> Thanks! Mind also updating this documentation?

Done.

mikex-oss (approved):

PR #1613, opened 2024-09-18
Add procs for interfacing with AXI peripherals in ZSTD Decoder

rw1nkler:
This PR introduces pros that enable the interface between DSLX modules and AXI subordinates. The attached [README](https://github.com/google/xls/blob/72e55dd291236fa28da3c96d765317e03b9ce6c5/xls/modules/zstd/memory/README.md) provides detailed documentation of the new functionality. Provided Verilog simulations show that the new procs allow for reading from and writing to RAM located on the AXI bus.

While working on this PR, we encountered a potential issue linked to afc720d6a35616c894b6aa342c49554afe0f35cd, which appears to cause indefinite parsing and type-checking loops for certain files, such as `mem_writer.x` (provided in this PR).

### DIFF
--- a/xls/modules/zstd/memory/BUILD
+++ b/xls/modules/zstd/memory/BUILD
@@ -32,3 +32,17 @@ xls_dslx_library(
     name = "axi_st_dslx",
     srcs = ["axi_st.x"],
 )
+
+xls_dslx_library(
+    name = "common_dslx",
+    srcs = [
+        "common.x",
+    ],
+    deps = [":axi_dslx"],
+)
+
+xls_dslx_test(
+    name = "common_dslx_test",
+    library = ":common_dslx",
+    tags = ["manual"],
+)
--- a/xls/modules/zstd/memory/BUILD
+++ b/xls/modules/zstd/memory/BUILD
@@ -224,3 +224,77 @@ place_and_route(
     tags = ["manual"],
     target_die_utilization_percentage = "10",
 )
+
+xls_dslx_library(
+    name = "axi_stream_downscaler_dslx",
+    srcs = ["axi_stream_downscaler.x"],
+    deps = [
+        ":axi_st_dslx",
+    ],
+)
+
+xls_dslx_test(
+    name = "axi_stream_downscaler_dslx_test",
+    library = ":axi_stream_downscaler_dslx",
+    tags = ["manual"],
+)
+
+axi_stream_downscaler_codegen_args = common_codegen_args | {
+    "module_name": "axi_stream_downscaler",
+    "pipeline_stages": "2",
+}
+
+xls_dslx_verilog(
+    name = "axi_stream_downscaler_verilog",
+    codegen_args = axi_stream_downscaler_codegen_args,
+    dslx_top = "AxiStreamDownscalerInst",
+    library = ":axi_stream_downscaler_dslx",
+    tags = ["manual"],
+    verilog_file = "axi_stream_downscaler.v",
+)
+
+xls_benchmark_ir(
+    name = "axi_stream_downscaler_opt_ir_benchmark",
+    src = ":axi_stream_downscaler_verilog.opt.ir",
+    benchmark_ir_args = axi_stream_downscaler_codegen_args | {
+        "pipeline_stages": "10",
+        "top": "__axi_stream_downscaler__AxiStreamDownscalerInst__AxiStreamDownscaler_0__8_8_128_16_32_4_4_3_next",
+    },
+    tags = ["manual"],
+)
+
+verilog_library(
+    name = "axi_stream_downscaler_verilog_lib",
+    srcs = [
+        ":axi_stream_downscaler.v",
+    ],
+    tags = ["manual"],
+)
+
+synthesize_rtl(
+    name = "axi_stream_downscaler_synth_asap7",
+    standard_cells = "@org_theopenroadproject_asap7sc7p5t_28//:asap7-sc7p5t_rev28_rvt",
+    tags = ["manual"],
+    top_module = "axi_stream_downscaler",
+    deps = [
+        ":axi_stream_downscaler_verilog_lib",
+    ],
+)
+
+benchmark_synth(
+    name = "axi_stream_downscaler_benchmark_synth",
+    synth_target = ":axi_stream_downscaler_synth_asap7",
+    tags = ["manual"],
+)
+
+place_and_route(
+    name = "axi_stream_downscaler_place_and_route",
+    clock_period = CLOCK_PERIOD_PS,
+    core_padding_microns = 2,
+    min_pin_distance = "0.5",
+    placement_density = "0.30",
+    stop_after_step = "global_routing",
+    synthesized_rtl = ":axi_stream_downscaler_synth_asap7",
+    tags = ["manual"],
+    target_die_utilization_percentage = "10",
+)
--- a/xls/modules/zstd/memory/BUILD
+++ b/xls/modules/zstd/memory/BUILD
@@ -563,3 +563,80 @@ place_and_route(
     tags = ["manual"],
     target_die_utilization_percentage = "10",
 )
+
+xls_dslx_library(
+    name = "axi_stream_add_empty_dslx",
+    srcs = ["axi_stream_add_empty.x"],
+    deps = [
+        ":axi_dslx",
+        ":axi_st_dslx",
+        ":axi_writer_dslx",
+        ":common_dslx",
+    ],
+)
+
+xls_dslx_test(
+    name = "axi_stream_add_empty_dslx_test",
+    library = ":axi_stream_add_empty_dslx",
+)
+
+axi_stream_add_empty_codegen_args = common_codegen_args | {
+    "module_name": "axi_stream_add_empty",
+    "pipeline_stages": "2",
+    "streaming_channel_data_suffix": "_data",
+}
+
+xls_dslx_verilog(
+    name = "axi_stream_add_empty_verilog",
+    codegen_args = axi_stream_add_empty_codegen_args,
+    dslx_top = "AxiStreamAddEmptyInst",
+    library = ":axi_stream_add_empty_dslx",
+    tags = ["manual"],
+    verilog_file = "axi_stream_add_empty.v",
+)
+
+xls_benchmark_ir(
+    name = "axi_stream_add_empty_opt_ir_benchmark",
+    src = ":axi_stream_add_empty_verilog.opt.ir",
+    benchmark_ir_args = axi_stream_add_empty_codegen_args | {
+        "pipeline_stages": "10",
+        "top": "__axi_stream_add_empty__AxiStreamAddEmptyInst__AxiStreamAddEmpty_0__16_32_4_2_32_32_next",
+    },
+    tags = ["manual"],
+)
+
+verilog_library(
+    name = "axi_stream_add_empty_verilog_lib",
+    srcs = [
+        ":axi_stream_add_empty.v",
+    ],
+    tags = ["manual"],
+)
+
+synthesize_rtl(
+    name = "axi_stream_add_empty_synth_asap7",
+    standard_cells = "@org_theopenroadproject_asap7sc7p5t_28//:asap7-sc7p5t_rev28_rvt",
+    tags = ["manual"],
+    top_module = "axi_stream_add_empty",
+    deps = [
+        ":axi_stream_add_empty_verilog_lib",
+    ],
+)
+
+benchmark_synth(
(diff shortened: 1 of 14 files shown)


lpawelcz:
Moved the cocotb tests to a separate PR: https://github.com/google/xls/pull/1616

lpawelcz:
Added performance improvements for the `MemWriter` proc.

The PR is ready for review, @proppy please take a look.

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,675 @@
...
+        ":axi_dslx",
+        ":axi_st_dslx",
+        ":common_dslx",
+    ],
+)
+
+xls_dslx_test(
+    name = "axi_reader_dslx_test",
+    library = ":axi_reader_dslx",
+    tags = ["manual"],
```
proppy, line 83: any reason we leave those as manual?

proppy: oups, I forgot we used this to filter them out of the main ci.

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,675 @@
...
+    "clock_period_ps": "1800",
+}
+
+xls_dslx_verilog(
+    name = "axi_reader_verilog",
+    codegen_args = axi_reader_codegen_args,
+    dslx_top = "AxiReaderInst",
+    library = ":axi_reader_dslx",
+    opt_ir_args = {
+        "top": "__axi_reader__AxiReaderInst__AxiReader_0__16_32_4_4_4_3_2_14_next",
```
proppy, line 101: is that still required?

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,675 @@
...
+    tags = ["manual"],
+)
+
+axi_reader_codegen_args = common_codegen_args | {
+    "module_name": "axi_reader",
+    "pipeline_stages": "2",
+    "streaming_channel_data_suffix": "_data",
+    "flop_inputs_kind": "skid",
+    "flop_outputs_kind": "skid",
+    "clock_period_ps": "1800",
```
proppy, line 92: any reason the clock period is different here? should we keep a list of those at the top for easier tuning?

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,675 @@
...
+
+xls_dslx_test(
+    name = "axi_stream_remove_empty_dslx_test",
+    library = ":axi_stream_remove_empty_dslx",
+    tags = ["manual"],
+)
+
+axi_stream_remove_empty_codegen_args = common_codegen_args | {
+    "module_name": "axi_stream_remove_empty",
+    "clock_period_ps": "1300",
```
proppy, line 168: any reason the clock period is different here? should we keep a list of those at the top for easier tuning?

xls/modules/zstd/memory/README.md
```diff
@@ -0,0 +1,103 @@
...
+   should be send to `data_in_r` channel. The structure consists of
+
+   - the actual `data`
+   - `length` of the `data` in bytes
+   - `last` flag used for marking the last `packet` with data.
+
+4. Wait for the response submitted on the `resp_s` channel, which indicates
+   if the write operation was successful or an error occurred.
+
+# Cocotb Simulation
```
proppy, line 99: can we move this section to #1616?

proppy:
The tests take quite a long time to execute:
```
[4,813 / 4,816] 18 / 21 tests; Parsing and type checking DSLX source files of target axi_writer_dslx; 2688s linux-sandbox ... (3 actions running)
```
is that expected?

rw1nkler:
This is the problem 

> The tests take quite a long time to execute:
> 
> ```
> [4,813 / 4,816] 18 / 21 tests; Parsing and type checking DSLX source files of target axi_writer_dslx; 2688s linux-sandbox ... (3 actions running)
> ```
> 
> is that expected?

This is the problem described in https://github.com/google/xls/issues/1615

proppy:
according to https://github.com/google/xls/issues/1615#issuecomment-2385046454, does this need to be updated?

xls/modules/zstd/memory/BUILD
lpawelcz: The critical path delay from the IR benchmark for this proc exceeds the 750ps that we use throughout the ZSTD Decoder. In order to provide valid constraints for the scheduler we increased the clock period to the value that is close to the critical path delay. Moved this to a variable at the top of the file where the `750ps` clock period is also defined.

xls/modules/zstd/memory/BUILD
lpawelcz: The critical path delay from the IR benchmark for this proc exceeds the 750ps that we use throughout the ZSTD Decoder. In order to provide valid constraints for the scheduler we increased the clock period to the value that is close to the critical path delay. Moved this to a variable at the top of the file where the `750ps` clock period is also defined.

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,675 @@
...
+    target_die_utilization_percentage = "10",
+)
+
+mem_reader_codegen_args = common_codegen_args | {
+    "module_name": "mem_reader",
+    "pipeline_stages": "4",
+    "streaming_channel_data_suffix": "_data",
+    "flop_inputs_kind": "skid",
+    "flop_outputs_kind": "skid",
+    "clock_period_ps": "2600",
```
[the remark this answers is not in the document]
lpawelcz: The critical path delay from the IR benchmark for this proc exceeds the 750ps that we use throughout the ZSTD Decoder. In order to provide valid constraints for the scheduler we increased the clock period to the value that is close to the critical path delay. Moved this to a variable at the top of the file where the `750ps` clock period is also defined.

xls/modules/zstd/memory/README.md
lpawelcz: Done

lpawelcz:
@proppy 
We rebased the PR and fixed the codebase to according to https://github.com/google/xls/issues/1615#issuecomment-2383843307.

After the rebase we experienced a regression in `place_and_route` rule for the `SequenceExecutor`, hence the https://github.com/google/xls/pull/1613/commits/9282bfb47b7ed6504e39b85137c202252cf7c45c.

Additionally, we noticed that `bazel run` calls in https://github.com/google/xls/blob/main/.github/workflows/modules-zstd.yml#L63 execute only the first bazel target from the `bazel query` output.
In https://github.com/google/xls/pull/1613/commits/94a6d6dd9495af73c4682f5987cc1acc48611eaf we fixed this by explicitly looping through all the targets and calling `bazel run` for one target at a time.

We addressed your review comments and slightly changed the interface of the `MemReader/Writer` by removing the `Ctrl` channel that was used to configure the procs with a base address. Now the requests to the procs contain absolute addresses instead of offsets from the base address.

proppy:
> After the rebase we experienced a regression in place_and_route rule for the SequenceExecutor, hence the 9282bfb.

did you report it to https://github.com/hdl/bazel_rules_hdl?

proppy:
> Additionally, we noticed that bazel run calls in https://github.com/google/xls/blob/main/.github/workflows/modules-zstd.yml#L63 execute only the first bazel target from the bazel query output.

That sounds weird, can you share the shell-expanded bazel command line that only run the first target?

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,714 @@
...
+)
+
+xls_dslx_test(
+    name = "common_dslx_test",
+    library = ":common_dslx",
+    tags = ["manual"],
+)
+
+CLOCK_PERIOD_PS = "750"
+# Clock periods for modules that exceed the 750ps critical path in IR benchmark
```
proppy, line 58: would increasing pipelining help?

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,714 @@
...
+    tags = ["manual"],
+    verilog_file = "axi_reader.v",
+)
+
+xls_benchmark_ir(
+    name = "axi_reader_opt_ir_benchmark",
+    src = ":axi_reader_verilog.opt.ir",
+    benchmark_ir_args = axi_reader_codegen_args | {
+        "pipeline_stages": "10",
+        "top": "__axi_reader__AxiReaderInst__AxiReader_0__16_32_4_4_4_3_2_14_next",
```
proppy, line 112: maybe use https://google.github.io/xls/bazel_rules_macros/#get_mangled_ir_symbol?

lpawelcz:
> > Additionally, we noticed that bazel run calls in https://github.com/google/xls/blob/main/.github/workflows/modules-zstd.yml#L63 execute only the first bazel target from the bazel query output.
> 
> That sounds weird, can you share the shell-expanded bazel command line that only run the first target?

That would be for example:
```
bazel run -c opt -- //xls/modules/zstd:axi_csr_accessor_opt_ir_benchmark //xls/modules/zstd:block_header_dec_opt_ir_benchmark //xls/modules/zstd:csr_config_opt_ir_benchmark //xls/modules/zstd:dec_mux_opt_ir_benchmark //xls/modules/zstd:frame_header_dec_opt_ir_benchmark //xls/modules/zstd:raw_block_dec_opt_ir_benchmark //xls/modules/zstd:repacketizer_opt_ir_benchmark //xls/modules/zstd:rle_block_dec_opt_ir_benchmark //xls/modules/zstd:sequence_executor_opt_ir_benchmark //xls/modules/zstd:window_buffer_opt_ir_benchmark //xls/modules/zstd:zstd_dec_internal_opt_ir_benchmark //xls/modules/zstd/memory:axi_reader_opt_ir_benchmark //xls/modules/zstd/memory:axi_stream_add_empty_opt_ir_benchmark //xls/modules/zstd/memory:axi_stream_downscaler_opt_ir_benchmark //xls/modules/zstd/memory:axi_stream_remove_empty_opt_ir_benchmark //xls/modules/zstd/memory:axi_writer_opt_ir_benchmark //xls/modules/zstd/memory:mem_reader_internal_opt_ir_benchmark
```

This will execute the IR benchmark only for the `axi_csr_accessor` proc.

xls/modules/zstd/memory/BUILD
lpawelcz: This is not enough for the scheduler as we constrain it from 3 sides:
* worst case throughput - 1
* target clock period - 750ps
* pipeline stages - as many as required

Combining `worst_case_throughput==1` with fixed clock period requirement enforces a condition that each `next()` evaluation must be possible to compute in a given clock period time limit.

Currently, some of the procs don't meet this requirement and scheduling in such case will fail with a message to increase the `worst_case_throughput`. We can do that or we can stop enforcing specific clock period for such procs. This way, the module would still evaluate the next() function in a single clock cycle and we would still have an estimate for the max clock period from the IR benchmark.

xls/modules/zstd/memory/BUILD
proppy: ah you're right:
>       is_proc_next: A boolean flag denoting whether the symbol is a
>         next proc function. The argument is mutually exclusive with arguments:
>        'parametric_values' and 'is_implicit_token'.

We should file a feature request for this to be supported, one workaround I can see is to define wrapper proc rather than putting the parameters in the BUILD file.

proppy:
> This will execute the IR benchmark only for the axi_csr_accessor proc.

Looks like this is by design:
https://bazel.build/docs/user-manual#running-executables

> The bazel run command is similar to bazel build, except it is used to build and run a single target.

proppy (approved):

xls/modules/zstd/memory/BUILD
```diff
@@ -0,0 +1,714 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_hdl//place_and_route:build_defs.bzl", "place_and_route")
+load("@rules_hdl//synthesis:build_defs.bzl", "benchmark_synth", "synthesize_rtl")
+load("@rules_hdl//verilog:providers.bzl", "verilog_library")
+load("@xls_pip_deps//:requirements.bzl", "requirement")
```
hzeller, line 18: This looks like an unused import.

Can you run

```
buildifier --lint=fix xls/modules/zstd/memory/BUILD
```

on it ? (Sorry, we should have built into the CI)

hzeller: Fixing this now on import, but it is always a good idea to run buildifier before.

Issue #1615, opened 2024-09-19
Indefinite parsing and type-checking loops for some parameterized procs


lpawelcz:
**Describe the bug**
While working on https://github.com/google/xls/pull/1613, we encountered an issue linked to https://github.com/google/xls/commit/afc720d6a35616c894b6aa342c49554afe0f35cd, which appears to cause indefinite parsing and type-checking loops for certain files, such as `mem_writer.x` (provided in this PR).

**To Reproduce**
1. Checkout on the [branch](https://github.com/antmicro/xls/tree/axi) from https://github.com/google/xls/pull/1613
2. Execute the command `bazel build -c opt //xls/modules/zstd/memory:mem_writer_verilog`
3. Observe indefinite loop in parsing and type-checking `mem_writer` DSLX files:
```
    Converting DSLX file to XLS IR: xls/modules/zstd/memory/mem_writer.x; 48s remote-cache, processwrapper-sandbox
    Parsing and type checking DSLX source files of target mem_writer_dslx; 48s remote-cache, processwrapper-sandbox
```
4. Cancel bazel build
5. Revert https://github.com/google/xls/commit/afc720d6a35616c894b6aa342c49554afe0f35cd
6. Repeat the bazel build command: `bazel build -c opt //xls/modules/zstd/memory:mem_writer_verilog`
7. Observe correct IR and verilog generation for the `MemWriter` proc.

**Expected behavior**
Toolchain should be able to generate IR for the provided `MemWriter` proc .

**Additional context**
This is a blocker for https://github.com/google/xls/pull/1613 and as a consequence also for https://github.com/google/xls/issues/1211.

richmckeever:
What's happening here is that a couple of bugs, which are inconsistent sequencing of the parametric pattern "AxiStream<DATA_W, DEST_W, ID_W, DATA_W_DIV8>" in the zstd code, are exposing parametric env leakage in type checking, where we build maps like:

DEST_W => DATA_W_DIV8
DATA_W => DATA_W
DATA_W_DIV8 => ID_W
ID_W => DEST_W

and then try to figure out what e.g. DATA_W_DIV8 is, using such a map.

The faulty "AxiStream<wrong-order parametrics>" are in axi_writer.x and mem_writer.x. You may be able to get unblocked by fixing those to be consistent with the struct declaration. In the meantime I will work on a fix in the type checker.

lpawelcz:
[...]

PR #1616, opened 2024-09-20
Add cocotb tests for AXI peripherals

lpawelcz:
This PR extends https://github.com/google/xls/pull/1613 by adding Verilog simulation and tests for `MemReader`, `AxiWriter` and `MemWriter` procs. The simulation and testing is carried out with [cocotb](https://github.com/cocotb/cocotb) framework and [Icarus Verilog](https://github.com/steveicarus/iverilog) simulator. The PR also includes utilities for driving and monitoring DSLX channels by the cocotb framework.

Cocotb and the provided utilities are Python packages and don't require any special rules in the workspace. The utilities we created are generic, and although they were placed in the `xls/modules/zstd/cocotb` directory, they can be moved to another location if this is more desired.

Please note that this code is based on top of https://github.com/google/xls/pull/1613 so the diff also contains all the changes from that PR. The IR generation for the `MemWriter` proc is also blocked by https://github.com/google/xls/issues/1615.

### DIFF
--- a/.github/workflows/modules-zstd.yml
+++ b/.github/workflows/modules-zstd.yml
@@ -75,6 +75,11 @@ jobs:
             bazel run -c opt $target -- --logtostderr;
           done
 
+      - name: Build and test verilog simulation of the ZSTD module components (opt)
+        if: ${{ !cancelled() }}
+        run: |
+          bazel test -c opt --test_output=errors -- $(bazel query 'filter(".*_cocotb_test", kind(rule, //xls/modules/zstd/...))')
+
       - name: Build ZSTD place and route targets (opt)
         if: ${{ !cancelled() }}
         run: |
--- a/dependency_support/pip_requirements.in
+++ b/dependency_support/pip_requirements.in
@@ -8,6 +8,10 @@ termcolor==1.1.0
 psutil==5.7.0
 portpicker==1.3.1
 pyyaml==6.0.1
+pytest==8.2.2
+cocotb==1.9.0
+cocotbext-axi==0.1.24
+cocotb_bus==0.2.1
 
 # Note: numpy and scipy version availability seems to differ between Ubuntu
 # versions that we want to support (e.g. 18.04 vs 20.04), so we accept a
--- a/dependency_support/pip_requirements_lock.txt
+++ b/dependency_support/pip_requirements_lock.txt
@@ -14,10 +14,68 @@ click==8.1.3 \
     # via
     #   -r dependency_support/pip_requirements.in
     #   flask
+cocotb==1.9.0 \
+    --hash=sha256:02a58ef6c941114964096e7c039bdd4e67e63816cfd2f6a9af6a34cd92b00e8e \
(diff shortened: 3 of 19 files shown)


lpawelcz:
CC @proppy - please take a look at this PR

proppy:
Does this replace #1057?

rw1nkler:
Regarding the content, this PR uses the ideas from #1057 but provides more helpers for writing the simulations. 
The previous PR was adding cocotb to a different directory, for use in other places in the workspace. Here we provide the scripts directly in the `modules/zstd` directory.  We can discard the previous PR, if you think that it is easier to merge the cocotb support with the AXI tests, or we can move the created libraries to #1057 and merge them separately.

proppy:
> We can discard the previous PR, if you think that it is easier to merge the cocotb support with the AXI tests

I think it'll be easier to maintain it along side the zstd modules for now, and we can generalize it to the wider xls project as it matures.

lpawelcz:
@proppy we rebased the PR onto current state of the https://github.com/google/xls/pull/1613 and updated cocotb testbenches to take into account the changes made to `MemReader` and `MemWriter`

proppy:
rebase?

lpawelcz:
Rebased onto newest main

CC @proppy

.github/workflows/modules-zstd.yml
```diff
@@ -75,6 +75,16 @@ jobs:
             bazel run -c opt $target -- --logtostderr;
           done
 
+      - name: Build and run verilog simulation of the ZSTD module components (opt)
+        if: ${{ !cancelled() }}
+        run: |
+          for target in $(bazel query 'filter(".*_cocotb_test$", kind(rule, //xls/modules/zstd/...))');
+          do
+            echo "running $target";
+            bazel run -c opt $target;
```
proppy, line 84: should we use bazel test here, that would also allow us to get rid of the for loop.

lpawelcz: Rebased and changed the bazel call from `bazel run` to `bazel test`. This way we indeed got rid of the for loop.
We can still run those targets with `bazel run` if need be.

lpawelcz:
CC @proppy 

The last CI failure was caused by `benchmark_main` [segmentation fault](https://github.com/google/xls/actions/runs/11494086318/job/32320996820?pr=1616#step:10:463) in `DecoderMux` proc IR benchmark.
This is an unrelated issue as we also experienced the same failure on current `main`. It appears that this particular IR benchmark started failing randomly.
We are investigating the root cause of the problem. We noticed that the failure happens when [the Block IR is evaluated with JIT](https://github.com/google/xls/blob/main/xls/dev_tools/benchmark_main.cc#L655). The evaluation is carried out with a random set of inputs which explains the unpredictable nature of the failures.
The problem might be caused by the usage of `assert!()` inside the proc but for now we were unable to prepare a minimal example for that.

lpawelcz:
@proppy closing this PR as it was superseded by #1857

PR #1621, opened 2024-09-22
[c_api] Break C API monolith into multiple sub-targets.

cdleary:
Since VAST is now in there (as a conceptually "protected" / more prone to change API) with the other public APIs seemed like a good time to break it into some translation units before introducing more "protected" APIs e.g. for DSLX inspection.

### DIFF
--- a/xls/public/c_api_impl_helpers.cc
+++ b/xls/public/c_api_impl_helpers.cc
@@ -1,0 +1,74 @@
+#include "xls/public/c_api_impl_helpers.h"
+
+#include <filesystem>
+#include <vector>
+
+#include "absl/log/check.h"
+
+namespace xls {
+
+std::vector<std::filesystem::path> ToCpp(const char* additional_search_paths[],
+                                         size_t additional_search_paths_count) {
+  std::vector<std::filesystem::path> additional_search_paths_cpp;
+  additional_search_paths_cpp.reserve(additional_search_paths_count);
+  for (size_t i = 0; i < additional_search_paths_count; ++i) {
+    const char* additional_search_path = additional_search_paths[i];
+    CHECK(additional_search_path != nullptr);
+    additional_search_paths_cpp.push_back(additional_search_path);
+  }
+  return additional_search_paths_cpp;
+}
+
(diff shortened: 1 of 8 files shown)


richmckeever (approved):

xls/public/c_api_impl_helpers.cc
```diff
@@ -0,0 +1,88 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "xls/public/c_api_impl_helpers.h"
+
+#include <filesystem>
```
proppy, line 17: `// NOLINT`?

xls/public/c_api_impl_helpers.h
```diff
@@ -0,0 +1,54 @@
...
+
+// Implementation helper functions for implementing the public C API -- note
+// these are not helpers for /using/ the C API, and so will not need to be
+// included by client code.
+
+#ifndef XLS_PUBLIC_C_API_IMPL_HELPERS_H_
+#define XLS_PUBLIC_C_API_IMPL_HELPERS_H_
+
+#include <cstddef>
+#include <filesystem>
```
proppy, line 23: `// NOLINT`?

xls/public/c_api_impl_helpers.cc
cdleary: Done, though note it's hard to anticipate what the linter will do for contributors, I don't think it's externally available right?

xls/public/c_api_impl_helpers.h
cdleary: Done.

xls/public/c_api_impl_helpers.cc
proppy: > Done, though note it's hard to anticipate what the linter will do for contributors, I don't think it's externally available right?

@hzeller was mentioning it would be nice to add the `NOLINT` annotation on import, or make sure we enforce the check externally (maybe enforcing `xls::common::Filesystem` usage?)

PR #1622, opened 2024-09-23
[c_api] Add start of C API for DSLX, including inspecting module-scoped type definitions.

cdleary:
Draft PR as this is dependent on #1621 landing (after which this will subsequently be rebased).

Note that this uses the multi-translation-unit structured added in the previous PR, adding a new one for DSLX "protected" APIs.

### DIFF
--- a/xls/dslx/frontend/ast.h
+++ b/xls/dslx/frontend/ast.h
@@ -2195,7 +2195,7 @@ class EnumDef : public AstNode {
   NameDef* name_def() const { return name_def_; }
 
   const std::vector<EnumMember>& values() const { return values_; }
-  std::vector<EnumMember>& values() { return values_; }
+  std::vector<EnumMember>& mutable_values() { return values_; }
 
   TypeAnnotation* type_annotation() const { return type_annotation_; }
   bool is_public() const { return is_public_; }
@@ -2274,7 +2274,7 @@ class StructDef : public AstNode {
   }
 
   const std::vector<StructMember>& members() const { return members_; }
-  std::vector<StructMember>& members() { return members_; }
+  std::vector<StructMember>& mutable_members() { return members_; }
 
   bool is_public() const { return public_; }
   const Span& span() const { return span_; }
--- a/xls/dslx/frontend/ast.h
+++ b/xls/dslx/frontend/ast.h
@@ -270,8 +270,7 @@ inline std::vector<AstNode*> ToAstNodes(absl::Span<NodeT* const> source) {
 // Abstract base class for type annotations.
 class TypeAnnotation : public AstNode {
  public:
-  TypeAnnotation(Module* owner, Span span)
-      : AstNode(owner), span_(std::move(span)) {}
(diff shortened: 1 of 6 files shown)


richmckeever (approved):

xls/dslx/frontend/ast.h
```diff
@@ -2194,7 +2193,10 @@ class EnumDef : public AstNode {
   const Span& span() const { return span_; }
   std::optional<Span> GetSpan() const override { return span_; }
   NameDef* name_def() const { return name_def_; }
+
   const std::vector<EnumMember>& values() const { return values_; }
+  std::vector<EnumMember>& values() { return values_; }
```
hzeller, line 2198: I'd suggest to call these mutable accesses `mutable_values()` and below `mutable_members()` as this sounds like it should be an explicit choice by callers.

xls/public/c_api_dslx.cc
```diff
@@ -0,0 +1,290 @@
...
+
+int64_t xls_dslx_enum_def_get_member_count(struct xls_dslx_enum_def* n) {
+  auto* cpp_enum_def = reinterpret_cast<xls::dslx::EnumDef*>(n);
+  return static_cast<int64_t>(cpp_enum_def->values().size());
+}
+
+struct xls_dslx_enum_member* xls_dslx_enum_def_get_member(
+    struct xls_dslx_enum_def* n, int64_t i) {
+  auto* cpp_enum_def = reinterpret_cast<xls::dslx::EnumDef*>(n);
+  xls::dslx::EnumMember& cpp_member = cpp_enum_def->values().at(i);
```
hzeller, line 169: Do you actually need mutable access to this, or could you return a `const struct xls_dslx_enum_member*` ?

xls/dslx/frontend/ast.h
cdleary: I followed your suggestion but I think this is not the convention the Google C++ style guide/arbiters recommend for new code (i.e. `mutable_` prefix was recommended in the pre "references are ok" era). Might be good to resolve for future code.

meheff: We don't general name things with mutable_ (I could only find one instance) and instead just have const and non-const overloads.

hzeller: fair enough. Sorry for the noise (I might've been socialized on earlier more strict code-bases, where it is common practice and strongly enforcuraged).

xls/public/c_api_dslx.cc
meheff: I think it's OOK without const. We use const sparingly through XLS especially in the IR and DSLX AST. See the following for justification (on the IR side):

https://google.github.io/xls/xls_style/#:~:text=Unlike%20most%20data,non%2Dconst%20pointers.

const in C++ has it's uses but I think using it too agressively can be counter productive given it's viral nature,

xls/dslx/frontend/ast.h
meheff: No problem. I do recall some earlier Google style which `mutable` was used. I guess it's still used in the protocol buffer generated C++ api.

PR #1640, opened 2024-09-28
[DSLX:LS] Add support for prepareRename and rename

cdleary:
Works for intra-function and intra-module definitions.

Support is not present for cross-workspace renames yet.

image

### DIFF
--- a/xls/dslx/lsp/find_definition.cc
+++ b/xls/dslx/lsp/find_definition.cc
@@ -54,9 +54,10 @@ const NameDef* GetNameDef(
 
 }  // namespace
 
-std::optional<const NameDef*> FindDefinition(const Module& m, const Pos& selected,
-                                   const TypeInfo& type_info,
-                                   ImportData& import_data) {
+std::optional<const NameDef*> FindDefinition(const Module& m,
+                                             const Pos& selected,
+                                             const TypeInfo& type_info,
+                                             ImportData& import_data) {
   std::vector<const AstNode*> intercepting = m.FindIntercepting(selected);
   VLOG(3) << "Found " << intercepting.size()
           << " nodes intercepting selected position: " << selected;
--- a/xls/dslx/lsp/find_definition.cc
+++ b/xls/dslx/lsp/find_definition.cc
@@ -54,10 +54,9 @@ const NameDef* GetNameDef(
 
 }  // namespace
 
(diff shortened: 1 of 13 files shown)


ericastor:
Mind syncing & rerunning? Looks like CI failed...

cdleary:
> Mind syncing & rerunning? Looks like CI failed...

Yeah, was really change related, had trouble seeing it since baseline was red. Setting parent link on module-top AST nodes was hazardous for some reason I don't fully understand so I made a simpler change. PTAL, thanks.

xls/dslx/lsp/find_definition.cc
```diff
@@ -56,7 +56,8 @@ const NameDef* GetNameDef(
 
 std::optional<Span> FindDefinition(const Module& m, const Pos& selected,
                                    const TypeInfo& type_info,
-                                   ImportData& import_data) {
+                                   ImportData& import_data,
+                                   const NameDef** name_def) {
```
grebe, line 60: `name_def` is inconsistent with the header's `name_def_out` and is shadowed below.

xls/dslx/lsp/find_definition.h
```diff
@@ -41,7 +41,8 @@ namespace xls::dslx {
 // colon-reference to a construct in another module will return nullopt.
 std::optional<Span> FindDefinition(const Module& m, const Pos& selected,
                                    const TypeInfo& type_info,
-                                   ImportData& import_data);
+                                   ImportData& import_data,
+                                   const NameDef** name_def_out = nullptr);
```
grebe, line 45: Rather than returning name_def_out through an argument, it seems easier to have this be `std::optional<NameDef*> FindDefinition(...)` and have existing usages update to get the span from the `NameDef`.

What do you think? Is it really annoying to update the usages?

grebe (changes requested):

xls/dslx/lsp/find_definition.h
cdleary: Good idea, done, PTAL.

xls/dslx/lsp/find_definition.cc
cdleary: No longer a problem since it returns the NameDef as you suggested.

grebe (approved):

hzeller:
Merged.

PR #1650, opened 2024-10-07
Generate clang-tidy comments in pull requests.

hzeller:
This only runs on changed files, so clang-tidy reports are useful and actionable.


hzeller:
Won't work right now with that action as it creates a container and only mounts a subset of what is needed to see all bazel files.

PR #1661, opened 2024-10-13
[dslx:ls] Add support for DocumentHighlight capability.

### LINKED ISSUE #1660: LSP: highlight all other places of a symbol under the cursor (`textDocument/documentHighlight`)
When the cursor is over a symbol, highlight all other places where this particular symbol is used or where it is defined.

This requires to understand the scope of the symbol, but that information should already be fully there (the rename functionality will already do that).

Spec:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentHighlight

This need to receive `verible::lsp::DocumentHighlightParams` and return a `std::vector<verible::lsp::DocumentHighlight>` which are already defined in the `lsp-protocol.yaml`.
The LSP will then advertise that it implements a `documentHighlightProvider` in its capabilities.

cdleary:
Fixes #1660

Sample view when some logging is added to confirm it's the feature is being triggered appropriately:

Screenshot from 2024-10-13 16-55-33

### DIFF
--- a/xls/dslx/lsp/dslx_ls.cc
+++ b/xls/dslx/lsp/dslx_ls.cc
@@ -85,6 +85,7 @@ InitializeResult InitializeServer(const nlohmann::json& params) {
       {"dynamicRegistration", false},
       {"prepareSupport", true},
   };
+  capabilities["documentHighlightProvider"] = true;
   capabilities["documentFormattingProvider"] = true;
   return InitializeResult{
       .capabilities = std::move(capabilities),
@@ -247,6 +248,19 @@ absl::Status RealMain() {
         LspLog() << "could not determine inlay hints; status: "
                  << inlay_hints_or.status() << "\n";
         return std::vector<verible::lsp::InlayHint>{};
+      });
+
+  dispatcher.AddRequestHandler(
+      "textDocument/documentHighlight",
+      [&](const verible::lsp::DocumentHighlightParams& params) {
+        auto highlights_or = language_server_adapter.DocumentHighlight(
+            params.textDocument.uri, params.position);
+        if (highlights_or.ok()) {
+          return highlights_or.value();
+        }
+        LspLog() << "could not perform document highlight(s); status: "
+                 << highlights_or.status() << "\n";
+        return std::vector<verible::lsp::DocumentHighlight>{};
       });
 
   // Main loop. Feeding the stream-splitter that then calls the dispatcher.
--- a/xls/dslx/lsp/language_server_adapter.cc
+++ b/xls/dslx/lsp/language_server_adapter.cc
@@ -421,4 +421,35 @@ LanguageServerAdapter::Rename(std::string_view uri,
   return std::nullopt;
 }
 
+absl::StatusOr<std::vector<verible::lsp::DocumentHighlight>>
+LanguageServerAdapter::DocumentHighlight(
+    std::string_view uri, const verible::lsp::Position& position) const {
+  if (ParseData* parsed = FindParsedForUri(uri); parsed && parsed->ok()) {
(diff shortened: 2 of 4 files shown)


xls/dslx/lsp/dslx_ls.cc
```diff
@@ -249,6 +250,19 @@ absl::Status RealMain() {
...
 
+  dispatcher.AddRequestHandler(
+      "textDocument/documentHighlight",
+      [&](const verible::lsp::DocumentHighlightParams& params) {
+        auto highlights_or = language_server_adapter.DocumentHighlight(
+            params.textDocument.uri, params.position);
+        if (highlights_or.ok()) {
+          return highlights_or.value();
+        }
+        LspLog() << "could not perform document highlight(s); status: "
```
hzeller, line 261: I suspect that log was good for debugging but might be pretty noisy now ? Not all places the cursor passes by is a symbol.

xls/dslx/lsp/language_server_adapter.cc
```diff
@@ -421,4 +421,39 @@ LanguageServerAdapter::Rename(std::string_view uri,
...
+    const Module& module = parsed->module();
+    std::optional<const NameDef*> maybe_definition = xls::dslx::FindDefinition(
+        module, pos, parsed->type_info(), parsed->import_data());
+    if (maybe_definition.has_value()) {
+      const NameDef* name_def = maybe_definition.value();
+      const Span& definition_span = name_def->span();
+      std::vector<verible::lsp::DocumentHighlight> highlights = {
+          verible::lsp::DocumentHighlight{
+              .range = ConvertSpanToLspRange(definition_span),
+              // TODO(cdleary): 2024-10-13 How do we indicate the kind?
```
hzeller, line 440: Currently there is none defined in the verible lsp stuff, so we just default to 'text'.
If we have meaningful read/write annotations we can make (that hopefully will show up in a useful way in editors), we can add it to the protocol schema where this TODO is currently:

 https://github.com/chipsalliance/verible/blob/master/common/lsp/lsp-protocol.yaml#L152

... and then adding the enum here
https://github.com/chipsalliance/verible/blob/master/common/lsp/lsp-protocol-enums.h

hzeller (changes requested):
Nice!

xls/dslx/lsp/dslx_ls.cc
hzeller: Testing it, looks like it is actually not an issue in practice.

hzeller (approved):
LGTM

xls/dslx/lsp/language_server_adapter.cc
```diff
@@ -421,4 +421,39 @@ LanguageServerAdapter::Rename(std::string_view uri,
...
+              //.kind = verible::lsp::DocumentHighlightKind::Write,
```
hzeller, line 441: this comment (and below in 450) look like the want a space between `//` and `.kind` according to the `clang-format` gods.... :)

cdleary: Makes sense, just removed for the time being.

xls/dslx/lsp/language_server_adapter.cc
cdleary: Understood -- not sure how the editor does/would show them differently so just removed them for now.

xls/dslx/lsp/dslx_ls.cc
cdleary: At least in VSCode it looks like it only happened on click. If it shows up as noisy in an editor down the road happy to remove.

PR #1666, opened 2024-10-21
[DSLX:LS] Diagnostics updates for dependent modules.

cdleary:
"Fixes" inconsistency previously observed in the handling across files in the language server.

Previously if you made a change to `leaf.x` that unbroke a module that depended on `leaf.x`, you wouldn't naturally get an update to the "problems" pane in VSCode that indicated the problem had been resolved.

Introduces a very lightweight virtualized filesystem layer so we can see updated text in the editor (language server) workspace reflected in the language server results.

[Screencast from 10-20-2024 02:19:10 PM.webm](https://github.com/user-attachments/assets/37a3dcec-fa0d-4b9a-bfe4-81e7b605e3a8)

### DIFF
--- a/xls/dslx/lsp/dslx_ls.cc
+++ b/xls/dslx/lsp/dslx_ls.cc
@@ -106,11 +106,26 @@ void TextChangeHandler(const std::string& file_uri,
     // Note: this returns a status, but we don't need to surface it from here.
     adapter.Update(file_uri, file_content).IgnoreError();
   });
-  verible::lsp::PublishDiagnosticsParams params{
-      .uri = file_uri,
-      .diagnostics = adapter.GenerateParseDiagnostics(file_uri),
-  };
-  dispatcher.SendNotification("textDocument/publishDiagnostics", params);
+
+  // For now we brute force evaluate all the files in the DAG that may be
+  // sensitive to the update. We'll need to be smarter as we observe scaling
+  // issues (e.g. only evaluate sensitive files if the original file went from
+  // being an import-time error to having no import-time error).
+  std::vector<std::string> sensitive_uris =
+      adapter.import_sensitivity().GatherAllSensitiveToChangeIn(file_uri);
+  for (std::string sensitive_uri : sensitive_uris) {
+    // Note to all dependent files that an update has occurred.
+    // We don't need to do this for the file that we updated directly.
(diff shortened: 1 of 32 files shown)


xls/dslx/lsp/dslx_ls.cc
```diff
@@ -106,11 +106,26 @@ void TextChangeHandler(const std::string& file_uri,
...
-  };
-  dispatcher.SendNotification("textDocument/publishDiagnostics", params);
+
+  // For now we brute force evaluate all the files in the DAG that may be
+  // sensitive to the update. We'll need to be smarter as we observe scaling
+  // issues (e.g. only evaluate sensitive files if the original file went from
+  // being an import-time error to having no import-time error).
+  std::vector<std::string> sensitive_uris =
+      adapter.import_sensitivity().GatherAllSensitiveToChangeIn(file_uri);
+  for (std::string sensitive_uri : sensitive_uris) {
```
hzeller, line 116: this should be a const reference.

xls/dslx/lsp/dslx_ls.cc
```diff
@@ -106,11 +106,26 @@ void TextChangeHandler(const std::string& file_uri,
...
+    if (sensitive_uri != file_uri) {
+      adapter.Update(sensitive_uri, std::nullopt).IgnoreError();
+    }
+
+    verible::lsp::PublishDiagnosticsParams params{
+        .uri = sensitive_uri,
+        .diagnostics = adapter.GenerateParseDiagnostics(sensitive_uri),
+    };
+    dispatcher.SendNotification("textDocument/publishDiagnostics", params);
+  }
```
hzeller, line 128: Can't add comment below in github UI, but around line 190 below you get a callback whenever a file is removed from the editor (i.e. file got closed in the editor). In that case you probably need to do something as you then need to fallback to the regular filesystem in that context, i.e. need to remove it from your virtual filesystem.

```c++
      [&](const std::string& uri, const EditTextBuffer* buffer) {
        if (buffer == nullptr) {
          return;  // <- remove from virtual filesystem ?
        }
        TextChangeHandler(uri, *buffer, dispatcher, language_server_adapter);
      });
```

hzeller (approved):
Modulo clang-tidy recommendations.

PR #1672, opened 2024-10-29
[public:c_api] More type APIs.

cdleary:
Note that this makes the ImportData take a defensive copy of additional search paths.
This is a long-lived context-like object so it should increase safety / ease of use and not change much resource-wise.
(I had previously accidentally written a use-after-free against this API since it previously captured the `Span`.)

* Add more TypeDim-oriented APIs.
* xls_bits_get_bit_count API and fix xls_dslx_type_info_get_type_struct_member
* Add ability to get at underlying TypeAnnotation node in an enum def.
* Add a bunch more APIs and a test for TypeRefs.


xls/public/c_api_dslx.cc
```diff
@@ -167,6 +167,15 @@ struct xls_dslx_enum_def* xls_dslx_module_get_type_definition_as_enum_def(
   return reinterpret_cast<xls_dslx_enum_def*>(cpp_enum_def);
 }
 
+struct xls_dslx_type_alias* xls_dslx_module_get_type_definition_as_type_alias(
+    struct xls_dslx_module* module, int64_t i) {
+  auto* cpp_module = reinterpret_cast<xls::dslx::Module*>(module);
+  xls::dslx::TypeDefinition cpp_type_definition =
+      cpp_module->GetTypeDefinitions().at(i);
+  auto* cpp_type_alias = std::get<xls::dslx::TypeAlias*>(cpp_type_definition);
```
proppy, line 175: are we ok `std::bad_variant_access` here, or should we catch and return `nullptr`?

cdleary: Similar functions just state it's a precondition so I did similar here -- note that the library is built with -fno-exceptions so it should std::terminate on exception.

xls/public/c_api_dslx.cc
```diff
@@ -224,9 +310,12 @@ const struct xls_dslx_type* xls_dslx_type_info_get_type_struct_def(
...
 const struct xls_dslx_type* xls_dslx_type_info_get_type_struct_member(
-    struct xls_dslx_type_info* type_info,
-    struct xls_dslx_struct_member* struct_member) {
-  return GetMetaTypeHelper(type_info, struct_member);
+    struct xls_dslx_type_info* type_info, struct xls_dslx_struct_member* struct_member) {
+  // Note: StructMember is not itself an AST node, it's just a POD struct, so
+  // we need to traverse to its type annotation.
+  auto* cpp_struct_member = reinterpret_cast<xls::dslx::StructMember*>(struct_member);
+  xls::dslx::TypeAnnotation* type = cpp_struct_member->type;
+  return GetMetaTypeHelper(type_info, type);
```
proppy, line 318: nit: should we pass `c_node` as an `AstNode`  here and in other `GetMetaTypeHelper` call site, since it always get casted and used that way?

xls/public/c_api_dslx.cc
```diff
@@ -341,4 +432,125 @@ bool xls_dslx_type_is_signed_bits(const struct xls_dslx_type* type,
...
+}
+
+bool xls_dslx_type_is_array(const struct xls_dslx_type* type) {
+  const auto* cpp_type = reinterpret_cast<const xls::dslx::Type*>(type);
+  return cpp_type->IsArray();
+}
+
+struct xls_dslx_type* xls_dslx_type_array_get_element_type(struct xls_dslx_type* type) {
+  const auto* cpp_type = reinterpret_cast<const xls::dslx::Type*>(type);
+  CHECK(cpp_type->IsArray());
```
proppy, line 452: q: should we check and return `nullptr` instead similar to the other function (or use `CHECK` everywhere?)

xls/public/c_api_dslx.cc
```diff
@@ -341,4 +432,125 @@ bool xls_dslx_type_is_signed_bits(const struct xls_dslx_type* type,
...
+  const auto* struct_def = reinterpret_cast<const xls_dslx_struct_def*>(&cpp_struct_def);
+  // const_cast is ok because the C API can only do immutable query-like things with
+  // the node anyway.
+  return const_cast<xls_dslx_struct_def*>(struct_def);
+}
+
+bool xls_dslx_type_to_string(const struct xls_dslx_type* type, char** error_out,
+                             char** result_out) {
+  const auto* cpp_type = reinterpret_cast<const xls::dslx::Type*>(type);
+  *error_out = nullptr;
```
proppy, line 493: q: should we use nullptr as a result here to be consistent with the other function, or is there something different about `to_string`?

xls/dslx/import_data.h
```diff
@@ -281,7 +282,7 @@ class ImportData {
   absl::flat_hash_map<Module*, AstNode*> typecheck_wip_;
   TypeInfoOwner type_info_owner_;
   const std::filesystem::path stdlib_path_;
-  absl::Span<const std::filesystem::path> additional_search_paths_;
+  std::vector<std::filesystem::path> additional_search_paths_;
```
proppy, line 285: q: was this the use after free you mentioned in the description?

proppy (approved):
just some nits and qs.

xls/public/c_api_dslx.cc
cdleary: Good point, it would have helped show this error was there to have the casts present instead of burying them in the type template. Fixed to be explicit.

xls/public/c_api_dslx.cc
cdleary: Yeah, there's 3-4 ways to approach APIs like this:
- Defensive -- try to detect everything that can go wrong and return null or an error via an expanded signature -- for functions that are documented as fallible we often return the result as an outparam for this
- Asserting -- document preconditions and try to CHECK them
- Debug-build Asserting -- document preconditions and DCHECK them
- Lax -- document preconditions and assume they've been appropriately followed

I'd say these APIs have been either Asserting or Lax, only trend towards "defensive" is there when the API signatures are expanded to e.g. propagate back absl::Status data.

LMK if that makes sense.

xls/dslx/import_data.h
cdleary: This is the defensive copy storage I mentioned. The use after free error I made was something I did in the Rust crate usage of the C API.

xls/public/c_api_dslx.cc
cdleary: It has a fallible C API, but it's not actually fallible, so we unconditionally know that the error output is null.

One could make the case we should switch it to reflect that it's infallible, but it leaves room in case ToString() returned a StatusOr in the future for some reason. Could do either way.

xls/public/c_api_dslx.cc
proppy: but is that different from `xls_dslx_type_definition_get_colon_ref` below which does check the return value from `std::get`? (is the difference that the later works directly w/ the `type_definition` object so that we expect people might want to check different scenario while this assume intent, i.e: the caller knows that the type definition at `i` is a `type_alias`?

xls/public/c_api_dslx.cc
proppy: Yep, thanks for taking the time to spell that out.

xls/public/c_api_dslx.cc
cdleary: Ah, this is because TypeDefinition is literally a variant. This allows us to perform both std::holds_alternative and std::get in a single call, it's like std::get_if. Pretty much all other functions here are not just reading out from a variant where we don't know the contents.

I added a comment in the `.h` on this.

cdleary:
Looks like this landed in https://github.com/google/xls/commit/85477aece89cadc21b5696ee06a83cc5877c68b4 but copybara didn't close out the PR?

proppy:
> Looks like this landed in 85477ae but copybara didn't close out the PR?

Yes, I think that's because there was some extra changes (comments) that got added after it got submitted internally.

PR #1689, opened 2024-11-02
[DSLX:TS] Give an error if there is an ellipsis without a type annotation.

cdleary:
Previously we would assume we do not need to extend beyond the provided (literal) element count, which led to errors when users did:

```
fn f() -> u32[4] {
    let x: u32[4] = [u32:42, ...];  // RHS is actually 1 element here, the u32[4] from the type annotation does /not/ propagate into the array expression as context!
}
```

This also relates to our recent unifying type inference discussions -- in unifying TI we'd make the right hand side type "at least a u32[1] if not more" and then it would unify with `u32[4]` to say "ok it's a u32[4] filled with the value 42". cc @richmckeever @meheff

### DIFF
--- a/xls/dslx/type_system/deduce_expr.cc
+++ b/xls/dslx/type_system/deduce_expr.cc
@@ -139,6 +139,15 @@ absl::StatusOr<std::unique_ptr<Type>> DeduceArray(const Array* node,
       member_types[0]->CloneToUnique(), member_types_dim);
 
   if (node->type_annotation() == nullptr) {
+    if (node->has_ellipsis()) {
+      const Span array_obrack_span(node->span().start(), node->span().start());
+      return TypeInferenceErrorStatus(
+        array_obrack_span, inferred.get(),
+        "Array has ellipsis (`...`) but does not have a type annotation; "
+        "please add a type annotation to indicate how many elements to expand to.",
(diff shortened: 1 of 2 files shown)


xls/dslx/type_system/deduce_expr.cc
```diff
@@ -139,6 +139,15 @@ absl::StatusOr<std::unique_ptr<Type>> DeduceArray(const Array* node,
       member_types[0]->CloneToUnique(), member_types_dim);
 
   if (node->type_annotation() == nullptr) {
+    if (node->has_ellipsis()) {
+      const Span array_obrack_span(node->span().start(), node->span().start());
+      return TypeInferenceErrorStatus(
+        array_obrack_span, inferred.get(),
+        "Array has ellipsis (`...`) but does not have a type annotation; "
+        "please add a type annotation to indicate how many elements to expand to.",
```
cchan, line 147: "For example: `u8[32]:[u8:0, ...]`"

cdleary: Good idea -- tried to make it relevant to what the user typed in instead of having a universal sample (which might be confusing if they didn't type in e.g. `u8` as the element type).

meheff (approved):

Issue #1049, opened 2023-06-28
[dslx] comment-to-doc extraction capabilities (a la rustdoc?)


cdleary:
With the auto-formatter we'll grow a way to associate comments with AST nodes, which we can leverage to automatically spit out javadoc like API / function documentation. We currently do that manually with the DSLX stdlib.

Also opens up the possibility for doctesting.

PR #1707, opened 2024-11-12
[DSLX:stdlib] Add `std::distinct`.

cdleary:


docs_src/dslx_std.md
```diff
@@ -1188,6 +1188,15 @@ index is used in a match expression (which will eagerly evaluate all of its
...
 ultimately discarded from the unselected match arm.
 
+#### `std::distinct`
+
+```dslx-snippet
+pub fn distinct<COUNT: u32, N: u32, S: bool>(items: xN[S][N][COUNT], valid: bool[COUNT]) -> bool
+```
+
+Returns whether all the `items` are distinct (i.e. there are no duplicate
+items) after the `valid` mask is applied.
```
ericastor, line 1198: This could possibly use an example or two, to clarify what it means to apply the `valid` mask in this case.

ericastor (commented):
Looks good! Will approve with an example or two added to the docs, since it seems worth explaining the mask a bit further.

docs_src/dslx_std.md
cdleary: I duplicated a unit test into the doc -- I'm not sure it's generally wise. Ideally we'd be able to link to a searched piece of text in the stdlib but github isn't capable of doing that. Really we need a rustdoc like system for DSLX to make this proposition more sustainable. See also #1049

ericastor (approved):

cdleary:
Thanks Eric! I'm also going to keep an eye out for rustdoc-style opportunities in case there are two-birds one-stone kind of situations there.

PR #1709, opened 2024-11-13
[DSLX] Rename slice builtin to array_slice to match array_rev (vs rev) and array_size (vs std::sizeof)

### LINKED ISSUE #1688: [enhancement] slice should be renamed array_slice

https://google.github.io/xls/dslx_std/#slice is noticably named differently from https://google.github.io/xls/dslx_std/#array_rev and https://google.github.io/xls/dslx_std/#array_size

We should probably call it `array_slice` for symmetry.

To keep calling it `slice`, since for bits we have bit slicing syntax.

I think ideally we'd rename them for consistency.

cdleary:
Fixes #1688

Also fix a `unsigned_sizeof` straggler in the docs I saw along the way.

### DIFF
--- a/docs_src/dslx_std.md
+++ b/docs_src/dslx_std.md
@@ -957,14 +957,14 @@ Counts the number of bits in `x` that are '1'.
 #### `std::extract_bits`
 
 ```dslx-snippet
-pub fn extract_bits<from_inclusive: u32, to_exclusive: u32, fixed_shift: u32,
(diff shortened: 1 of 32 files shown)


meheff (approved):

cdleary:
Looks like this got merged just not in a way copybara closed it, thanks all!

proppy:
> Looks like this got merged just not in a way copybara closed it, thanks all!

Yes, we had to also fix internal code that was impacted by this.

PR #1720, opened 2024-11-15
[VAST] `ExternPackageType` & tighten space invariants.

cdleary:
Adds an `ExternPackageType` to refer to types defined in other packages (that we do not know the underlying bit width for or have a definition to refer to in the current `VerilogFile`).

In the process found it was hard to reason about the invariants of `Emit()` giving back associated spaces for types and `EmitWithIdentifier()` and similar -- reworked that so all Emit functionality gives back a value with no whitespace to strip so the caller knows what to expect. Previously it was confusing/confused because DataKind could be empty string and DataType could also be empty string in various cases. Case in point this fixed an error in the test where there were two spaces after the `output` keyword.

Also adds a bunch of tests for constructs we're about to use in stitching just to ensure they were all possible.

### DIFF
--- a/xls/codegen/vast/vast.cc
+++ b/xls/codegen/vast/vast.cc
@@ -117,6 +117,25 @@ bool CannotStripWhitespace(std::string_view s) {
   return stripped.size() == s.size();
 }
 
+// Helper that combines DataKind and DataType strings -- since either of these
+// can be empty under certain circumstances we consolidate the handling of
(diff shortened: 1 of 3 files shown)


cdleary:
@proppy note I got "Nighly Build MkDocs And Commit HTML / deploy (pull_request)" failure on this PR

meheff (approved):
Nice! Thanks for cleaning up the whitespace mess.

Issue #1731, opened 2024-11-19
LSP: `DSLX_PATH` stopped working


hzeller:
The `DSLX_PATH` environment variable used to be the way to tell he language server what the base path is for finding import paths.

Let's assume the XLS checkout is in a subdirectory of my homedir: `/home/myhomedir/src/xls` and thelanguage server configured for the editor.

Now opening a DSLX file. To make sure the language server finds all import paths, I am setting the `DSLX_PATH` to the root of the directory:

```
DSLX_PATH=/home/myhomedir/src/xls emacs src/xls/xls/dslx/stdlib/float32.x
```

This now lets the language server crash while opening the file:
```
XLS testing language server
Path configuration:
        stdlib=xls/dslx/stdlib
        dslx_path=/home/myhomedir/src/xls
        cwd=/home/myhomedir/src/xls
F1119 11:00:35.471252 1009793 language_server_adapter.cc:232] Check failed: absl::StartsWith(importer_filename, "file://") || absl::StartsWith(importer_filename, "memfile://") importer_filename: xls/dslx/stdlib/apfloat.x imported: "xls/dslx/stdlib/std.x"
```

It looks like some part of the code assumes regular host filenames, while others assume the `file://` URI notation, which then clashes here.

hzeller:
Looks like this broke with the introduction of the virtual filesystem in 1b175c677d758252af957a6c0cbba675cdda9b46
Maybe @cdleary sees immediately what is amiss.

cdleary:
@cdleary Please close if this is working for you now, I think with the test I added it should be 🙏

hzeller:
Yes, works now, thanks!

PR #1737, opened 2024-11-21
[DSLX:LS] Fix/tighten up URI/path conversions, use VirtualizableFilesystem pervasively in DSLX

cdleary:
Review note:
- most of this change is plumbing to ensure consistency with respect to the VirtualizableFilesystem layer and resulting consolidation (e.g. before we had get_contents lambda which is now subsumed by the VirtualizableFilesystem, and there are a few variants for common test uses of get_contents, like `UniformContentFilesystem`).
- the remainder of it is using a LspUri type to ensure types represent the conversion between filesystem paths and LSP URIs

---

We now have a wrapper type to indicate when things have been checked to be in URI form, and that type offers convenience APIs for conversion to `string_view` for the underlying and to `std::filesystem::path` to strip off the URI prefix.

Previously we ran into trouble in #1731 because the stdlib path was a relative path and pointed directly at the filesystem, and thus was clashing with the expectation that everything would be in URI form; i.e. a module would import a stdlib module and then that would have a filesystem (non-URI) path, and that would potentially import other stdlib modules with more filesystem (non-URI) paths.

We check as invariant in the file table that no URIs are held there -- the conversion into and out of URI form happens on the import boundary.

This PR also plumbs virtualizable filesystem into the error printing routines so they don't try to look at the real filesystem when they should be looking at the virtualized filesystem contents, e.g. in any error reporting done for parse errors encounterd by the LSP. Now that the LSP-parsed span filenames look indistinguishable from real filesystem filenames we have to be more careful to always try to use the virtual filesystem abstraction instead of `xls/common/file/filesystem.h` routines in retrieving file contents.

This, as a result, plumbs the virtual filesystem in everywhere to places that GetContents is called to ensure consistency across the DSLX facilities. In general we shouldn't need to rely on `//xls/common/file:filesystem` as a dep except in tools that write output files (as the VirtualizableFileSystem doesn't have SetContents). Note that usually there is an ImportData around somewhere, as it is a fairly primordial context-like object, and there is a `vfs()` hanging off of that.

Ran clang-tidy and bant dwyu


xls/dslx/frontend/ast_cloner_test.cc
```diff
@@ -1442,12 +1442,8 @@ proc MyProc {
   absl::StatusOr<std::unique_ptr<Module>> module_or =
       ParseModule(kProgram, "fake_path.x", "the_module", file_table);
   if (!module_or.ok()) {
-    TryPrintError(
-        module_or.status(),
-        [&](std::string_view path) -> absl::StatusOr<std::string> {
-          return std::string(kProgram);
-        },
-        file_table);
+    UniformContentFilesystem vfs(kProgram);
```
hzeller, line 1445: there is no header included for `UniformContentFilesystem `

xls/dslx/virtualizable_file_system.h
```diff
@@ -0,0 +1,95 @@
+#ifndef XLS_DSLX_VIRTUALIZABLE_FILE_SYSTEM_H_
+#define XLS_DSLX_VIRTUALIZABLE_FILE_SYSTEM_H_
+
+#include <filesystem>  // NOLINT
+
```
hzeller, line 5: Some `#include`s missing for `<string>`, `<string_view>`, `<optional>`, StrFormat...

see clang-tidy output.

xls/dslx/lsp/lsp_uri.h
```diff
@@ -0,0 +1,69 @@
+#ifndef XLS_DSLX_LSP_LSP_URI_H_
+#define XLS_DSLX_LSP_LSP_URI_H_
+
+#include <filesystem>  // NOLINT
+#include <string>
+
```
hzeller, line 6: `<string_view>`, `<ostream>`...

xls/dslx/lsp/lsp_uri.cc
```diff
@@ -0,0 +1,24 @@
+#include "xls/dslx/lsp/lsp_uri.h"
+
```
hzeller, line 2: `<filesystem>`,`<string>`, `<utility>`...

hzeller:
Very nice!

Running `xls/dev_tools/run-clang-tidy-cached.sh`, I do get a bunch of missing includes for some of the referenced filesystem classes or some of the system includes such as `<string_view>`, so maybe that needs some attention (`grep misc-include-cleaner xls_clang-tidy.out`).

(I found that our current way of building the compilation DB has issues (at least on Linux) to properly succeed because it is confused about the dylib stuff, so I found on Linux I have to comment out in `xls/public/BUILD`

```
#libxls_dylib_binary()
#pytype_test_test_c_api_symbols()
```
)

cdleary:
> Very nice!
> 
> Running `xls/dev_tools/run-clang-tidy-cached.sh`, I do get a bunch of missing includes for some of the referenced filesystem classes or some of the system includes such as `<string_view>`, so maybe that needs some attention (`grep misc-include-cleaner xls_clang-tidy.out`).

Fixed almost all actionable clang-tidies outside of contrib, PTAL.

meheff (approved):

meheff:
> Running `xls/dev_tools/run-clang-tidy-cached.sh`, I do get a bunch of missing includes for some of the referenced filesystem classes or some of the system includes such as `<string_view>`, so maybe that needs some attention (`grep misc-include-cleaner xls_clang-tidy.out`).

The github PR to google internal submission is already cumbersome so I think we should avoid adding the additional requirement of keeping things include clean. Include cleanliness, in general, is of small value (not zero, but small). I don't think it is worth people's time to either check for cleanliness on PR's or go much out of your way to fix them.  We have tools internally which can better handle these cleanups such as automatically flagging them in code review and transforming the code automatically. We can do this on a regular cadence internally if we care enough.

cdleary:
> This is currently failing to import for internal review due to the lack of the license header on the new files being added. Would you mind adding those?

Done, sorry I forgot. I'll contribute a dev_utils script to help catch this in a followup PR.

PR #1751, opened 2024-11-28
[codegen] Support genvars & generate loop, use in simple array updates.

cdleary:
Previously we would create a lot of Verilog output spew (even for simple cases), which can choke downstream tools that are not accustomed to generated files.

This PR sets its sights a bit low and only handles a simple case (1D array of bits typed objects) to keep things a bit more manageable for review. We can extend this technique to nested generate loops in the future from this basis.

VAST note: generate loops are interesting in that they are not statement blocks in a sense, they are “AST construct yielding”, i.e. they yield AST constructs at elaboration time that are appropriate for the enclosing scope. Which is to say, it could be creating ModuleMembers even though the generate loop’s body block is seemingly “within” the direct module scope. As a result, we have the body for a generate loop be a sequence of VastNodes to avoid artificially limiting what can be placed inside of them; i.e. a StatementBlock is not usable in this context.


meheff (approved):

hzeller:
Nice!

PR #1758, opened 2024-12-03
[c_api] Add a few more DSLX accessors that help with conversion to SV.

cdleary:
- xls_dslx_module_get_name
- xls_dslx_colon_ref_get_attr
- xls_dslx_import_get_subject
- xls_dslx_module_get_name


hzeller (approved):

xls/public/c_api_dslx.cc
```diff
@@ -149,6 +149,12 @@ xls_dslx_type_definition_kind xls_dslx_module_get_type_definition_kind(
                      cpp_type_definition);
 }
 
+char* xls_dslx_module_get_name(struct xls_dslx_module* module) {
+  auto* cpp_module = reinterpret_cast<xls::dslx::Module*>(module);
+  std::string result = cpp_module->name();
```
hzeller, line 154: why not `const std::string&` (also in other places).

cdleary: Done.

PR #1774, opened 2024-12-07
[fuzz] Fuzz `xN` usage and fix the fallout.

cdleary:
You'll see changes/fixes are all in the DSL and fuzzer implementation.

- There was at least one instance of `down_cast` I missed in the previous PR because I was searching for `dynamic_cast`.
- Some of the bugs came from if/else ladders where array-ness was being tested before BitsLike polymorphism, where now array-of-bits-constructor makes you bits like so that should be tested first.
- Needed to add ability to interrogate width/signedness of xN in the AST helpers so the fuzzer could leverage it in the existing ways it queries primitive types.
- Removed a few of the double-checks from the bytecode emitter on casts, they only serve to conflict / act as change detectors with respect to the type checker.


meheff (approved):

xls/fuzzer/crashers/crasher_2024-12-07_9fe0.x
```diff
@@ -0,0 +1,194 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// BEGIN_CONFIG
+// # proto-message: xls.fuzzer.CrasherConfigurationProto
+// exception: "SampleError: Result miscompare for sample 0:\nargs: bits[4]:0x8; bits[46]:0x1224_4d98_43c9; bits[3]:0x4; bits[34]:0x8; bits[9]:0x9b; bits[56]:0x60_7640_1285_1346\nevaluated opt IR (JIT), evaluated opt IR (interpreter), evaluated unopt IR (JIT), evaluated unopt IR (interpreter) =\n   (bits[13]:0x0, bits[44]:0x7ff_ffff_ffff, bits[13]:0x0)\ninterpreted DSLX =\n   (bits[13]:0x0, bits[44]:0x0, bits[13]:0x0)"
+// issue: "DO NOT SUBMIT Insert link to GitHub issue here."
```
hzeller, line 17: Can you update the `DO NOT SUBMIT` line ?

meheff (approved):

xls/fuzzer/crashers/crasher_2024-12-07_9fe0.x
meheff: No github issue is fine. We only file issues if there is non-zero time between bug found and bug fixed.

mikex-oss:
@cdleary this was in the process of going in but then hit some merge conflict on xls/dslx/constexpr_evaluator.cc .

Now I see that the latest push https://github.com/google/xls/commit/04b529f4e1643dd5958fc5b2db55989fb4d37933 seems to reintroduce the DO NOT SUBMIT in the crasher.

cdleary:
@mikex-oss thanks for the heads up, I thought it had to be rebased to head but I accidentally sliced off the last commit -- should be good now if this passes tests :crossed_fingers:

mikex-oss:
@cdleary heads up...I think this PR may be responsible for a few fuzzing failures overnight with two signatures:

1. "Attempting to fetch value instead of handling error INVALID_ARGUMENT: Type "xN" has no defined signedness."
2. Generating empty arrays for xN[false][N] array constants, leading to "Empty array must have a type annotation; please add one"

cdleary:
@mikex-oss is it possible to paste the full crasher? Happy to fix but relatively difficult to go hunting if we've got some samples on hand.

mikex-oss:
@cdleary there weren't any crasher artifacts since they fail before then.

For (1), here's the full stack trace [fail1.log](https://github.com/user-attachments/files/18104988/fail1.log).

For (2), here's basically the crasher (pulled from the error logs using some IDE column selection mode editing...
[fail2.x.txt](https://github.com/user-attachments/files/18104993/fail2.x.txt). (Not exactly the same one from above, as I couldn't dig it up).

PR #1785, opened 2024-12-11
[DSLX] Switch UseTreeEntry to an AstNode, make a cloner implementation.

cdleary:
Switch to AstNode for UseTreeEntry is per discussion with @richmckeever -- the rationale is, though we don't quite have it at the moment, in the future we may want to make an arbitrary matcher or replacer that was polymorphic on AstNode types, which makes it useful here to opt to reflect more structure of the AST in the node class form.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -23,6 +23,9 @@ build --copt "-D_HAS_AUTO_PTR_ETC=0"   --host_copt "-D_HAS_AUTO_PTR_ETC=0"
 build --copt "-Wall"                   --host_copt "-Wall"
 build --copt "-Wextra"                 --host_copt "-Wextra"
 
+# Turn some specific warnings into errors.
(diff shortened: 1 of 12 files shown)


mikex-oss (approved):

dplassgit:
This is failing compilation: `xls/dslx/type_system/type_info_to_proto.cc:55... enumeration value 'kUseTreeEntry' not handled in switch [-Werror,-Wswitch]`

cdleary:
Oh, I think I see what happened, it's flagged as a warning but probably warnings-as-errors is on in google3 in some way that's not reflected in the OSS repo:
Screenshot 2024-12-11 at 11 11 26 AM

cdleary:
Note this is ready to go and I updated the bazelrc to flag an error on this warning in the OSS build going forward.

dplassgit:
Thanks, looks good.

PR #1787, opened 2024-12-11
[docs] Fix missing interpreter_main flags in hello xls tutorial.

schilkp:
Small addition to the "Hello XLS" tutorial: The `trace!` output is not visible by default, and requires `--alsologtostderr` to be seen.

### DIFF
--- a/docs_src/tutorials/hello_xls.md
+++ b/docs_src/tutorials/hello_xls.md
@@ -107,7 +107,7 @@ up! Open a terminal and execute the following in the XLS checkout root
 directory:
 
 ```
-$ ./bazel-bin/xls/dslx/interpreter_main hello_xls.x
+$ ./bazel-bin/xls/dslx/interpreter_main hello_xls.x --alsologtostderr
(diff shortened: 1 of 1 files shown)


docs_src/tutorials/hello_xls.md
```diff
@@ -119,6 +119,9 @@ trace of hello_string @ hello.x:4:17-4:31: [72, 101, 108, 108, 111, 44, 32, 88,
 [===============] 1 test(s) ran; 0 failed; 0 skipped.
 ```
 
+Note that the `trace!` output is treated as `INFO` level logging,
+which is printed to stderr by adding the '--alsologtostderr', making it visible.
```
mikex-oss, line 123: Nit: the multiple clauses is a bit wordy. 

How about "which is made visible in stderr by adding the '--alsologtostderr'.

schilkp: Sure! Fixed + force-pushed :)

PR #1798, opened 2024-12-17
[xls][mlir] Fix XLS IR `Next` node generation in MLIR translation.

schilkp:
Fixes a small typo in the generation of XLS IR `Next` node in the `xls_translate` tool, which previously assigned the wrong `value`.

Also enables and fixes FileCheck in the `ops_translate.mlir` test, and adds a second `(predicate, next_value)` pair to the `xls.next_value` op. This makes it a valid `xls.next_value` op  (one and only one predicate is true) and triggers the bug.


jpienaar (approved):
Thanks for the fix!

Issue #352, opened 2021-03-22
If DSLX mimics Rust, why not simply use the keyword "use", but "import"?


cg31:
I think if DXL keeps things close to Rust, someday it will be easy to port code between them.

cdleary:
Good point, this is a vestige from "an old way" we did things when we were first prototyping (before we started mimicing Rust as much as possible). We currently use path-based resolution of modules for import, so the mechanism is slightly different at the moment, but we can at a minimum remove the unnecessary superficial differences here.

taktoa:
The inflexibility of our imports in comparison to Rust's was kind of unpleasant during recent work; in particular, `use foo::bar as baz;` would have been handy.

cdleary:
I think rough steps here are:
- Parse the construct
- Teach frontend routines how to traverse the symbols in a similar to fashion to imports
- Profit???

There might be challenges where currently the symbol table always knows what local symbol are types vs more arbitrary identifiers; i.e. if you refer to `foo` with no colon ref it's known what "kind" of thing it is (i.e. type, constant, function, etc), only when you do a ColonRef a la `some_module::identifier` are you unsure what kind of thing it is. I think cleaning up any of those kinds of instances will just be a positive for the frontend's making fewer assumptions.

cc @richmckeever as FYI I may start on this -- I /could/ put it behind a feature flag, but we haven't really been doing that, more treating head as "nighly" style where features get developed, so we'll just port samples in the code base over once the functionality is fully supported. LMK if you have thoughts.

meheff:
Just chatted with Rich and in general it sounds good to us. Do you have examples of what fancier imports might look like? I don't know rust well.

cdleary:
To take an example from the repo something like:

```
import xls.modules.rle.rle_common as rle_common;

type DecInData  = rle_common::CompressedData;
type DecOutData = rle_common::PlainData;
```

would instead look like:

```
use xls::modules::rle::rle_common::{CompressedData as DecInData, PlainData as DecOutData};
```

The main difference from normal Rust is that we use all filesystem-based resolution of the imported path. In Rust, at least in Cargo, crates are placed into your namespace implicitly by a metadata file. Currently we can get something kind of similar via `--dslx_path` flag.

rw1nkler:
In my opinion, this can greatly improve the readability of DSLX

cdleary:
Note also that the stdlib is getting bigger over time, and support for `use` will make it easier to break it into pieces grouped by functionality.

PR #1812, opened 2024-12-21
[DSLX] Language test for a simple module with a multi-`use` binding

cdleary:
- Create a helper for linearizing the subjects in a use tree
- Generalize import routines to accommodate both `import` and `use` via common helpers
- Language test with multiple use bindings from a single imported module
- Fake filesystem helper so we can pass a map contents for use as a filesystem

Towards #352


cdleary:
Closing for now, want to think a bit through the approach for importing parameterized functions, before instantiation was driven through ColonRef showing it was an external reference.

PR #1815, opened 2024-12-29
Include the language server in the binary distribution tar.

### LINKED ISSUE #1814: Distribution should include language server
For folks getting started with XLS, the language server is very helpful and should be included in the standard binary distribution.

hzeller:
Fixes #1814


hzeller:
TODO: also add version check in https://github.com/google/xls/blob/main/.github/workflows/nightly-ubuntu-22.04.yml#L51-L63

hzeller:
@proppy alright, added version check as suggested (also: added `dslx_fmt`, as that was not covered yet)

proppy (approved):
nit: maybe add a link in the change description about the missing version check

PR #1819, opened 2024-12-31
[DSLX:BC] Fix disagreement between typechecker and interpreter on whether `trace` returns its operand

cdleary:
Previously the bytecode interpreter thought it returned like `trace_fmt!` does, but historically `trace!` acted as an identity function so you could easily wrap it around any given subexpression, like:

```
let binding = f(a * b + c);
```

instrumenting the interior subexpression like so:
```
let binding = f(trace!(a*b)+c);
```

The fact they disagreed led to typecheck flagging a program through but then getting a bytecode interpreter error in the bowels of InterpValue.

Note that `trace_fmt!` and `trace!` are different in that the latter requires and operand but the former can just take a format string and zero operand args, so they are implemented as two bytecode ops that call into common functionality.

### DIFF
--- a/xls/dslx/bytecode/bytecode.h
+++ b/xls/dslx/bytecode/bytecode.h
@@ -174,7 +174,10 @@ class Bytecode {
     // Swaps TOS0 and TOS1 on the stack.
     kSwap,
     // Prints information about the given arguments to the terminal.
-    kTrace,
+    kTraceFmt,
+    // Prints a single operand argument to the terminal and acts as identity
+    // function (i.e. places the operand value back on the stack).
+    kTraceArg,
     // Slices out TOS0 bits of the array- or bits-typed value on TOS2,
     // starting at index TOS1.
     kWidthSlice,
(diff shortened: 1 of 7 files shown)


xls/dslx/bytecode/bytecode.h
```diff
@@ -174,7 +174,11 @@ class Bytecode {
     // Swaps TOS0 and TOS1 on the stack.
     kSwap,
     // Prints information about the given arguments to the terminal.
-    kTrace,
+    kTraceFmt,
+    // Prints a single operand argument to the terminal and acts as identity
+    // function
```
richmckeever, line 179: Nit: is the line break after "function" supposed to be there?

richmckeever (approved):

xls/dslx/bytecode/bytecode.h
cdleary: Fixed, thanks.

meheff (approved):

PR #1825, opened 2025-01-03
[DSLX:TS] Fixes ability to use `xN` based types as bit slices, use for `std::{min,max}`

cdleary:
Note that this is a straggling case of over-assumption about which types are `BitsLike`, many of which were fixed by b7a6e5faea864b5a14e4682f3c19f8f40510b0f8 and c54b7f7e4f9e56a821d240dafb69414a2682e7b7


meheff (approved):
Nice!

cdleary:
Landed with https://github.com/google/xls/commit/e8dacaaa7d18740240718963ba8a5b97714d5224

PR #1834, opened 2025-01-07
Fix incorrect/missing IR semantics documentation

schilkp:
- Adds missing `nand` and `nor` operations and clarifies their behavior.

- Adds missing reduction operations.

- Correct typing of `eq` and `ne` operands: They can be any type, not just `bits`

Please especially check the following, as far as I can tell this is the behavior of the IR interpreter:

> If only a single argument is provided, the operations `and`, `or`, and `xor` are a no-op, while `nand` and `nor` act as a bit-wise not.

For reduction ops:

> For an operand of width 1, they act as no-ops.

Cheers!

### DIFF
--- a/docs_src/ir_semantics.md
+++ b/docs_src/ir_semantics.md
@@ -191,13 +191,16 @@ Operation  | Opcode          | Semantics
 ### Variadic bitwise operations
 
 Performs a bit-wise operation on one-or-more identically-typed bits operands. If
-only a single argument is provided the operation is a no-op.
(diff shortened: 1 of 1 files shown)


meheff (approved):
Thanks for the documentation fixes!

allight:
Ok so whatever messed up #1828 seems to have also hit this cl and the changes are already merged but attributed to https://github.com/google/xls/commit/085321c23331668a2cb185e3cbc8e00a173311ae

allight:
yeah this was merged but due to some issues with our internal migrations it ended up squashed

PR #1843, opened 2025-01-13
[xls][mlir] Add missing MLIR dialect ops

schilkp:
@jpienaar 

Adds the following XLS-supported ops to the MLIR dialect:

- nand, nor
- and_reduce, or_reduce, xor_reduce
- umulp, smulp
- gate

Note that for the partial product operations, operands of different bit widths and operands of different bit width than the result are explicitly allowed.

Please double check the following in the tablegen ODS for partial product ops:
```tablegen
TypesMatchWith<"result types must match", "result_lhs", "result_rhs", "$_self">
```
Is this a sane way of constraining the two result values to have the same type? Is there a different constraint that matches better? Seems like a bit of a hack to use this with the "identity transform expression" "$_self" :)

### DIFF
--- a/xls/contrib/mlir/IR/xls_ops.td
+++ b/xls/contrib/mlir/IR/xls_ops.td
@@ -241,10 +241,37 @@ def Xls_AndOp : Xls_NaryOp<"and", [Commutative, Pure, SameOperandsAndResultType]
   );
 }
 
+def Xls_NandOp : Xls_NaryOp<"nand", [Commutative, Pure, SameOperandsAndResultType]> {
+  let summary = "Bitwise NAND operation";
+  let description = [{
+    result = ~(operand_0 & ... & operand_{N-1})
+  }];
+  let arguments = (ins
+    Variadic<Xls_Bits>:$inputs
+  );
+  let results = (outs
+    Xls_Bits:$result
+  );
+}
+
 def Xls_OrOp : Xls_NaryOp<"or", [Commutative, Pure, SameOperandsAndResultType]> {
   let summary = "Bitwise OR operation";
   let description = [{
     result = operand_0 | ... | operand_{N-1}
+  }];
+  let arguments = (ins
+    Variadic<Xls_Bits>:$inputs
+  );
+  let results = (outs
+    Xls_Bits:$result
+  );
+}
+
+
+def Xls_NorOp : Xls_NaryOp<"nor", [Commutative, Pure, SameOperandsAndResultType]> {
+  let summary = "Bitwise NOR operation";
+  let description = [{
+    result = ~(operand_0 | ... | operand_{N-1})
   }];
   let arguments = (ins
     Variadic<Xls_Bits>:$inputs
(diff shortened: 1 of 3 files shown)


xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -265,6 +292,48 @@ def Xls_XorOp : Xls_NaryOp<"xor", [Commutative, Pure, SameOperandsAndResultType]
...
+  Xls_Op<name, !listconcat(traits, [
+    ShapesAreConsistent<["input", "result"]>,
+    Elementwise,
+    Scalarizable,
+  ])> {
+  let arguments = (ins
+    Xls_Bits:$input
+  );
+  let results = (outs
+    Xls_Bits:$result
```
jmolloy, line 309: Am I correct in saying that the element type of $result should always be i1? If so, can this please be encoded as a constraint?

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -305,22 +374,65 @@ class Xls_ArithBinaryOp<string name, string desc>
   );
 }
 
+class Xls_PartialProdOp<string name, string desc> :  Xls_Op<name, [
+  Pure,
+  ShapesAreConsistent<["lhs", "rhs", "result_lhs", "result_rhs"]>,
+  Elementwise,
+  Scalarizable,
+  TypesMatchWith<"result types must match", "result_lhs", "result_rhs", "$_self">
```
jmolloy, line 382: I think simply AllTypesMatch<["result_lhs", "result_rhs"]> should suffice here.

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -1881,6 +1993,48 @@ def Xls_TraceOp : Xls_Op<"trace", [TensorArrayTypeFungible,
...
+// Other side-effecting operations
+//===----------------------------------------------------------------------===//
+
+def Xls_GateOp :
+  Xls_Op<"gate", [
+    ShapesAreConsistent<["data", "result"]>,
+    Elementwise,
+    Scalarizable,
+]> {
+  let summary = "Gates an arbitrarily-typed value based on a condition.";
```
jmolloy, line 2006: nit: remove period at the end of summary (it is used to autogenerate error messages).

jmolloy (approved):

jpienaar (approved):
Looks good with James's comments addressed.

schilkp:
Regarding the `AllTypesMatch` constraint: 

Ok this sent me on a bit of a journey :)

I had actually tried `AllTypesMatch` originally, but this did not link because
`inferReturnTypes` was not defined for the partial product ops. Since we cannot 
infer return types here, I assumed this constraint was not the correct one and used 
`TypesMatchWith`.

With your review I dug into this a bit more. Adding the constraint
`AllTypesMatch<["result_lhs", "result_rhs"]>` to `Xls_PartialProdOp`, resulted
in the following:

- For `SmulpOp`, `inferReturnTypes` was both declared and implemented.
- For `umulpOp`, `inferReturnTypes` was only declared.

This makes little sense, because no result types should be inferable. Also,
there is no reason for the two ops to behave differently: Their ODS is
perfectly identical except for name and description.

I tracked this down to a bug that triggers some UB inside the MLIR TableGen 
backend. This explains the non-deterministic behaviour.

See upstream PR here:
https://github.com/llvm/llvm-project/pull/122717

In short: `AllTypesMatch` fell apart with multiple result constraints, because 
of incorrect indexing (result index vs combined arg&result index).

Do you want to keep this as-is, or wait for upstream to propagate down and use `AllTypesMatch`?


@jmolloy @jpienaar

schilkp:
LLVM fix merged upstream. This should not be merged until an LLVM version with this is included is pinned: https://github.com/llvm/llvm-project/commit/7aebacbee965a83c5cc69f6a723605436651672c

xls/contrib/mlir/IR/xls_ops.td
schilkp: You are correct.

Here I was mirroring the behavior of all other ops that produce single-bit results (eq,  ne, sge, ...) with the exception of them providing a builder with an I1 return type default.

I guess for the sake of  your scalarize-setup I can't just have a return type of I1.

Do you prefer a simple verifier for this, or a new `Xls_Bit` type (that is either I1 or a tensor thereof) which then could also be used for better typing of the comparison ops down the road?

schilkp:
Rebased. Required upstream commit is now available.

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -305,22 +374,65 @@ class Xls_ArithBinaryOp<string name, string desc>
...
   let hasFolder = 1;
   let hasCanonicalizeMethod = 1;
 }
+def Xls_UmulpOp : Xls_PartialProdOp<"umulp", [{
+  Unsigned partial product multiply.
+
+  Returns two results such that:
+    result_lhs + result_rhs = lhs * rhs
+
+  The two results have the same type, but are not fully constrained: Any
```
jpienaar, line 427: Nit: lets make this two sentences. I think the type is constrained but the values aren't. The comma makes me think this is type related.

schilkp:
I saw this got the "Reviewing Internally" label. Please hold-off merging this for a bit - I will address the final review above once #1856 is merged.

schilkp:
> I saw this got the "Reviewing Internally" label. Please hold-off merging this for a bit - I will address the final review above once #1856 is merged.

#1856 is merged and this has been updated + rebased.

jmolloy (approved):

PR #1853, opened 2025-01-15
[xls][mlir] Add verification to NextValue node

schilkp:
Checks the following basic invariants:
  - Non-zero number of values/predicates
  - All values have matching type
  - Values type matches return type

@jpienaar @jmolloy

### DIFF
--- a/xls/contrib/mlir/IR/xls_ops.cc
+++ b/xls/contrib/mlir/IR/xls_ops.cc
@@ -822,6 +822,22 @@ LogicalResult SSendOp::verify() {
                                         getData().getType());
 }
 
+LogicalResult NextValueOp::verify() {
+  auto types = getValues().getTypes();
+  if (types.empty()) {
+    return emitOpError()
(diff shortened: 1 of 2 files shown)


xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -718,6 +718,22 @@ LogicalResult SSendOp::verify() {
...
+           << "at least one type-predicate tuple must be present";
+  }
+  if (!llvm::all_of(types, [&](Type elem) { return elem == types.front(); })) {
+    return emitOpError() << "all input values must have the same type";
+  }
+  if (types.front() != getResult().getType()) {
+    return emitOpError()
+           << "the type of the input values and return type must match";
+  }
+  return llvm::success();
```
jmolloy, line 734: turbo-nit: use mlir::success() as we're using MLIR not LLVM.

jmolloy (approved):

PR #1865, opened 2025-01-17
Fix fuzz tests relying on zip binary in docker image build.

### LINKED ISSUE #1862: Docker build fails tests
**Describe the bug**
Docker build fails tests. This means it's unusable since the test step is the build step so since the step doesn't complete I don't even get a semi-functional xls Docker image to run

**To Reproduce**
As described in README: `docker build . -f Dockerfile-ubuntu-22.04` in a fresh checkout, that was 9fea976f in my case

**Expected behavior**
Successful build

**Screenshots**
```
3625.2 Executed 1661 out of 1661 tests: 1652 tests pass and 9 fail locally.
3625.2 There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
------
Dockerfile-ubuntu-22.04:41
--------------------
  39 |     # Test everything (opt), exclude xlscc for now due to increased build time when
  40 |     # we add Clang.
  41 | >>> RUN bazel test -c opt -- //xls/... -//xls/contrib/xlscc/...
  42 |
--------------------
ERROR: failed to solve: process "/bin/sh -c bazel test -c opt -- //xls/... -//xls/contrib/xlscc/..." did not complete successfully: exit code: 3
```

 - OS: NixOS 25.05beta733374.8f3e1f807051 (Warbler) x86_64
 - But like this is Docker so that shouldn't matter at all

cc @QuantamHD

cdleary:
Primarily adds "zip" to the installation -- we were relying on this somehow as a system dependency for our xls/fuzzer test targets, I didn't dig into exactly how.

Add common flags and wildcards I use when doing project wide testing.

Fixes #1862

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -30,7 +30,7 @@ jobs:
           restore-keys: bazel-cache-nightly-${{ runner.os }}-
 
       - name: Install dependencies via apt
-        run: sudo apt-get install python3-distutils python3-dev python-is-python3 libtinfo5  build-essential liblapack-dev libblas-dev gfortran
+        run: sudo apt-get install python3-dev python-is-python3 build-essential liblapack-dev libblas-dev gfortran
 
       - name: Bazel Build Tools (opt)
         run: |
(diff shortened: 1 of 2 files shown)


cdleary:
Also had to do the same fix for the github actions workflow I did in one of the other outstanding PRs.

meheff (approved):

PR #1866, opened 2025-01-17
[xls][mlir] Add LiteralOp to XLS MLIR Dialect.

schilkp:
This operation constructs a literal of arbitrarily complex type using a body region that construct this value using basic scalar operations.

It more closely resembles the XLS literal IR node for loss-less conversion between the IR and MLIR, but is not intended to be used during complex operations and transforms in MLIR due to the lack of constant folding.

A pass to remove this operation by inlining (and potentially a pass to reconstruct such literal ops by grouping ops) will be added in the future.

Only a small subset of XLS operations is permitted inside the body region. The new LiteralMemberOpInterface is used to identify such ops.

@jpienaar @jmolloy

### DIFF
--- a/xls/contrib/mlir/IR/interfaces.td
+++ b/xls/contrib/mlir/IR/interfaces.td
@@ -78,5 +78,23 @@ def PredicatableOpInterface : OpInterface<"PredicatableOpInterface"> {
   ];
 }
 
+// Declares that an op is permissible inside a literal region.
+def LiteralMemberOpInterface : OpInterface<"LiteralMemberOpInterface"> {
+  let description = [{
+    Declares that an op is permissible inside a literal region.
+  }];
+  let cppNamespace = "::mlir::xls";
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns true if the operation, in its current form, is permissible inside a literal region.",
+      /*retTy=*/"bool",
+      /*methodName=*/"isPermissibleInsideLiteral",
+      /*args=*/(ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/"return true;"
+    >,
+  ];
+}
 
 #endif
--- a/xls/contrib/mlir/IR/xls_ops.cc
+++ b/xls/contrib/mlir/IR/xls_ops.cc
@@ -484,6 +484,33 @@ Type getI1TypeOf(Type type) {
 
 namespace mlir::xls {
 
+LogicalResult LiteralOp::verifyRegions() {
+  Block& b = getInitializerBlock();
+
+  // Verify return op & typing:
+  auto ret = dyn_cast<YieldOp>(b.getTerminator());
+  if (!ret) {
+    return emitOpError(
+        "initializer region must be terminated by an xls.yield op");
+  }
+  if (ret->getNumOperands() != 1) {
+    return emitOpError("initializer region must return exactly one value");
+  }
+  if (ret->getOperand(0).getType() != getType()) {
+    return emitOpError("initializer region type ")
+           << ret->getOperand(0).getType() << " does not match literal type "
+           << getType();
+  }
+  // Verify that operations inside are allowed inside a literal region:
+  for (Operation& op : b) {
+    auto iface = dyn_cast<LiteralMemberOpInterface>(op);
+    if (!iface || !iface.isPermissibleInsideLiteral()) {
+      return op.emitError() << "op not permissible inside literal region";
+    }
+  }
+  return success();
+}
+
 LogicalResult TupleOp::inferReturnTypes(
     MLIRContext* context, std::optional<Location> location, ValueRange operands,
     DictionaryAttr attributes, OpaqueProperties properties, RegionRange regions,
--- a/xls/contrib/mlir/IR/xls_ops.td
+++ b/xls/contrib/mlir/IR/xls_ops.td
@@ -562,7 +562,12 @@ def Xls_SendOp : Xls_Op<"send", [
 // Array operations
 //===----------------------------------------------------------------------===//
 
-def Xls_ArrayOp : Xls_NonScalarizableNaryOp<"array", [Pure, SameTypeOperands, TensorArrayTypeFungible]> {
+def Xls_ArrayOp : Xls_NonScalarizableNaryOp<"array", [
+  Pure,
+  SameTypeOperands,
+  TensorArrayTypeFungible,
+  DeclareOpInterfaceMethods<LiteralMemberOpInterface>
+]> {
   let summary = "Array creation operation";
   let description = [{
     Constructs an array of its operands.
@@ -721,8 +726,12 @@ def Xls_ArrayConcatOp : Xls_Op<"array_concat", [Pure, TensorArrayTypeFungible]> 
 // Tuple operations
 //===----------------------------------------------------------------------===//
 
-def Xls_TupleOp : Xls_Op<"tuple", [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface>,
-                                   TensorArrayTypeFungible]> {
+def Xls_TupleOp : Xls_Op<"tuple", [
+  Pure,
+  DeclareOpInterfaceMethods<InferTypeOpInterface>,
+  DeclareOpInterfaceMethods<LiteralMemberOpInterface>,
+  TensorArrayTypeFungible,
+]> {
   let summary = "Tuple operation";
   let description = [{
     Constructs a tuple of its operands.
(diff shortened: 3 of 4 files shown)


xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -390,6 +390,32 @@ Type getI1TypeOf(Type type) {
 
 namespace mlir::xls {
 
+LogicalResult LiteralOp::verifyRegions() {
+  Block& b = getInitializerBlock();
+  // Verify return op & typing:
+  auto ret = dyn_cast<YieldOp>(b.getTerminator());
+  if (!ret) {
```
jpienaar, line 397: Shouldn't this hold due to the verification already generated? (e.g., `SingleBlockImplicitTerminator<"YieldOp">`).

xls/contrib/mlir/IR/interfaces.td
```diff
@@ -78,5 +78,23 @@ def PredicatableOpInterface : OpInterface<"PredicatableOpInterface"> {
   ];
 }
 
+// Declares that an op is permissible inside a literal region.
+def LiteralMemberOpInterface : OpInterface<"LiteralMemberOpInterface"> {
```
jpienaar, line 82: Why not make this a trait and only add when true?

xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -390,6 +390,32 @@ Type getI1TypeOf(Type type) {
...
+  if (ret->operand_type_begin() == ret.operand_type_end()) {
+    return emitOpError("initializer region cannot return void");
+  }
+  if (*(ret->operand_type_begin()) != getType()) {
+    return emitOpError("initializer region type ")
+           << *ret.operand_type_begin() << " does not match literal type "
+           << getType();
+  }
+  // Verify that operations inside are allowed inside a literal region:
+  for (Operation& op : b) {
```
jpienaar, line 410: If you make it a trait, you should be able to use llvm all\_of kind of check here with trait checking.

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -1226,6 +1255,55 @@ def Xls_ConstantScalarOp : Xls_Op<"constant_scalar", [ConstantLike, Pure]> {
...
+]> {
+  let summary = "Complex Literal";
+  let description = [{
+    Produces a complex (non-bits) constant value from a region containing
+    simple constants or tuple/array construction operations.
+
+    The body must be terminated by a `xls.yield` op, which returns the
+    constructed value.
+
+    Besides this yield, only the following operations are permissible inside
```
jpienaar, line 1272: I think it would be computationally cheaper to just have a helper function with the TypeIDs of these ops rather than using traits (you'd be doing pointer comparison based on name rather than doing a linear scan per op through trait list). If these are this well known and we don't need to support any other dialects/unknown supported op types.

xls/contrib/mlir/IR/interfaces.td
schilkp: I considered this, but as it stands right now the decision for if an op is permissible inside a literal is dynamic:

XLS IR allows for token literals:
```
literal.1: token = literal(value=token, id=1)
```

According to Eric this is equivalent to generating a new token with an empty `join()` in DSLX, resp. `after_all` in XLS IR without literals. I am assuming the reason this is supported is because it allows the literal IR op to construct complex types with tokens inside.

James and I agreed that for the MLIR side we would not introduce an extra op for this, and simply allow an `xls.after_all` with empty operands inside literal regions. This, in turn, means that for `AfterAllOp`, this can't really be a trait and needs some (simple) logic:

```tablegen
def Xls_AfterAllOp : Xls_Op<"after_all", [
  Pure,
  DeclareOpInterfaceMethods<LiteralMemberOpInterface>
]> {
  // ....
  let extraClassDeclaration = [{
    bool isPermissibleInsideLiteral() {
      // xls.after_all is only permissible inside literal regions if it is
      // used to create a new token with no prior dependencies.
      return getOperands().empty();
    }
  }];
}
```

Hence I went with the interface because it still allowed me to have this operation property included in a (halfway) declarative fashion in the ODS.

If you prefer, I can of course refactor this into a helper func which checks op IDs + checks `after_all` for an empty operand list.

Let me know :)

jmolloy: That rationale makes sense to me.

xls/contrib/mlir/IR/interfaces.td
```diff
@@ -78,5 +78,23 @@ def PredicatableOpInterface : OpInterface<"PredicatableOpInterface"> {
...
+  let description = [{
+    Declares that an op is permissible inside a literal region.
+  }];
+  let cppNamespace = "::mlir::xls";
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns true if the operation, in its current form, is permissibe inside a literal region.",
```
jmolloy, line 90: nit: permissable

jmolloy: ... permissible, obviously *facepalm*

xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -390,6 +390,28 @@ Type getI1TypeOf(Type type) {
 
 namespace mlir::xls {
 
+LogicalResult LiteralOp::verifyRegions() {
+  Block& b = getInitializerBlock();
+  // Verify return op & typing:
+  auto ret = cast<YieldOp>(b.getTerminator());
```
jmolloy, line 396: You shouldn't assume the terminator is a yield in a verifier (verify it, don't assert it).

xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -390,6 +390,28 @@ Type getI1TypeOf(Type type) {
...
+  if (ret->operand_type_begin() == ret.operand_type_end()) {
```
jmolloy, line 397: Can you use getNumOperands() == 0 instead here to be a bit clearer?

xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -390,6 +390,28 @@ Type getI1TypeOf(Type type) {
...
+    return emitOpError("initializer region cannot return void");
+  }
+  if (*(ret->operand_type_begin()) != getType()) {
```
jmolloy, line 400: you're assuming only one result here but you haven't checked it (only that there are > 0 results).

Also can you just use if (ret.getOperand().getType() != getType()) ?

jmolloy: sorry, it's a YieldOp so it may have multiple operands, so you need ret.getOperand(0).

xls/contrib/mlir/IR/xls_ops.td
jmolloy: Not an unreasonable suggestion, but personally I like that this is all having tablegen as the source of truth, and I'm not majorly worried about linear scans of literal regions in the verifier (they'll be small).

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -1226,6 +1255,55 @@ def Xls_ConstantScalarOp : Xls_Op<"constant_scalar", [ConstantLike, Pure]> {
...
+    SizedRegion<1>:$body
+  );
+  let results = (outs
+    Xls_BitsOrTuple:$result
+  );
+  let hasRegionVerifier = 1;
+  let extraClassDeclaration = [{
+    /// Return the initializer region. This must always include an initializer
+    /// block terminated by a `xls.yield`.
+    Region &getInitializerRegion() {
```
jmolloy, line 1293: There are autogenerated functions like this that tblgen provides, but they are keyed on the name of the region in your `regions` list.

So if you rename $body to, say, $initializer, you should get approximately these functions generated for you.

jmolloy (changes requested):
Looks great!

schilkp:
Rebased.

xls/contrib/mlir/IR/xls_ops.td
schilkp: I renamed to `$initializer` which gets me the `getInitializer` func that returns the region. 

My understanding is that `SizedRegion<1>: $initializer` should constrain to exactly one block, but `getInitializer` returns only the region and there is no other block getter. 

For that reason, I kept the `getInitializerBlock` because that should never fail (there is always a block -> its always safe to get it).

Am I wrong here? I remember tablegen generating block accessors so I am not sure if I am missing something or found another tblgen edge case or am overlooking smth.

jmolloy: You've got the SingleBlockImplicitTerminator trait, so you've done everything you can. LGTM.

jmolloy (approved):

PR #1869, opened 2025-01-19
[DSLX] Get quickchecks of enums to work properly.

### LINKED ISSUE #1586: Quickcheck falsifying example dump falls over with implicit tokens
**Describe the bug**
If the underlying function being quickchecked has something like an `assert`, then an implicit token and activation bit get created in the IR which appears to be what is used for `DoQuickCheck`. If the quick check fails, then it tries to generate the falsifying example. This currently has two problems:

1. It errors out on `ValueToInterpValue` for tokens: 
    ```
    E0904 15:06:28.061705    2490 run_routines.cc:106] Internal error: INVALID_ARGUMENT: Cannot convert IR value to 
    interpreter value: token
    === Source Location Trace: === xls/dslx/interp_value_utils.cc:329
    xls/dslx/run_routines/run_routines.cc:368
    ```
2. If you try to fix that by converting tokens explicitly, then you run into the fact that the DSLX function and the IR function don't have the same # of arguments. The `params` vector is gathered from the DSLX function type info, while the `last_argset` vector is from the evaluated IR function.

**To Reproduce**
Run the following quickcheck, which correctly fails because of how`upcast` handles subnormals/nans.

```
import bfloat16;
import float32;

#[quickcheck]
fn bfloat16_bits_to_float32_bits_upcast_is_zero_pad(x: bits[bfloat16::BF16_TOTAL_SZ]) -> bool {
    (x ++ bits[u32:16]:0 ==
    float32::flatten(
        apfloat::upcast<float32::F32_EXP_SZ, float32::F32_FRACTION_SZ>(bfloat16::unflatten(x))))
}
```

**Expected behavior**
The quickcheck should properly report the falsifying sample when found (e.g. subnormal is flushed to 0 by upcast but not by the zero padding).

### LINKED ISSUE #1712: Quickcheck falsifying example dump falls over for enums
**Describe the bug**
If the input argument is an `enum`, then when the falsifying example is reported, it segfaults on ValueToInterpValue 

```
[ RUN QUICKCHECK        ] blue_is_always_favorite count: 1000
*** SIGSEGV (@0x8), ... received by PID 929 (TID 929) on cpu 29; stack trace: ***
PC: @     0x5570f910052d  (unknown)  xls::dslx::ValueToInterpValue()
    @     0x5570fc25f021       1888  FailureSignalHandler()
    @     0x7fc603949e80  (unknown)  (unknown)
    @     0x5570f910052d        368  xls::dslx::ValueToInterpValue()
    @     0x5570f8fbd40b       2096  xls::dslx::AbstractTestRunner::ParseAndTest()
    @     0x5570f8f85cd9       1232  main
    @     0x7fc6037da3d4        192  __libc_start_main
    @     0x5570f8f8402a  (unknown)  _start
E1113 19:29:42.758104     929 process_state.cc:806] RAW: Raising signal 11 with default behavior
```

**To Reproduce**

```
enum Color : u2 {
    White = u2:0,
    Red = u2:1,
    Green = u2:2,
    Blue = u2:3,
}

fn favorite_color(color: Color) -> Color { Color::Blue }

#[quickcheck]
fn blue_is_always_favorite(color: Color) -> bool { favorite_color(color) == Color::White }
```

**Expected behavior**
The quickcheck should properly report the falsifying sample when found.

As a workaround, you can use `u2` as the input type for the quickcheck arg, and then cast to the enum, i.e.

```
enum Color : u2 {
    White = u2:0,
    Red = u2:1,
    Green = u2:2,
    Blue = u2:3,
}

fn favorite_color(color: Color) -> Color { Color::Blue }

#[quickcheck]
fn blue_is_always_favorite(color: u2) -> bool { favorite_color(color as Color) == Color::White }
```

will properly report:

FailureError: The program being interpreted failed! Found falsifying example after 1 tests: [u2:2]

cdleary:
Also support/test quickchecks of functions that use the implicit-token calling convention.

Note that the IsKnownU1 is duplicated from the other PR exactly in the same spot so it should be easy to merge.

Fixes #1712
Also fixes #1586 and confirms with a test for the reported case


meheff (approved):

xls/dslx/run_routines/run_routines_test.cc
```diff
@@ -190,6 +192,161 @@ fn trivial(x: u2) -> bool { id(true) }
...
+  options.vfs_factory = [kProgram] {
+    return std::make_unique<UniformContentFilesystem>(kProgram, "test.x");
+  };
+  options.run_comparator = &jit_comparator;
+  XLS_ASSERT_OK_AND_ASSIGN(
+      TestResultData result,
+      ParseAndTest(kProgram, kModuleName, kFilename, options));
+  EXPECT_THAT(result, IsTestResult(TestResult::kSomeFailed, 1, 0, 1));
+  ASSERT_EQ(result.failures().size(), 1);
+  std::string_view failure_message = result.failures()[0];
```
allight, line 248: This is a lifetime error. 'result.failures()' returns a vector<std::string> so the string_view is pointing at freed memory.

Please make std::string instead

meheff: This is fixed internally  and should land shortly. No need to update the PR

cdleary: Thanks -- I'll make a followup PR to rename `failures()` since it looks like a span-style getter, lowercase methods aren't supposed to do meaningful work.

cdleary: https://github.com/google/xls/pull/1878

PR #1873, opened 2025-01-22
Add pre-commit check for EOF newlines.

cdleary:
Make them consistent.

As you can see in the github workflow, the mechanism is:

```sh
pip install pre-commit
pre-commit install # If developing locally to run before a commit is created in git
pre-commit run --all-files  # To run checks manually
```


hzeller:
Nice/
Also (probably not this PR) we should probably recommend somewhere in the contributors documentation to enable honoring the[.editorconfig](https://github.com/google/xls/blob/main/.editorconfig) in their editor if they haven't. It should be fairly common these days, but still not everyone has.
It will take care of removing extra whitespace at EOL and ensure having a newline at EOF.

cdleary:
@hzeller It looks like VSCode doesn't natively support `.editorconfig` files (which is what I used when I made the problematic file), looks like via plugin that you have to know to go install.

hzeller:
@cdleary Interesting. I remember I might also had to explicitly use some extra plugin in emacs.
So yes, documenting it is probably good for these reasons.

cdleary:
Looks like this landed in https://github.com/google/xls/commit/e96eb8c85c46c1d9c372d92f7f15d4e0ad434514 so closing the PR

PR #1884, opened 2025-01-24
[dslx_fmt] Support multiple arguments as input, similar to other formatters

cdleary:
This lets us slot into a pre-commit hook more easily, looks like it batches up about 8 files at a time by default to avoid too many subprocess invocations.

### DIFF
--- a/xls/dslx/dslx_fmt.cc
+++ b/xls/dslx/dslx_fmt.cc
@@ -52,7 +52,7 @@ ABSL_FLAG(bool, error_on_changes, false,
 ABSL_FLAG(std::string, dslx_path, "",
           "Additional paths to search for modules (colon delimited).");
 ABSL_FLAG(std::string, mode, "autofmt",
-          "whether to use reflowing auto-formatter");
+          "whether to use reflowing auto-formatter; choices: autoformat|parse");
 ABSL_FLAG(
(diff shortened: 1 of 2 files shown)


xls/dslx/dslx_fmt.cc
```diff
@@ -133,38 +136,87 @@ absl::Status RealMain(std::string_view input_path,
...
+  // - If stdin is the input, it should be the only input.
+  // - If we're *not* doing in-place formatting, there should be only one input
+  // (otherwise things will be all jumbled together in stdout),
+  // - If we error-on-changes or use the opportunistic postcondition, there
+  //   should be only one input, to avoid semantic ambiguity when we're
+  //   side-effecting the formatting of files and when we flag the error.
+  bool has_stdin_arg =
+      std::any_of(input_paths.begin(), input_paths.end(),
+                  [](std::string_view path) { return path == "-"; });
+  if (has_stdin_arg) {
```
dplassgit, line 162: Thoughts on adding tests for each of these cases?

cdleary: I think I've got testing in the .py for the interesting modes and combinations now -- LMK if you see any more that seem like they're missing!

dplassgit (approved):

PR #1886, opened 2025-01-25
[build] Update for fuzztest bazelrc to be excluded from line endings pre-commit.

cdleary:


cdleary:
This is unnecessary now due to https://github.com/google/xls/commit/13a141366409ebfd5619746bbabc33914d5ae202

ericastor:
This might be a better fix than mine! WDYT about landing this & reverting my change in a single PR?

Issue #810, opened 2022-12-16
Make Match Arm Patterns support Array Type


vincent-mirian-google:
Match Arm Patterns do not support array types.

For example, the following does not compile:
```
let x = match x11 {
  u32[2]:[0, 1] => y,
  _=> y+u32:77
};
```

PR #1905, opened 2025-02-03
[DSLX:TS] Exhaustiveness analysis. See `dslx_pattern_exhaustiveness.md`

### LINKED ISSUE #204: Exhaustiveness checking for pattern matches
Please add exhaustiveness checking. Why?

### To prevent silly errors like the following, where I believe the first two match arms are actually doing something useful. 

Exhaustiveness checking would notice that `Green` was binding to `v`, instead of being matched as the enumeration value. Then I would be warned that the `Red` arm is doing nothing. (note that #203 would also help me realize I had an error in my code)

```
enum Color : u1 {
  Red = 0,
  Green = 1
}

fn match_color(v : Color) -> u32 {
    match v {
      Green => u32: 0,
      Red => u32: 1,
  }
}

#![test]
fn t() {
  let _ = assert_eq(match_color(Color::Green), u32: 0);
  let _ = assert_eq(match_color(Color::Red), u32: 1);
  ()
}
```

### To avoid code adding redundant irrefutable matches. 

I would like to write code like this:

```
enum Color : u1 {
  Red = 0,
  Green = 1
}

fn match_color(v : Color) -> u32 {
    match v {
      Color::Green => u32: 0,
      Color::Red => u32: 1,
  }
}

#![test]
fn t() {
  let _ = assert_eq(match_color(Color::Green), u32: 0);
  let _ = assert_eq(match_color(Color::Red), u32: 1);
  ()
}
```

but I can't, so I'm forced to do something like this:

```
enum Color : u1 {
  Red = 0,
  Green = 1
}

fn match_color(v : Color) -> u32 {
    match v {
      Color::Green => u32: 0,
      Color::Red => u32: 1,
      _ => u32: 1,
  }
}

#![test]
fn t() {
  let _ = assert_eq(match_color(Color::Green), u32: 0);
  let _ = assert_eq(match_color(Color::Red), u32: 1);
  ()
}
```

Why are redundant irrefutable matches bad? 

Well, consider what happens when, in the future, I add `Blue` to `Color`. In my first code listing, it would immediate trigger a warning/error that the match was not exhaustive, and I could update the code to handle Blue. However, the second code listing is what I'm forced to write, and the compiler will not give me any warning that the match wasn't updated to handle Blue.

That is, this will compile with no warnings, even though I would have liked a warning, and would have gotten one had I not been forced to add a `_ =>` case

```
enum Color : u2 {
  Red = 0,
  Green = 1,
  Blue = 2
}

fn match_color(v : Color) -> u32 {
    match v {
      Color::Green => u32: 0,
      Color::Red => u32: 1,
      _ => u32: 1,
  }
}

#![test]
fn t() {
  let _ = assert_eq(match_color(Color::Green), u32: 0);
  let _ = assert_eq(match_color(Color::Red), u32: 1);
  ()
}
```

cdleary:
Warning is disabled by default for now (so large code bases can adapt before enabling by default) -- see `warning_kind.h`

Note: this is stacked on top of #1902 -- will rebase it when that has landed.

Related: #810 (for the lack of array patterns)

Fixes #204


mikex-oss:
FYI I see some "Match is already exhaustive before this pattern" failures in ZSTD, even though I think the PR says these should be disabled by default?

cdleary:
> FYI I see some "Match is already exhaustive before this pattern" failures in ZSTD, even though I think the PR says these should be disabled by default?

Thanks, sorry for the oversight -- looks like `ir_converter_main` doesn't start with the default warnings set we have available as a global constant, will fix and then update the PR.

cdleary:
@mikex-oss Ok, should be good to go now, added a test for flag enabledness and shared between `interpreter_main` and `ir_converter_main` :+1:

PR #1915, opened 2025-02-07
[c_api] Public API for DSLX to IR w/ warning knobs.

cdleary:
Keeps the prior C APIs but adds the versions that take warning parameters as well.

Note that the /public C++ API/ is changed so that we can consolidate a bit vs having APIs with tons of parameters on that side of the equation -- we're not going for a flat "C ABI style" design there. This will require a global check on import.

Adds:
- `xls_convert_dslx_path_to_ir_with_warnings` to supplement `xls_convert_dslx_path_to_ir`
- `xls_convert_dslx_to_ir_with_warnings` to supplement `xls_convert_dslx_to_ir`
- `xls_c_strs_free` to free the dynamically sized warnings array that can be returned from the API

### DIFF
--- a/xls/public/c_api.cc
+++ b/xls/public/c_api.cc
@@ -58,22 +58,104 @@ void xls_init_xls(const char* usage, int argc, char* argv[]) {
   (void)(xls::InitXls(usage, argc, argv));
 }
 
-bool xls_convert_dslx_to_ir(const char* dslx, const char* path,
-                            const char* module_name,
-                            const char* dslx_stdlib_path,
-                            const char* additional_search_paths[],
-                            size_t additional_search_paths_count,
-                            char** error_out, char** ir_out) {
+bool xls_convert_dslx_to_ir_with_warnings(
+    const char* dslx, const char* path, const char* module_name,
+    const char* dslx_stdlib_path, const char* additional_search_paths[],
+    size_t additional_search_paths_count, const char* enable_warnings[],
+    size_t enable_warnings_count, const char* disable_warnings[],
+    size_t disable_warnings_count, bool warnings_as_errors,
+    char*** warnings_out, size_t* warnings_out_count, char** error_out,
+    char** ir_out) {
   CHECK(dslx != nullptr);
   CHECK(path != nullptr);
   CHECK(dslx_stdlib_path != nullptr);
   CHECK(error_out != nullptr);
 
   std::vector<std::filesystem::path> additional_search_paths_cpp =
-      xls::ToCpp(additional_search_paths, additional_search_paths_count);
+      xls::ToCppPaths(additional_search_paths, additional_search_paths_count);
+  std::vector<std::string_view> enable_warnings_cpp =
+      xls::ToCppStringViews(enable_warnings, enable_warnings_count);
+  std::vector<std::string_view> disable_warnings_cpp =
+      xls::ToCppStringViews(disable_warnings, disable_warnings_count);
 
-  absl::StatusOr<std::string> result = xls::ConvertDslxToIr(
-      dslx, path, module_name, dslx_stdlib_path, additional_search_paths_cpp);
+  std::vector<std::string> warnings_out_cpp;
+
+  const xls::ConvertDslxToIrOptions options = {
+      .dslx_stdlib_path = dslx_stdlib_path,
+      .additional_search_paths = additional_search_paths_cpp,
+      .enable_warnings = enable_warnings_cpp,
+      .disable_warnings = disable_warnings_cpp,
+      .warnings_as_errors = warnings_as_errors,
+      .warnings_out = &warnings_out_cpp,
+  };
+
+  absl::StatusOr<std::string> result =
+      xls::ConvertDslxToIr(dslx, path, module_name, options);
+
+  if (warnings_out != nullptr) {
+    xls::ToOwnedCStrings(warnings_out_cpp, warnings_out, warnings_out_count);
+  } else if (warnings_out_count != nullptr) {
+    *warnings_out_count = warnings_out_cpp.size();
+  }
+
+  return xls::ReturnStringHelper(result, error_out, ir_out);
+}
+
+bool xls_convert_dslx_to_ir(const char* dslx, const char* path,
+                            const char* module_name,
+                            const char* dslx_stdlib_path,
+                            const char* additional_search_paths[],
+                            size_t additional_search_paths_count,
+                            char** error_out, char** ir_out) {
+  const char* enable_warnings[] = {};
+  const char* disable_warnings[] = {};
+  return xls_convert_dslx_to_ir_with_warnings(
+      dslx, path, module_name, dslx_stdlib_path, additional_search_paths,
+      additional_search_paths_count, enable_warnings, 0, disable_warnings, 0,
+      /*warnings_as_errors=*/false,
+      /*warnings_out=*/nullptr,
+      /*warnings_out_count=*/nullptr, error_out, ir_out);
+}
+
+bool xls_convert_dslx_path_to_ir_with_warnings(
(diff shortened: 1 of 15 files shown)


xls/public/c_api.h
```diff
@@ -60,11 +60,34 @@ bool xls_convert_dslx_to_ir(const char* dslx, const char* path,
...
+// As above, but also takes `enable_warnings` and `disable_warnings` which
+// are arrays of warning names to enable or disable respectively against the
+// default warning set.
+//
+// Note: `warnings_out` should be freed via `xls_c_strs_free`.
+bool xls_convert_dslx_to_ir_with_warnings(
+    const char* dslx, const char* path, const char* module_name,
+    const char* dslx_stdlib_path, const char* additional_search_paths[],
+    size_t additional_search_paths_count, const char* enable_warnings[],
+    size_t enable_warnings_count, const char* disable_warnings[],
```
grebe, line 72: The argument count seems to be getting pretty high, how would you feel about merging all the warnings into structs? It seems we only have opaque structs, but we don't make any ABI stability promises and something like the following seems like it would help make it reasonable to use the API, right?

```
struct string_view_span {
  const char** views;
  size_t size;
};

struct warnings {
  struct string_view_span enable_warnings;
  struct string_view_span disable_warnings;
};
```

I'm not sure what all the considerations in using the C API are- is struct packing more obnoxious than it's worth?

xls/public/c_api_impl_helpers.h
```diff
@@ -46,10 +46,20 @@ bool ReturnStringHelper(absl::StatusOr<std::string>& to_return,
...
+
 // Converts the C representation of filesystem search paths into a C++
 // representation.
-std::vector<std::filesystem::path> ToCpp(const char* additional_search_paths[],
-                                         size_t additional_search_paths_count);
+std::vector<std::filesystem::path> ToCppPaths(const char* paths[],
+                                              size_t paths_count);
+
+// Converts the C representation of warnings into a C++ representation.
+std::vector<std::string_view> ToCppStrings(const char* strings[],
```
grebe, line 61: Nit: the return type makes it clear, but maybe better to be named `ToCppStringViews()`?

xls/public/c_api_impl_helpers.cc
```diff
@@ -41,10 +42,29 @@ std::vector<std::filesystem::path> ToCpp(const char* additional_search_paths[],
...
+  return strings_cpp;
+}
+
 char* ToOwnedCString(std::string_view s) { return strndup(s.data(), s.size()); }
 
-// Helper function that we can use to adapt to the common C API pattern when
-// we're returning an `absl::StatusOr<std::string>` value.
+void ToOwnedCStrings(absl::Span<const std::string> cpp, char*** c_out,
+                     size_t* c_out_count) {
+  *c_out = new char*[cpp.size()];
```
grebe, line 61: `new char*[0]` is a weird thing to do, right? Maybe `if (cpp.empty()) { *c_out_count = 0; return; }` would be safe? Maybe also `*c_out = nullptr;`?

I'm pretty sure this shouldn't happen, but seems good to cover it regardless.

xls/public/c_api.cc
```diff
@@ -58,39 +58,116 @@ void xls_init_xls(const char* usage, int argc, char* argv[]) {
...
+      .dslx_stdlib_path = dslx_stdlib_path,
+      .additional_search_paths = additional_search_paths_cpp,
+      .enable_warnings = enable_warnings_cpp,
+      .disable_warnings = disable_warnings_cpp,
+      .warnings_as_errors = warnings_as_errors,
+      .warnings_out = &warnings_out_cpp,
+  };
+  absl::StatusOr<std::string> result = xls::ConvertDslxPathToIr(path, options);
+
+  if (warnings_out != nullptr && !warnings_out_cpp.empty()) {
```
grebe, line 150: See other comment, but maybe if `ToOwnedCStrings()` handles empty vector it's a little more clear to just have `if (warnings_out != nullptr) { ToOwnedCString(...); }`

grebe (changes requested):

xls/public/c_api.h
cdleary: I just wanted to make potential for the ABI considerations minimal (including passing structs by value) by making them flat. If we passed the struct instead by pointer we'd need a more complicated invocation sequence.

In an abstract sense these C APIs are supposed to be fairly unchanging in that we probably make one with a different name if we need different functionality. Probably we won't keep strict rules just to make sure the project can continue evolving without undue burden, but that's roughly the tradeoff space I've been trying to hit for the C APIs.

xls/public/c_api_impl_helpers.cc
cdleary: It's a /valid/ thing to do if that's the concern. The challenge/tradeoff here is whether users could think that returning null for an empty array is an indication of failure of some sort. If we were being paranoid about what our API was signaling we'd probably want to give back a non-null value for the array even in the zero element case. The way it's written here it's a uniform behavior path even if there are no warnings -- we always allocate and we always deallocate. (It may have less performance, because why make a useless zero size allocation for the case where we have no warnings, but it makes our expectations as caller clearer.)

LMK what you think about that tradeoff.

xls/public/c_api.cc
cdleary: Yeah, good point, this is not consistent with the "I will always populate the pointer with something non-null if you provide it" that I was hypothesizing above. If I remove the `empty()` guard here I think the case I make above for what we might want to do paranoia wise makes more sense.

cdleary: Okay I made it so that when it's zero count it always gives back nullptr, PTAL if that seems reasonable @grebe . I put a comment on the C API that says in the case of zero warnings the `warnings_out` can be populated with nullptr.

xls/public/c_api_impl_helpers.cc
cdleary: Resolving this one in favor of the other thread, I changed it up from my original hypothesis, now we'll avoid the useless zero sized allocation and just try to document it clearly for the caller that it can be nullptr they get back.

xls/public/c_api_impl_helpers.h
cdleary: Done.

xls/public/c_api.h
grebe: Yeah, that makes sense.

xls/public/c_api.cc
```diff
@@ -58,39 +58,119 @@ void xls_init_xls(const char* usage, int argc, char* argv[]) {
...
+      .enable_warnings = enable_warnings_cpp,
+      .disable_warnings = disable_warnings_cpp,
+      .warnings_as_errors = warnings_as_errors,
+      .warnings_out = &warnings_out_cpp,
+  };
+
+  absl::StatusOr<std::string> result =
+      xls::ConvertDslxToIr(dslx, path, module_name, options);
+
+  if (warnings_out != nullptr && !warnings_out_cpp.empty()) {
```
grebe, line 95: Should the `empty()` check be dropped here?

xls/public/c_api.cc
grebe: Sounds good to me, maybe a similar comment about the length-zero case should be added to `ToOwnedCStrings()`.

grebe (approved):

xls/public/c_api.cc
cdleary: Cool thanks, done.

xls/public/c_api.cc
cdleary: Good catch, thanks. Done.

cdleary:
All comments addressed now, thanks.

grebe (approved):

PR #1916, opened 2025-02-07
[xls][mlir] Support XLS fifo properties/config in MLIR channels.

schilkp:
As discussed via email.

Support XLS fifo properties/config in MLIR channels.

This also adds support for converting these properties to/from XLS, and updates different passes (array_to_bits, index_type_conversion) to ensure these attributes are maintained when channels are modified.

Finally, these properties are also exposed for sproc channels, and the proc elaboration pass is updated to assign said properties to all eproc channels that are generated from an sproc channel.

To be discussed: Are the input/output flopping fifo configuration values required?

### DIFF
--- a/xls/contrib/mlir/IR/xls_ops.cc
+++ b/xls/contrib/mlir/IR/xls_ops.cc
@@ -620,7 +620,9 @@ ParseResult SchanOp::parse(OpAsmParser& parser, OperationState& result) {
   result.addAttribute("type", TypeAttr::get(type));
   result.types.push_back(SchanType::get(parser.getContext(), type, false));
   result.types.push_back(SchanType::get(parser.getContext(), type, true));
-  return success();
+
+  return failure(
+      parser.parseOptionalAttrDictWithKeyword(result.attributes).failed());
 }
 
 void SchanOp::print(OpAsmPrinter& printer) {
@@ -630,6 +632,9 @@ void SchanOp::print(OpAsmPrinter& printer) {
   printer << '(';
   printer.printString(getName());
   printer << ')';
+  printer << ") ";
+  printer.printOptionalAttrDictWithKeyword(getOperation()->getAttrs(),
+                                           {"type", "name"});
 }
 
 void SprocOp::print(OpAsmPrinter& printer) {
--- a/xls/contrib/mlir/IR/xls_ops.td
+++ b/xls/contrib/mlir/IR/xls_ops.td
@@ -135,6 +135,75 @@ def Xls_TranslationLinkage : Xls_Attr<"TranslationLinkage"> {
   let attrName = "xls.translation_linkage";
 }
 
+def Xls_FlopKindNone        : I32EnumAttrCase<"kNone", 0, "none"> {
+  let summary = "No flop.";
+  let description = [{
+    Adds no buffer to the input/output.
+
+    See https://google.github.io/xls/codegen_options/#io-behavior for more information.
+  }];
+}
+def Xls_FlopKindFlop        : I32EnumAttrCase<"kFlop", 1, "flop"> {
+  let summary = "Basic flop.";
+  let description = [{
+    Adds a pipeline stage to the input/output. This is essentially a single-element FIFO
+
+    See https://google.github.io/xls/codegen_options/#io-behavior for more information.
+  }];
+}
+def Xls_FlopKindSkid        : I32EnumAttrCase<"kSkid", 2, "skid"> {
+  let summary = "Skip flop.";
+  let description = [{
+    Adds a skid buffer to the input/output. The skid buffer can hold 2 entries.
+
+    See https://google.github.io/xls/codegen_options/#io-behavior for more information.
+  }];
+}
+def Xls_FlopKindZeroLatency : I32EnumAttrCase<"kZeroLatency", 3, "zero-latency"> {
+  let summary = "Zero-latency flop.";
+  let description = [{
+    Adds a zero-latency buffer to the input/output. This is essentially a
+    single-element FIFO with bypass.
+
+    See https://google.github.io/xls/codegen_options/#io-behavior for more information.
+  }];
+}
+def Xls_FlopKind : I32EnumAttr<"FlopKind", "XLS flop kind",
+  [Xls_FlopKindNone, Xls_FlopKindFlop, Xls_FlopKindSkid, Xls_FlopKindZeroLatency]> {
+  let genSpecializedAttr = 0;
+  let cppNamespace = "::mlir::xls";
+}
+def Xls_FlopKindAttr : EnumAttr<Xls_Dialect, Xls_FlopKind, "flop_kind"> {
+  let summary = "Channel flop/buffer kind.";
+  let description = [{
+    The inputs/outputs of procs/blocks can optionally be fitted with one of a
+    number of buffer types. This is specified on the channel connected to this
+    proc input/output, using the `input_flop_kind` and `output_flop_kind`
+    property.
+
+    See https://google.github.io/xls/codegen_options/#io-behavior for more information.
+  }];
+}
+
+def Xls_FifoConfig : Xls_Attr<"FifoConfig"> {
+  let summary = "FIFO Configuration Attribute";
+  let description = [{
+    Definies the properties of the physical FIFO to be instantiated from
+    an (internal) channel.
+  }];
+
+  let parameters = (ins
+    "int64_t":$fifo_depth,
+    "bool":$bypass,
+    "bool":$register_push_outputs,
+    "bool":$register_pop_outputs
+  );
+
+  let mnemonic = "fifo_config";
+
+  let assemblyFormat = "`<` struct(params) `>`";
+}
+
 class GetShapeSplat<string name> :
     StrFunc<"getShapeSplat($" # name # ".getType())">;
 
@@ -1626,6 +1695,9 @@ def Xls_ChanOp : Xls_Op<"chan", [
   let arguments = (ins
     SymbolNameAttr:$sym_name,
     TypeAttr:$type,
+    OptionalAttr<Xls_FifoConfig>: $fifo_config,
+    OptionalAttr<Xls_FlopKindAttr>: $input_flop_kind,
+    OptionalAttr<Xls_FlopKindAttr>: $output_flop_kind,
     DefaultValuedAttr<BoolAttr, "true">:$send_supported,
     DefaultValuedAttr<BoolAttr, "true">:$recv_supported
   );
@@ -1653,7 +1725,6 @@ def Xls_ChanOp : Xls_Op<"chan", [
     void setArgAttrsAttr(ArrayAttr) { return; }
     void setResAttrsAttr(ArrayAttr) { return; }
   }];
-
 }
 
 def Xls_InstantiateEprocOp : Xls_Op<"instantiate_eproc", [DeclareOpInterfaceMethods<SymbolUserOpInterface>]> {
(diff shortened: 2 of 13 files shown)


ericastor:
FWIW, the I/O flopping configuration values are important to many FIFO libraries - they change the way the FIFOs are instantiated. You can see one example of how it's relevant here, in our own FIFO generator: https://github.com/google/xls/blob/main/xls/codegen/materialize_fifos_pass.cc#L162

They don't need to be in v1 of this, but they should probably be added at some point!

schilkp:
> They don't need to be in v1 of this, but they should probably be added at some point! (EDIT: And if they are here, they should almost certainly stay.)

Got it - I was not sure so I implemented them while I was at it so they are good to go :) 

> FWIW, the I/O flopping configuration values are important to many FIFO libraries - they change the way the FIFOs are instantiated. You can see one example of how it's relevant here, in our own FIFO generator: https://github.com/google/xls/blob/main/xls/codegen/materialize_fifos_pass.cc#L162

Just for my own understanding - I was digging around this part of codegen (also for #1896) and I still don't quite understand how `input_flop_kind` and `output_flop_kind` (the two pieces of `ChannelConfig` that are outside of `FifoConfig`) are used in the materialize FIFOs pass. 

As far as I can tell only `FifoConfig` is used to construct the fifo because that pass does not have access to the original channel's `ChannelConfig` (in other words, it uses `depth`, `bypass`, `register_push_outputs`, and `register_pop_outputs` only) 

Are those values somehow constructed from `input_flop_kind` and `output_flop_kind`?

The only other reference I can find to those values is during the "Convert IR to Blocks" pass, which seems to construct registers on the proc blocks the channels are attached to? [^1]


I am asking because I have been working on a project that relies on precisely how XLS instantiates buffers, so I am trying to make sure I really understand it :)



[^1]: https://github.com/google/xls/blob/1d94eb0a27491db56174cb40cc18fe3e5332c8dd/xls/codegen/proc_block_conversion.cc#L1324

grebe:
I believe `ChannelConfig` holds the metadata for both internal (FIFO-based) and boundary channels.  It might be more accurate it's constructor arg was a `std::variant<FifoConfig, FlopKind, FlopKind>`.

This is why the materialize_fifo_pass doesn't deal with FlopKinds- it only concerns itself with FIFO configurations, as FIFOs are used for internal channels, not boundary channels. The pass doesn't need access to the whole `ChannelConfig` b/c the `FifoConfig` is the only relevant portion.

schilkp:
Thanks you two for the explaination. That really clears it up.  I was very confused why the flopping was not relevant for the fifo instantiation directly.

To make sure I get this right:

For input(output) boundary channels:
- The input(output) flopping determines if the connected proc input(output) gets a flop/buffer.
- The fifo config is ignored.

For internal channels:
- The input+output flopping determines if the connected proc input and output gets a flop/buffer
- The fifo config is used to parametrize the actually FIFO that makes up the channel.

In other words - the input and output flopping is relevant for both internal and boundary channels?

ericastor:
That's correct!

schilkp:
rebased.

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -135,6 +135,36 @@ def Xls_TranslationLinkage : Xls_Attr<"TranslationLinkage"> {
...
 }
 
+def Xls_FlopKindNone        : I32EnumAttrCase<"kNone", 0, "none">;
+def Xls_FlopKindFlop        : I32EnumAttrCase<"kFlop", 1, "flop">;
+def Xls_FlopKindSkid        : I32EnumAttrCase<"kSkid", 2, "skid">;
+def Xls_FlopKindZeroLatency : I32EnumAttrCase<"kZeroLatency", 3, "zero-latency">;
+def Xls_FlopKind : I32EnumAttr<"FlopKind", "XLS flop kind",
+  [Xls_FlopKindNone, Xls_FlopKindFlop, Xls_FlopKindSkid, Xls_FlopKindZeroLatency]> {
+  let genSpecializedAttr = 0;
+  let cppNamespace = "::mlir::xls";
```
jpienaar, line 145: Could we perhaps document the cases here? It goes a bit above what you;ve already done, but these  in isolation one has to go look up again :) (perhaps link to [https://google.github.io/xls/codegen_options/](https://google.github.io/xls/codegen_options/) and then copy the first lines over here so easy to see/I need to enable the documentation generation for the dialect still somewhere, but would be useful then))

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -135,6 +135,36 @@ def Xls_TranslationLinkage : Xls_Attr<"TranslationLinkage"> {
...
+}
+def Xls_FlopKindAttr : EnumAttr<Xls_Dialect, Xls_FlopKind, "flop_kind">;
+
+def Xls_FifoConfig : Xls_Attr<"FifoConfig"> {
+  let summary = "Fifo Configuration Attribute";
```
jpienaar, line 150: Nit: FIFO

xls/contrib/mlir/IR/xls_ops.td
```diff
@@ -1774,7 +1809,7 @@ def Xls_SchanOp : Xls_Op<"schan", [Pure, HasParent<"SprocOp">]> {
     OpBuilder<(ins "::mlir::StringRef":$name, "::mlir::Type":$element_type), [{
       auto inChanTy = SchanType::get($_builder.getContext(), element_type, /*isInput=*/true);
       auto outChanTy = SchanType::get($_builder.getContext(), element_type, /*isInput=*/false);
-      build($_builder, $_state, ::mlir::TypeRange{outChanTy, inChanTy}, name, element_type);
+      build($_builder, $_state, ::mlir::TypeRange{outChanTy, inChanTy}, name, element_type, /*fifo_config=*/nullptr, /*input_flop_kind=*/nullptr, /*output_flop_kind=*/nullptr);
```
jpienaar, line 1812: Urg, I thought I have fixed this one upstream already :). Feel free to add "TODO(jpienaar): Remove when this is generated automatically".

xls/contrib/mlir/transforms/proc_elaboration.cc
```diff
@@ -141,7 +141,10 @@ class ElaborationContext
     for (auto [i, arg] : llvm::enumerate(sproc.getNextChannels())) {
       auto chan = builder.create<ChanOp>(
           sproc.getLoc(), absl::StrFormat("%s_arg%d", originalName.str(), i),
-          cast<SchanType>(arg.getType()).getElementType());
+          cast<SchanType>(arg.getType()).getElementType(),
+          /*fifo_config=*/nullptr,
```
jpienaar, line 145: And this is as these are external or as we don't have any other info here?

jpienaar (approved):
Nice, sorry got delayed with CGO/C4ML prep. Looks good and nice mapping retaining the core info. One could also do a roundtrip test (in one lit test convert back and forth and diff) to easily test both paths.

schilkp:
Thanks for the review :)Traveling today, will get to this tomorrow. 

> One could also do a roundtrip test (in one lit test convert back and forth and diff) to easily test both paths.

Is there a LIT-specific mechanism for this, or would you just do an `// RUN:` That pipes between translation tools into filecheck, with a number of `// CHECK:`s that validate nothing changed?

xls/contrib/mlir/IR/xls_ops.td
schilkp: I rebased to the lasted XLS LLVM version, unfortunately not. All the generated builders have no default values for the optional args:
```c++
static void build(::mlir::OpBuilder &odsBuilder,
                  ::mlir::OperationState &odsState,
                  ::mlir::Type out, ::mlir::Type in,
                  ::mlir::StringAttr name,
                  ::mlir::TypeAttr type,
                  /*optional*/mlir::xls::FifoConfig fifo_config,
                  /*optional*/mlir::xls::FlopKindAttr input_flop_kind,
                  /*optional*/mlir::xls::FlopKindAttr output_flop_kind);
```

For MLIR attrs with default value, the builders also get a default value for the arg in the builder/Cpp.

It seems that this what https://github.com/llvm/llvm-project/issues/57118 is getting at :)

Also added the same `TODO` to a few other places that suffer from this.

xls/contrib/mlir/transforms/proc_elaboration.cc
schilkp: To be very honest I was not exactly 100% sure how this works - pinging @jmolloy :)

Definitely, based soley on the args of the `next` functions as used here, there are no fifo properties to be propagated since they are in the SChanOps somewhere else. This is why boundary SChans do not have fifo props with this setup. But this here are not just boundary channels.

My understanding is that this function converts an sproc to an eproc once for caching, which means the channels here are not actually real physical channels/do not actually get used. As far as I can tell during testing the channel props are propagated in all scenarios but I would appreciate James' input :)

schilkp:
@jpienaar feedback addressed except for one point I would @jmolloy's feedback on :)

jpienaar:
> Is there a LIT-specific mechanism for this, or would you just do an `// RUN:` That pipes between translation tools into filecheck, with a number of `// CHECK:`s that validate nothing changed?

Pretty much the latter yes. One can diff with the output of opt tool which would strip the comments (and then you don't need explicit CHECK lines), but often just checking some local invariant is fine for the lit one (it at least verifies nothing failed and specific information retained). Having a more complete diff tool that understands what can safely be ignored is better, and then could be run uniformly over all tests even.

xls/contrib/mlir/transforms/proc_elaboration.cc
jmolloy: You're exactly right, the channels created here are "template" channels that get associated with real channels in `xls.instantiate_eproc` ops. After `-instantiate-eprocs` the templates are replicated for each xls.instantiate_eproc op and the channels replaced.

So as long as the "global" channel list given to xls.instantiate_eproc contains the fifo config, the "local" channels (which these are) do not (and they will be deleted eventually).

xls/contrib/mlir/IR/xls_ops.cc
```diff
@@ -620,6 +620,11 @@ ParseResult SchanOp::parse(OpAsmParser& parser, OperationState& result) {
   result.addAttribute("type", TypeAttr::get(type));
   result.types.push_back(SchanType::get(parser.getContext(), type, false));
   result.types.push_back(SchanType::get(parser.getContext(), type, true));
+
+  if (parser.parseOptionalAttrDictWithKeyword(result.attributes)) {
```
jpienaar, line 624: One could also do

return failure(parser.parseOptionalAttrDictWithKeyword(result.attributes));

to collapse conditional.

jpienaar (approved):
Looks good to me, thanks!

xls/contrib/mlir/transforms/proc_elaboration.cc
schilkp: Great. Thanks for the confirmation :)

jpienaar (approved):
Approving again as seems workflows didn't start (else I may manually go and poke)

schilkp:
Ah I was quickly writing the round-trip test if you want to hold off for another 20min or so :) 

I ping you when done.

schilkp:
> Ah I was quickly writing the round-trip test if you want to hold off for another 20min or so :)
> 
> I ping you when done.

@jpienaar Actually it seems that nobody every did a MLIR -> IR -> MLIR roundtrip and this is broken because of some loc-related bugs. I have it mostly fixed but won't wrap it up today. Feel free to merge this and I open a PR for the location-based bugs and round-trip test of fifo props tomorrow, or feel free to hold off merge this :)

jpienaar:
I'll submit this one. For locations: remember to enable printing locations, else parsing will go south/drop it :)

PR #1919, opened 2025-02-08
[DSLX:ir_conv] Add forgotten flag and add unit test for it.

cdleary:
I don't believe this will trigger the IFTTT.

### DIFF
--- a/xls/dslx/ir_convert/ir_converter_options_flags_test.cc
+++ b/xls/dslx/ir_convert/ir_converter_options_flags_test.cc
@@ -1,0 +1,66 @@
+// Copyright 2025 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "xls/dslx/ir_convert/ir_converter_options_flags.h"
+
+#include "absl/cleanup/cleanup.h"
+#include "absl/flags/flag.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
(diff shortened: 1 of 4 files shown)


xls/dslx/ir_convert/ir_converter_options_flags_test.cc
```diff
@@ -0,0 +1,64 @@
...
+     WarningOptionsBothEnableAndDisableSetsMultiEntry) {
+  // Validate the default value we're going to reset this flag to.
+  EXPECT_EQ(absl::GetFlag(FLAGS_warnings_as_errors), true);
+
+  absl::SetFlag(&FLAGS_enable_warnings,
+                "already_exhaustive_match,should_use_assert");
+  absl::SetFlag(&FLAGS_disable_warnings, "constant_naming,member_naming");
+  absl::SetFlag(&FLAGS_warnings_as_errors, false);
+  absl::Cleanup reset_flags([] {
+    absl::SetFlag(&FLAGS_enable_warnings, std::nullopt);
```
grebe, line 49: I think using `absl::GetFlag(...).DefaultValue()` might make the intent a bit more clear?

xls/dslx/ir_convert/ir_converter_options_flags_test.cc
```diff
@@ -0,0 +1,64 @@
...
+  EXPECT_TRUE(options.enable_warnings().empty());
+  EXPECT_TRUE(options.disable_warnings().empty());
+  EXPECT_TRUE(options.has_warnings_as_errors());
+  EXPECT_EQ(options.warnings_as_errors(), true);
+}
+
+TEST(IrConverterOptionsFlagsTest,
+     WarningOptionsBothEnableAndDisableSetsMultiEntry) {
+  // Validate the default value we're going to reset this flag to.
+  EXPECT_EQ(absl::GetFlag(FLAGS_warnings_as_errors), true);
```
grebe, line 42: I think you should check all the flags, right? It might make sense for these to be `ASSERT_EQ()` too.

grebe (changes requested):
Just some nits, otherwise LGTM

xls/dslx/ir_convert/ir_converter_options_flags_test.cc
grebe: Ah, that's a bummer. I think it's good to be explicit.

grebe (approved):

PR #1921, opened 2025-02-09
[DSLX:ir_conv] IR convert trailing match arms w/ multiple patterns.

cdleary:
Previously we would give an IR conversion error, which was a vestige from where we needed the trailing arm to be a wildcard. Now with exhaustiveness we know the trailing match arm must be usable as a default case in the priority select that we lower to.

### DIFF
--- a/xls/dslx/ir_convert/function_converter.cc
+++ b/xls/dslx/ir_convert/function_converter.cc
@@ -1206,8 +1206,8 @@ absl::Status FunctionConverter::HandleMatch(const Match* node) {
     arm_values.push_back(arm_rhs_value);
   }
 
-  // For compute of the default arm the control predicate is "none of the other
-  // arms matched".
+  // For compute of the default (last) arm the control predicate is "none of the
+  // other arms matched".
   ScopedControlPredicate scp(
       this, [&](const PredicateFun& orig_control_predicate) {
         // The default arm is "activated" when:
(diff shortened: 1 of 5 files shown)


richmckeever (approved):

cdleary:
Rebased this, PTAL.

meheff (approved):

proppy:
@cdleary do we want to also update the existing examples to remove unnecessary trailing match clause w/ wildcards?

cdleary:
@proppy Yeah I'd think so, can't think of a reason why not -- I did a quick search for `.md` that have `_ =>` and nothing was jumping out at me, do you see some I should have updated but missed?

proppy:
found at least one :)
https://github.com/google/xls/blob/6f68726f8bafd61bbf9a1aa76b56543137aaa80a/xls/examples/jpeg/idct_chen.x#L239

but I was also wondering if we should make this a failure/warning during conversion?

cdleary:
> found at least one :)
> 
> https://github.com/google/xls/blob/6f68726f8bafd61bbf9a1aa76b56543137aaa80a/xls/examples/jpeg/idct_chen.x#L239
> 
> but I was also wondering if we should make this a failure/warning during conversion?

@proppy This is a warning https://sourcegraph.com/github.com/google/xls/-/blob/xls/dslx/warning_kind.h?L45 already-exhaustive-match , it's just disabled by default so Google-internal codebase can do the global change that enables it by default. In my repo I have a toolchain-level config that enables it by default.

PR #1922, opened 2025-02-09
[DSLX:TS] Fix bug with exhaustiveness sampling enum from another module.

cdleary:


xls/dslx/exhaustiveness/match_exhaustiveness_checker.cc
```diff
@@ -413,13 +413,19 @@ MatchExhaustivenessChecker::SampleSimplestUncoveredValue() const {
       // We have to project back from dense space to enum name space.
       const EnumType& enum_type = type.AsEnum();
       const EnumDef& enum_def = enum_type.nominal_type();
+
+      absl::StatusOr<const TypeInfo*> enum_type_info =
```
grebe, line 417: Nit: because there's an `enum_type` and `enum_def`, it's a bit confusing as to which this type info refers. Maybe `enum_def_type_info` is a better name?

grebe (changes requested):

xls/dslx/exhaustiveness/match_exhaustiveness_checker.cc
cdleary: Done.

grebe (approved):

PR #1924, opened 2025-02-11
[DSLX] `use` syntax: enable using a module.

cdleary:
analogous (but more powerful than) import, e.g.

```
use std;
```

and

```
use path::to::my_module;
```

Then we can colon-ref off of that module name.

The type of a module reference is an existential wrapper of the module identity itself. This lets us keep type information for the AST node in the TypeInfo mapping without having a "hole" where an AST node has no associated type information.

Implementation-wise this mostly means that `UseTreeEntry*` nodes have to ride the same pathways that `Import*` nodes were already riding in order to resolve module attributes.

### DIFF
--- a/xls/dslx/frontend/ast.cc
+++ b/xls/dslx/frontend/ast.cc
@@ -951,7 +951,7 @@ ColonRef::ColonRef(Module* owner, Span span, Subject subject, std::string attr,
 
 ColonRef::~ColonRef() = default;
 
-std::optional<Import*> ColonRef::ResolveImportSubject() const {
+std::optional<ImportSubject> ColonRef::ResolveImportSubject() const {
   if (!std::holds_alternative<NameRef*>(subject_)) {
     return std::nullopt;
   }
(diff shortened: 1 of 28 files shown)


xls/dslx/frontend/ast.cc
```diff
@@ -951,7 +951,8 @@ ColonRef::ColonRef(Module* owner, Span span, Subject subject, std::string attr,
 
 ColonRef::~ColonRef() = default;
 
-std::optional<Import*> ColonRef::ResolveImportSubject() const {
+std::optional<std::variant<UseTreeEntry*, Import*>>
```
richmckeever, line 954: For this variant, which occurs 25 times in here, it may be cleaner to do one of the following:

- Alias the variant type
- Wrap the variant in an object whose interface allows consumers to rarely care which variant option underlies it, and deal that object out from here.
- Have the underlying objects implement a common interface to achieve the same effect

This should also make it easier to eventually delete Import if that's the plan.

richmckeever (approved):

xls/dslx/frontend/ast.cc
cdleary: Done, added `ImportSubject` as a new alias to the variant (picked that one because I try to avoid polymorphism unless it reflects behavior sharing, which I think may have been also why you put it as the first choice).

Just to explain my original lots-of-typing in a side note, typically I lean pretty hard into the "hidden cost of typing less" (which I think used to be a tip of the week? but I can't find it online...) argument that sometimes aliases can obscure underlying types so just typing more for "plain" types like vectors and maps is not a bad tradeoff to make things more superficially clear. I tend to put variants in that bucket too. No strong opinions though, definitely there is a cost to typing more and the more "behavior representing" a type becomes the more an alias for it makes sense.

PR #1925, opened 2025-02-11
Check in MODULE.bazel.lock.

grebe:
This follows the best practices outlined in https://bazel.build/external/lockfile#best-practices.


hzeller (approved):
LG, but I still think that unless someone actually makes an effort in manually verifying the 1125 content of the *.lock file, it is not super-useful but will result in more headaches than benefits down the line...

PR #1927, opened 2025-02-12
Generate `nodata` fifo wrapper for FIFOs with width 0.

schilkp:
Since verilog does not allow zero-width channels or parametric removal of module I/O, a FIFO with zero data cannot be (ergonomically) implemented by parameterizing the general "with data" FIFO.

Update the block generator to generate instantiations of `xls_nodata_fifo_wrapper` instead of `xls_fifo_wrapper` in this case.

Also add test cases that exercise this instantiation.

---

*Note:*
ATM the tests are just copies of the other two fifo block generator tests with small adjustments. I looked into parameterizing the tests with fifo width so the same ones can be used, but because the tests already use the parametric `VerilogTestBase ` (which selects different simulation targets) I don't see a clean and simple way of doing this?

### DIFF
--- a/xls/codegen/block_generator.cc
+++ b/xls/codegen/block_generator.cc
@@ -837,46 +837,56 @@ class BlockGenerator {
             connections);
       } else if (xls::FifoInstantiation* fifo_instantiation =
                      dynamic_cast<FifoInstantiation*>(instantiation)) {
-        std::initializer_list<Connection> parameters{
-            Connection{
-                .port_name = "Width",
-                .expression = mb_.file()->Literal(
-                    UBits(fifo_instantiation->data_type()->GetFlatBitCount(),
-                          32),
-                    SourceInfo(),
-                    /*format=*/FormatPreference::kUnsignedDecimal)},
-            Connection{.port_name = "Depth",
-                       .expression = mb_.file()->Literal(
-                           UBits(fifo_instantiation->fifo_config().depth(), 32),
-                           SourceInfo(),
-                           /*format=*/FormatPreference::kUnsignedDecimal)},
-            Connection{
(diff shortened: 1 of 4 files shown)


schilkp:
As discussed in #1896 - @grebe

xls/codegen/block_generator.cc
```diff
@@ -908,8 +918,11 @@ class BlockGenerator {
           return a.port_name < b.port_name;
         });
 
+        const char* wrapper_name =
```
grebe, line 921: Please use std::string_view

xls/codegen/block_generator_test.cc
```diff
@@ -259,22 +339,23 @@ class BlockGeneratorTest : public VerilogTestBase {
...
-  absl::StatusOr<CodegenResult> MakeMultiProc(FifoConfig fifo_config) {
+  absl::StatusOr<CodegenResult> MakeMultiProc(FifoConfig fifo_config,
+                                              uint64_t data_width) {
     Package package(TestName());
-    Type* u32 = package.GetBitsType(32);
-    XLS_ASSIGN_OR_RETURN(
-        Channel * in,
-        package.CreateStreamingChannel("in", ChannelOps::kReceiveOnly, u32, {},
-                                       std::nullopt, FlowControl::kReadyValid));
+    Type* type_data = package.GetBitsType(data_width);
```
grebe, line 345: Super-Nit: I think `Type* uN = ...` would be a bit easier to read

grebe (changes requested):
Sorry to be slow!

schilkp:
> Sorry to be slow!

No worries and thanks for the review. Resolved.

grebe (approved):

PR #1944, opened 2025-02-22
[c_api] Support more value APIs (used in xlsynth crate)

cdleary:
- xls_value_get_element_count
- xls_value_get_element
- xls_value_clone
- xls_value_make_ubits
- xls_value_make_sbits


meheff (approved):

xls/public/c_api_test.cc
```diff
@@ -331,6 +331,25 @@ TEST(XlsCApiTest, FlattenTupleValueToBits) {
   xls_value* tuple = xls_value_make_tuple(/*element_count=*/2, elements);
```
mikex-oss, line 331: Should we add some coverage for xls_value_get_element_count here? (Separate test is fine too)

cdleary: Done.

PR #1966, opened 2025-02-28
[c_api] Expose function JIT facilities (used in xlsynth crate)

cdleary:
Note that `xls_value_from_bits` erroneously moved out of the const ref before and that's fixed in this PR (I didn't know you could even do that).

### DIFF
--- a/xls/public/c_api.h
+++ b/xls/public/c_api.h
@@ -50,6 +50,7 @@ struct xls_function_base;
 struct xls_type;
 struct xls_function_type;
 struct xls_schedule_and_codegen_result;
+struct xls_function_jit;
 
 void xls_init_xls(const char* usage, int argc, char* argv[]);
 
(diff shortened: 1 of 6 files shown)


meheff (approved):

xls/public/c_api.h
```diff
@@ -347,8 +356,34 @@ bool xls_function_type_to_string(struct xls_function_type* xls_function_type,
...
+
+struct xls_trace_message {
+  char* message;
+  int64_t verbosity;
+};
+
+// Runs the given `jit` function with the given `args` (an array of size `argc`)
+// and returns the result in `result_out`.
+//
+// `trace_messages_out` should be freed by the caller via ...
```
mikex-oss, line 375: Is the "..." intentional?

cdleary: I hadn't written the function for that yet and forgot to come back and fill it in, will send a fix.

PR #1975, opened 2025-03-05
[xls][mlir] Add MLIR LSP server specialization with relevant dialects

schilkp:
See https://mlir.llvm.org/docs/Tools/MLIRLSP/#mlir-lsp-language-server--mlir-lsp-server

@jpienaar @jmolloy 

(I am sure you having something like this internally, but that is probably tied to many other projects)


xls/contrib/mlir/tools/xls_mlir_lsp_server/xls_mlir_lsp_server.cc
```diff
@@ -0,0 +1,17 @@
+#include "mlir/include/mlir/Dialect/Arith/IR/Arith.h"
```
jpienaar, line 1: Sorry I missed this: copyright notice is missing.

schilkp:
Fixed. Sorry :)

jpienaar (approved):

PR #1979, opened 2025-03-06
Centralize definition/validation of supported FIFO config.

schilkp:
Replace distributed (and inconsistent) checks for supported FIFO configs.

Fixes tests with unsupported FIFO configs.

---

As mentioned in https://github.com/google/xls/pull/1896.

I assume the mismatch here is because they got out of sync, and not because of different requirements.

Note that, because the checks were inconsistent and there is little documentation on the purpose of these restrictions, I took the union/most restrictive set.

Also it would be helpful to add documentation/comments to the `isSupported` function that explain these limits. I don't quite know where they come from  :)

I assume that @allight and @grebe are responsible here. Thanks! :)

### DIFF
--- a/xls/ir/channel.h
+++ b/xls/ir/channel.h
@@ -63,6 +63,30 @@ class FifoConfig {
   bool bypass() const { return bypass_; }
   bool register_push_outputs() const { return register_push_outputs_; }
   bool register_pop_outputs() const { return register_pop_outputs_; }
+  absl::Status Validate() const {
+    if (depth_ == 0) {
+      if (!bypass_) {
+        return absl::FailedPreconditionError(
+            "Zero-depth FIFOs are direct connections, and hence cannot be "
+            "bypass-less");
+      }
+      if (register_push_outputs_ || register_pop_outputs_) {
+        return absl::FailedPreconditionError(
+            "Zero-depth FIFOs are memory-less/direct connections, and hence do "
+            "not support registers");
+      }
+    }
+    if (depth_ == 1) {
+      if (register_pop_outputs_) {
+        return absl::FailedPreconditionError(
+            "register_pop_outputs adds an extra FIFO entry for the output and "
+            "still requires a depth>=1 FIFO in front of the output stage, so "
+            "we require depth>=2");
+        // Please file a github issue if you have different FIFO needs :)
+      }
+    }
+    return absl::OkStatus();
+  }
 
   bool operator==(const FifoConfig& other) const = default;
   bool operator<=>(const FifoConfig& other) const = default;
--- a/xls/codegen/block_stitching_pass_test.cc
+++ b/xls/codegen/block_stitching_pass_test.cc
@@ -4905,7 +4905,7 @@ TEST_F(ProcInliningPassTest, ProcWithNonblockingReceivesWithPassthrough) {
   constexpr std::string_view ir_text = R"(package test
 
 chan in(bits[32], id=0, kind=streaming, ops=receive_only, flow_control=ready_valid)
-chan internal(bits[32], id=1, kind=streaming, ops=send_receive, flow_control=ready_valid, fifo_depth=$0, bypass=$1, register_push_outputs=$2, register_pop_outputs=$2)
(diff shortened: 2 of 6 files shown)


xls/ir/channel.h
```diff
@@ -63,6 +63,24 @@ class FifoConfig {
   bool bypass() const { return bypass_; }
   bool register_push_outputs() const { return register_push_outputs_; }
   bool register_pop_outputs() const { return register_pop_outputs_; }
+  bool isSupported() const {
```
grebe, line 66: Our [style guide](https://google.github.io/styleguide/cppguide.html#Function_Names) says this should be `IsSupported()`.

Names are hard, so feel free to push back or propose something else, but "Supported" feels to me not quite strong enough. Maybe `IsLegal()`?

An alternative is to have something like

```
absl::Status Validate() {
  if (...) {
    return absl::FailedPreconditionError("<Descriptive error message here>");
  }
  return absl::OkStatus();
}
```

and turn `CHECK()` into `CHECK_OK()` (and `XLS_RETURN_IF_ERROR()` lets you avoid having to make new error messages, I guess).

grebe (changes requested):

grebe:
Thanks for doing this btw, it's nice to catch these things and clean up the mess!

xls/ir/channel.h
schilkp: This gets to my comment about not knowing why these limits are in place: 

> Also it would be helpful to add documentation/comments to the isSupported function that explain these limits. I don't quite know where they come from :)

Two possible reasons for these limits:

- (A) FIFO configuration is "illegal" because it does not make physical sense.
- (B) FIFO configuration is "unsupported" because they are physically sensible/possible but XLS can't deal with them atm.

I guess the restriction for `depth=0`/bypass is bin `A` ( it does not make sense to have a non-bypassed zero-depth fifo)

How about the restriction for `register_push/pop_outputs`? I have not dug into what 'register_push/pop_outputs` does? Is that also bin `A`? 

Would be helpful for me to know so I can make the error messages descriptive :)

Anyway I like `Validate` better - covers both ;P

---

Re using absl error stuff/naming: Good call ofc - I spent most my time in LLVM/MLIR world so I am not used to the "normal" xls style.

Would again be good if you could clarify why these limits exist - that way I can make the error msgs descriptive.

grebe: Sorry, I got caught up in the nitpicks and forgot the important bit.

For the depth=0, the restriction's rationale is as you say: depth=0 means "directly connect push<->pop" with wires. Possibly, it would be less confusing for FifoConfig to be a `std::variant<>` where one is "direct connect" and the other has depth > 0.

For the register_pop_outputs -> depth > 1, I think the rationale is that register_pop_outputs conceptually adds a depth=1 FIFO constructed to cut timing paths at the end, so it makes a (depth - 1) FIFO and appends the register_pop_outputs "FIFO" to that. I don't think the special output FIFO works correctly if you attach it directly to the pop inputs, so you need an otherwise normal depth>=1 FIFO in front of the special output FIFO. depth - 1 needs to be >= 1, so register_pop_outputs only works for depth>=2.

Currently, our fifo parameterization is tied pretty tightly to our (unfortunately internal) FIFO implementation, so these parameters mostly mean whatever they mean for our FIFO. I'd like to abstract the parameters a bit more and make them a bit less confusing. For example, [these](https://github.com/xlsynth/bedrock-rtl/blob/main/fifo/rtl/br_fifo_flops.sv) fifos are pretty similar and it would be nice for us to be able to target them in OSS (but note that they require depth>=2 and have different depth=1 FIFOs).

How to condense this into useful error messages? I think it's OK to say something like:

- depth = 0 FIFOs are direct connections, setting the other options differently only makes sense if there is a memory
- register_pop_outputs adds an extra FIFO entry for the output and still requires a depth>=1 FIFO in front of the output stage, so we require depth>=2. Please file a github issue if you have different FIFO needs :)

schilkp:
Feedback implemented + Rebased :)

grebe (approved):

schilkp:
Is that timeout on `//xls/jit:function_jit_test` my fault? Everything works-on-my-machine©️™️  :)

PR #2003, opened 2025-03-17
[IR] Add case type verification for OneHotSel and PrioritySel

cdleary:
Previously we would not flag that one hot sel and prio sel had type inconsistency among their cases (i.e. even when verify arg was given as true to the builder).


ericastor (approved):
Thanks, Chris!

PR #2004, opened 2025-03-18
[xls][mlir] Add xls->mlir translation of `trace` IR node

schilkp:
@jpienaar

### DIFF
--- a/xls/contrib/mlir/tools/xls_translate/xls_translate_to_mlir.cc
+++ b/xls/contrib/mlir/tools/xls_translate/xls_translate_to_mlir.cc
@@ -1264,6 +1264,41 @@ absl::StatusOr<Operation*> translateOp(::xls::CountedFor& node,
       *loc, result_type, *initial_value, invar_args, trip_count, *body, stride);
 }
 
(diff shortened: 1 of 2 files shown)


jpienaar (approved):

xls/contrib/mlir/tools/xls_translate/xls_translate_to_mlir.cc
```diff
@@ -1264,6 +1264,41 @@ absl::StatusOr<Operation*> translateOp(::xls::CountedFor& node,
       *loc, result_type, *initial_value, invar_args, trip_count, *body, stride);
 }
 
+absl::StatusOr<Operation*> translateOp(::xls::Trace& node, OpBuilder& builder,
```
jpienaar, line 1267: (Unrelated to this chance as this is the same signature everywhere, but builder does a context, so one could avoid passing context additionally and just query it where needed https://mlir.llvm.org/doxygen/classmlir_1_1Builder.html#aa6d1e114c8047cad1b014b504688a868)

schilkp: ah I did not know. I might clean that up in a different PR if I have a moment.

PR #2007, opened 2025-03-18
[c_api] Add ir builder target to `:c_api` as dep and header to `c_api.h`.

cdleary:
This seems required for OS X builds to link the artifact -- not sure why it wasn't required on other platforms, maybe there's some different per-object-artifact culling.

### DIFF
--- a/xls/public/BUILD
+++ b/xls/public/BUILD
@@ -243,6 +243,7 @@ cc_library(
         ":c_api_dslx",
         ":c_api_format_preference",
         ":c_api_impl_helpers",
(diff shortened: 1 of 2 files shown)


meheff (approved):

proppy:
@cdleary it seems that the https://crates.io/crates/xlsynth-sys/0.0.101 crate already assume this patch is in, is that intentional?

proppy:
@cdleary this also fixed internal builds of the xlsynth crate, thanks!

PR #2009, opened 2025-03-19
Implement downcasting for apfloat when exponent width is also possibly reduced.

meheff:
This change hides the method downcast_fractional_rne as private to apfloat.x and instead exposes the more general downcast_rne, which also allows the exponent width to be reduced.

This change also introduces a helper method exponent_bias, which returns the constant offset used to encode the exponent for a given type (e.g. 127 for types with an 8-bit exponent), and simplifies the methods bias() and unbiased_exponent().

FolderOrigin-RevId: /google/src/cloud/meheff/clean/google3/..

### DIFF
--- a/docs_src/floating_point.md
+++ b/docs_src/floating_point.md
@@ -178,27 +178,42 @@ exponent field. For single precision floats this value is -126.
 ### `apfloat::unbiased_exponent`
 
 ```dslx-snippet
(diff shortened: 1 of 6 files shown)


meheff:
@rmlarsen, FYI

cdleary (approved):

rmlarsen:
@meheff thanks! 

Hi @cdleary . Hope the new gig  is going well. :-)

meheff:
Landed internally (5878eafac2c08849e58313b306a0954de1609e4c)

rmlarsen:
Thanks, @meheff

Issue #2006, opened 2025-03-18
Running clang-format on github does not result in no changes


hzeller:
When running `clang-format` (as provided from the packaged tool-chain), this results in a few changes, which will generate unnecessary churn if contributors want to provide a clean PR and run clang-format before and accidentally verschlimmbesser the diff.

To reproduce:

```bash
CLANG_FORMAT=$(bazel run -c opt --run_under="echo" @llvm_toolchain//:clang-format 2>/dev/null)
${CLANG_FORMAT} -i $(find xls -name "*.h" -o -name "*.cc")
```
Afterwards `git diff` does not result in an empty diff.

Observations and possible mitigation

  * Header ordering changes. Reason is that the path prefixes are possibly different in the copybara roundtrip. Possible **Mitigation**: run `clang-format` as last step when copy code in copybara (something like that is already done with `buildifier`).
  * Difference in formatting: in constructors, an empty `{}` body is formatted with space in front coming from the git repo, but without space in front after `clang-format`. From what I can tell, this is due to the llvm version. XLS [uses toolchain 17.0.6](https://github.com/google/xls/blob/2037a514041ea1f7279340ced0df0e9aa42e985b/MODULE.bazel#L11-L13), I think this different formatting is done with newer clang formats. Possible **Mitigation** Update version in used toolchain.
  * Right hugging pointers and reference symbols are formatted left-hugging (`int *a` -> `int* a`) according to [the configuration](https://github.com/google/xls/blob/2037a514041ea1f7279340ced0df0e9aa42e985b/.clang-format#L6-L8) but there are a few files in the repo where these symbols were right-aligned and subsequently fixed with the clang-format run. I suspect the special handling for these symbols is not set in the internal repo configuration for clang-format and `DerivePointerAlignment` is true there. **Mitigation** Use same clang-format configuration internally.

PR #2014, opened 2025-03-21
Use or-tools from Bazel Central Registry.

hzeller:
Or-tools is now available in BCR
https://github.com/google/or-tools/issues/4551

This implicitly also pushed protobuf, abseil-cpp and fuzztest dependencies, so make that state of versions explicit in the root project.

Newer protobuf does not compile with clang 17 anymore [^1], so update clang (this might also help fix some clang-format differences we see in issue #2006)

protoc-gen-validate did not work with newer protobuf rules anymore, but luckily it got updated in BCR in the meantime ans works without local patch now.
Grpc now also exhibit a compile error, so update that to head as well.

Issues: #931

[^1]: https://github.com/protocolbuffers/protobuf/issues/20645


hzeller:
Converting to draft. Looks like that new llvm toolchain has that dreaded linking against tinfo problem.
I suspect we might need to install ncurses to have that lib available ?

hzeller:
Alright, it resolved it by using LLVM 18.1.7 instead of 18.1.8.
For whatever reason, the latter is linked against an ancient `libtinfo5` (which is not on Ubuntu), while the former is linked against a current `libtinfo6`

fuzztest.bazelrc
```diff
@@ -20,12 +21,19 @@ build:fuzztest-common --copt=-UNDEBUG
...
 # See https://libcxx.llvm.org/UsingLibcxx.html#enabling-the-safe-libc-mode
 build:fuzztest-common --copt=-D_LIBCPP_ENABLE_ASSERTIONS=1
+
 ### ASan (Address Sanitizer) build configuration.
 #
 # Use with: --config=asan
 
 build:asan --linkopt=-fsanitize=address
 build:asan --copt=-fsanitize=address
+
```
hzeller, line 31: FYI, this file changed because of the updated fuzztest which generates a new bazelrc include.

ericastor:
Great cleanup - thanks, Henner!

ericastor (approved):

PR #2019, opened 2025-03-28
Adapt fuzzer known failure regex to new error message format.

schilkp:
Introduced in https://github.com/google/xls/commit/adcc9707cc4c0ef1eaf6264144d34c0ea8463e5c

@grebe


schilkp:
I did not fix the regexes in the options of the checked-in crashers - Should those also be updated?

Issue #1140, opened 2023-10-02
Fuzzer sometimes generates programs which fail codegen with 'cannot achieve the specified pipeline length'


allight:
See xls/fuzzer/crashers/crasher_2023-10-02_7c14.x

This should be remedied by either making the fuzz codegen able to pick the right pipeline or not generate these sorts of issues. For now programs which crash in this way are filtered out and not considered failures.

PR #2020, opened 2025-03-28
Fuzzer: Correctly propagate pipeline info through counted-fors

schilkp:
This is one of the cause of https://github.com/google/xls/issues/1140.

A token produced by a send, passed through a counted_for, and then consumed by a receive would not correctly increase the `min_stage`. For example, the following would incorrectly report a min pipeline stages of 1:

```rust
let x3: token = send(join(), x0, u1:1);
let x4: token = for (i, x) in u4:0x0..u4:0x8 {
    x
}(x3);
let x5: (token, u37, bool) = recv_non_blocking(x4, x1, u37:45812984490);
```

Not sure if there are other reasons for https://github.com/google/xls/issues/1140 or if the `known_failure` can be removed?

Edit: Still seems to happen - 1140 not resolved.

### DIFF
--- a/xls/fuzzer/ast_generator.cc
+++ b/xls/fuzzer/ast_generator.cc
@@ -1894,7 +1894,10 @@ absl::StatusOr<TypedExpr> AstGenerator::GenerateCountedFor(Context* ctx) {
       fake_span_, std::vector<Statement*>{body_stmt}, /*trailing_semi=*/false);
   For* for_ = module_->Make<For>(fake_span_, name_def_tree, tree_type, iterable,
                                  block, /*init=*/e.expr);
-  return TypedExpr{for_, e.type};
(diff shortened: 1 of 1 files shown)


schilkp:
Not sure who to tag - probably @ericastor since you added the min_stage/pipeline tracking?

ericastor (approved):
Well-spotted! I'm not sure it'll ever be possible to remove this `known_failure` entirely, since that amounts to knowing (in advance) how many pipeline stages are actually required to schedule a circuit. The logic I added is a useful approximation that cuts down on our wasted time, but I don't think it can fully eliminate it.

The only way I can see fixing this once and for all would be to provide a method for the user to tell XLS "just schedule this into the fewest pipeline stages possible", and having the fuzzer use that.

PR #2023, opened 2025-03-28
[stdlib] Add new functions for extracting the `n` LSBs/MSBs from a value

ericastor:
Where `n` is variable, there are many suboptimal ways to do this; to guide users to more efficient implementations, we provide stdlib functions `std::mask_lsbs` and `std::mask_msbs`, along with their complements `std::without_lsbs` and `std::without_msbs`.

### DIFF
--- a/docs_src/dslx_std.md
+++ b/docs_src/dslx_std.md
@@ -939,7 +939,7 @@ Extracts the MSb (Most Significant bit) from the value `x` and returns it.
 pub fn mask_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
 ```
 
-Returns a copy of `x` with all but the `n` least-significant bits set to 0.
+Returns a copy of `x` with all but the `n` least-significant bits set to 0. If `n >= WIDTH`, returns `x`.
 
 #### `std::mask_msbs`
 
@@ -947,7 +947,7 @@ Returns a copy of `x` with all but the `n` least-significant bits set to 0.
 pub fn mask_msbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
 ```
 
-Returns a copy of `x` with all but the `n` most-significant bits set to 0.
+Returns a copy of `x` with all but the `n` most-significant bits set to 0. If `n >= WIDTH`, returns `x`.
 
 #### `std::without_lsbs`
 
@@ -955,7 +955,7 @@ Returns a copy of `x` with all but the `n` most-significant bits set to 0.
 pub fn without_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
 ```
 
-Returns a copy of `x` with the `n` least-significant bits set to 0.
+Returns a copy of `x` with the `n` least-significant bits set to 0. If `n >= WIDTH`, returns `zero!<uN[WIDTH]>`.
 
 #### `std::without_msbs`
 
@@ -963,7 +963,7 @@ Returns a copy of `x` with the `n` least-significant bits set to 0.
 pub fn without_msbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
 ```
 
-Returns a copy of `x` with the `n` most-significant bits set to 0.
+Returns a copy of `x` with the `n` most-significant bits set to 0. If `n >= WIDTH`, returns `zero!<uN[WIDTH]>`.
 
 #### `std::or_reduce_lsb`
 
--- a/docs_src/dslx_std.md
+++ b/docs_src/dslx_std.md
@@ -933,34 +933,34 @@ pub fn msb<N: u32>(x: uN[N]) -> u1
 
 Extracts the MSb (Most Significant bit) from the value `x` and returns it.
 
-#### `std::mask_lsbs`
+#### `std::keep_lsbs`
 
 ```dslx-snippet
-pub fn mask_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
+pub fn keep_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
 ```
 
 Returns a copy of `x` with all but the `n` least-significant bits set to 0. If `n >= WIDTH`, returns `x`.
 
-#### `std::mask_msbs`
+#### `std::keep_msbs`
 
 ```dslx-snippet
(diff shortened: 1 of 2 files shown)


ericastor:
Sorry for the CLA mixup; fixed now.

hzeller:
API wise, did you consider just having one function, returning the msb and lsb parts returned in a tuple ? It might make for slightly less awkward naming of two different functions, there are certainly situations needed in which both parts would be needed, and performance-wise, there is no difference as if someone discards one of the results, it will not hit silicon.

CC  @proppy who has a good feel for DSLX APIs in the field.

cdleary (approved):
@ericastor This looks good to me thanks for doing it as a PR! My natural inclination would be to call this keep_msbs or keep_lsbs (and perhaps correspondingly drop_msbs / drop_lsbs -- maybe the better antonym would be take_msbs instead of keep?) because "mask" is slightly ambiguous as to whether you're masking things on or off. Would recommend taking a straw poll on the clearest naming with the team if you're amenable to that, thanks!

ericastor:
I like take/drop, I think - but I'll run a quick straw poll for it!

docs_src/dslx_std.md
```diff
@@ -933,6 +933,38 @@ pub fn msb<N: u32>(x: uN[N]) -> u1
 
 Extracts the MSb (Most Significant bit) from the value `x` and returns it.
 
+#### `std::mask_lsbs`
+
+```dslx-snippet
+pub fn mask_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH]
+```
+
+Returns a copy of `x` with all but the `n` least-significant bits set to 0.
```
proppy, line 942: should we document the behavior when `n > WIDTH` ?

xls/dslx/stdlib/std.x
```diff
@@ -1081,6 +1081,90 @@ pub fn sizeof_signed<N: u32>(x: sN[N]) -> u32 { N }
 
 pub fn sizeof_unsigned<N: u32>(x: uN[N]) -> u32 { N }
 
+// Returns `x` with all but the least-significant `n` bits set to zero.
+pub fn mask_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH] {
+    x & !(all_ones!<uN[WIDTH]>() << n)
```
proppy, line 1086: should we add some `assert` to verify that `n` is <= `WIDTH`?
also should `n` width depend on `WIDTH` (`std::clog2(WIDTH)`) or will it get too anoying with casting?

proppy:
I realize that this might sound orthogonal and beyond the scope of this PR: one could think of slicing and masking as two different things (and maybe this warrant discussion in a separate issue?) 

But I'd recommend that we unify the naming and arguments of all slicing-themed functions of the stdlib at some point, as they share a similar argument domain:
- a start (or a end) index
- a width (either deduced or explicit)

More specifically I'm referring to:
- https://google.github.io/xls/dslx_std/#stdextract_bits (static bit slice combined w/ left shift)
- https://google.github.io/xls/dslx_std/#stdvslice (alternative verilog-like flavor)
- https://google.github.io/xls/dslx_std/#bit_slice_update, https://google.github.io/xls/dslx_std/#array_slice (extract the width from an additional arg/sentinel)
- https://google.github.io/xls/dslx_std/#stdlsb, https://google.github.io/xls/dslx_std/#stdmsb (extract a single bit)

and to some extend:
- https://google.github.io/xls/dslx_reference/#bit-slice-expressions (support negative index to differentiate between lsb and msb)

If we had a "dynamic_slice" function (or operator) that accept a signed  index argument ([like the regular slice operator](https://google.github.io/xls/dslx_reference/#:~:text=For%20example%2C%20to%20get,%5B%2D2%3A%5D)), would we still reach to dedicated functions for masking msbs and lsbs?

The function added in this PR would effectively translate into this:
- `mask_lsbs(u8:0xab, u4:4)` -> `let x = u8:0xab; let n = s5:4; dynamic_bit_slice(x, s5:0, n) `  or `let x = u8:0xab; let n = s5:4; x[[:n]]`
- `mask_msbs(u8:0xab, u4:4)` -> `let x = u8:0xab; let n = s5:4; dynamic_bit_slice(x, -n, bit_count<u8>()) << (n - bit_count<u8>())` or `let x = u8:0xab; let n = s5:4; x[[-n:]] << (bit_count<u8>() - n)`
- `without_lsbs(u8:0xab, u4:4)` -> `let x = u8:0xab; let n = s5:4; dynamic_bit_slice(x, n, bit_count<u8>()) << n`  or `let x = u8:0xab; let n = s5:4; x[[n:]] << n`
- `without_msbs(u8:0xab, u4:4)` -> `let x = u8:0xab; let n = s5:4; dynamic_bit_slice(x, 0, -n) << n`  or `let x = u8:0xab; let n = s5:4; x[[:-n]]`

assuming the extra shift could easily be optimized away by the compiler.

ericastor:
RE: https://github.com/google/xls/pull/2023#issuecomment-2765251233

I think I agree that this discussion is beyond the scope of this PR. I'd be glad to have it elsewhere. Maybe we should start a [Discussions](https://github.com/google/xls/discussions) thread? Before we split that out, I'd like to point out that it'd be unclear what the return type of the "dynamic_bit_slice" operator would be, assuming the width was not constexpr... this is visible in the cases you outline for replacing (e.g.) `without_msbs`.

xls/dslx/stdlib/std.x
ericastor: I think I explicitly want to accept `n > WIDTH`, and as a result, I don't think we need any restriction on `n`. However, as you suggested above, we do want to document what `n > WIDTH` should mean!

xls/dslx/stdlib/std.x
```diff
@@ -1081,6 +1081,90 @@ pub fn sizeof_signed<N: u32>(x: sN[N]) -> u32 { N }
 
 pub fn sizeof_unsigned<N: u32>(x: uN[N]) -> u32 { N }
 
+// Returns `x` with all but the least-significant `n` bits set to zero.
+pub fn keep_lsbs<WIDTH: u32, N_WIDTH: u32>(x: uN[WIDTH], n: uN[N_WIDTH]) -> uN[WIDTH] {
+    x & !(all_ones!<uN[WIDTH]>() << n)
```
proppy, line 1086: wouldn't `mask_bits` bit a little bit less verbose here and after?

ericastor: `mask_bits<WIDTH>()`
`all_ones!<WIDTH>()`

are exactly the same length, funnily enough! Also, I'm surprised we have both of those functions.

proppy: I think the different is that:
`all_ones!` takes a type while `std::mask_bits` takes a bit length.

so `all_ones!<T> == std::mask_bits<bit_count<T>>`, but in this case we already have `N_WIDTH`  inferred to we might as well use `std::mask_bits`?

proppy: fyi, that's a nit not a blocking comment.

PR #2024, opened 2025-03-31
[DSLX:stdlib] Add absolute difference module to stdlib.

cdleary:
Absolute difference is a key calculation in e.g. floating point exponent comparison.

### DIFF
--- a/xls/dslx/stdlib/abs_diff.x
+++ b/xls/dslx/stdlib/abs_diff.x
@@ -74,13 +74,14 @@ pub fn is_zero<N: u32>(adr: AbsDiffResult<N>) -> bool {
 ///
 /// Implementation note: this is a technique also used in literature that optimizes
 /// absolute difference calculations, e.g. for motion estimation, such that you can accumulate
-/// correction terms over many calculations without needing carries at every step.
+/// correction terms over many calculations without needing carries at every step. See e.g.
+/// Jehng et al. 10.1109/78.193224 fig 11(a) for a digram exhibiting the boolean formula.
 ///
 /// Proof derivation steps given here thanks to ericastor@:
 /// ```
 /// result = z ^ signex(!carry_out)
 ///        = if carry_out { z } else { !z }  // XOR with signex == controlled-not
-///        = if y < x { z } else { !z }  // noted above
+///        = if y < x { z } else { !z }  // because carry_out = (y < x)
 ///        = if y < x { x + !y } else { !(x + !y) }  // definition of z
 ///        = if y < x { x - y - 1 } else { !(x - y - 1) }  // x - y = x + !y + 1
 ///        = if y < x { x - y - 1 } else { -(x - y - 1) - 1 }  // -q = !q + 1
--- a/xls/dslx/stdlib/abs_diff.x
+++ b/xls/dslx/stdlib/abs_diff.x
@@ -1,0 +1,193 @@
+// Copyright 2025 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Functionality for computing an absolute difference between two numbers with a
+//! "residual" correction term. By keeping a correction term we avoid eagerly
+//! incorporating a second carry chain in computing the final result.
+//!
+//! Sample usage:
+//!
+//! ```dslx
+//! let x: u4 = u4:0;
+//! let y: u4 = u4:1;
+//! let adr: AbsDiffResult<4> = abs_diff(x, y);
+//! assert_eq(to_corrected(adr), u4:1);
+//! assert_eq(is_x_larger(adr), false);
+//! assert_eq(is_zero(adr), false);
+//! ```
+
+import std;
+
+/// Describes the result of an absolute difference operation.
+///
+/// By keeping the result in two terms with one being a correction term, we
+/// can avoid a second carry chain being applied in the operation.
+///
+/// Additionally, this information can be used to tell us which of the two operands
+/// was larger -- use `is_x_larger` for this.
+pub struct AbsDiffResult<N: u32> {
+    correction: bool,
+    /// Note: this is the *uncorrected* absolute difference -- the `correction` term
+    /// must be added to get the true absolute difference -- use `to_corrected` to
+    /// flatten this structure to the absolute difference value.
+    uncorrected: uN[N],
+}
+
+/// Returns `true` if the first operand was larger, `false` if the second operand was larger.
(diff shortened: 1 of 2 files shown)


ericastor:
It's funny, I was just looking at this and other similar operations that play a large role in floating point computations - and this seems like a cool solution for this operation. I'll take a look if you don't mind!

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,177 @@
...
+/// This is a "residual" absolute difference operation -- the result is a structure
+/// with a `correction` term and an `uncorrected` term.
+///
+/// Use `to_corrected` to collapse the result into a single absolute difference value, but
+/// note that this incurs a second carry chain.
+pub fn abs_diff<N: u32>(x: uN[N], y: uN[N]) -> AbsDiffResult<N> {
+    let ynot = !y;
+    let (carry_out, z): (bool, uN[N]) = std::uadd_with_overflow<N>(x, ynot);
+    let nc = !carry_out;
+    let xor_mask = (nc as s1 as sN[N]) as uN[N];
```
ericastor, line 78: This seems like a good application for the [signex](https://google.github.io/xls/dslx_std/#signex) builtin?

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,177 @@
...
+    let result = (z ^ xor_mask);
+    AbsDiffResult { correction: carry_out, uncorrected: result }
```
ericastor, line 80: Think it could be worth adding comments documenting why this computation is correct, or a reference? It's not obvious, and I'm not able to find a reference with a quick search.

My best version of the proof:

First, note that `x + !y` overflows iff `y < x`, so `carry_out = (y < x)`. (Reference: Hacker's Delight, Section 2-12, "Unsigned Add/Subtract", which references [MIPS].)

Then, do a quick series of simplifications/re-expressions,
```
result = z ^ signex(!carry_out)
       = if carry_out { z } else { !z }  // XOR with signex == controlled-not
       = if y < x { z } else { !z }  // noted above
       = if y < x { x + !y } else { !(x + !y) }  // definition of z
       = if y < x { x - y - 1 } else { !(x - y - 1) }  // x - y = x + !y + 1
       = if y < x { x - y - 1 } else { -(x - y - 1) - 1 }  // -q = !q + 1
       = if y < x { x - y - 1 } else { y - x }  // algebraic simplification
```

This seems worth documenting for future readers?

ericastor (commented):
This is a neat demonstration & optimized reference! I thought it was best to respond quickly; I'll make sure to respond later to say whether this looks to produce better code as well.

xls/dslx/stdlib/abs_diff.x
cdleary: Thanks for that! Added that with attribution and some additional context 🙏

xls/dslx/stdlib/abs_diff.x
cdleary: Done, good idea.

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,193 @@
...
+///
+/// Implementation note: this is a technique also used in literature that optimizes
+/// absolute difference calculations, e.g. for motion estimation, such that you can accumulate
+/// correction terms over many calculations without needing carries at every step.
+///
+/// Proof derivation steps given here thanks to ericastor@:
+/// ```
+/// result = z ^ signex(!carry_out)
+///        = if carry_out { z } else { !z }  // XOR with signex == controlled-not
+///        = if y < x { z } else { !z }  // noted above
```
ericastor, line 83: This probably needs to be explained, rather than "noted above" - now that it's not noted above?

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,194 @@
...
+/// Implementation note: this is a technique also used in literature that optimizes
+/// absolute difference calculations, e.g. for motion estimation, such that you can accumulate
+/// correction terms over many calculations without needing carries at every step. See e.g.
+/// Jehng et al. 10.1109/78.193224 fig 11(a) for a digram exhibiting the boolean formula.
+///
+/// Proof derivation steps given here thanks to ericastor@:
+/// ```
+/// result = z ^ signex(!carry_out)
+///        = if carry_out { z } else { !z }  // XOR with signex == controlled-not
+///        = if y < x { z } else { !z }  // because carry_out = (x < y)
```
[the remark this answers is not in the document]
cdleary: Yeah fixed immediately after I hit resolve sorry about that, should be good now.

ericastor (approved):

xls/dslx/stdlib/abs_diff.x
proppy: curious if this could be expression as a quick check (or in z3?)

ericastor: I believe the proof of equivalence already is included as a quickcheck, including a BUILD entry to prove it? It just seemed worth also having an explanation that was human readable.

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/// Returns `true` if the first operand was larger, `false` if the second operand was larger.
+pub fn is_x_larger<N: u32>(adr: AbsDiffResult<N>) -> bool { adr.correction }
+
+/// Collapses the `AbsDiffResult` structure into a single absolute difference value.
+///
+/// (Note that this incurs the second carry chain that we avoided by keeping the
+/// result in two terms.)
+pub fn to_corrected<N: u32>(adr: AbsDiffResult<N>) -> uN[N] {
```
proppy, line 54: curious if those could be impl yet? /cc @erinzmoore

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,194 @@
...
+pub struct AbsDiffResult<N: u32> {
+    correction: bool,
+    /// Note: this is the *uncorrected* absolute difference -- the `correction` term
+    /// must be added to get the true absolute difference -- use `to_corrected` to
+    /// flatten this structure to the absolute difference value.
+    uncorrected: uN[N],
+}
+
+/// Returns `true` if the first operand was larger, `false` if the second operand was larger.
+pub fn is_x_larger<N: u32>(adr: AbsDiffResult<N>) -> bool { adr.correction }
```
proppy, line 48: nit: what about `has_correction` or `is_positive` ? I feel like `x` here would not really be visible to the reader: 
- it's not really a member of the `AbsDiffResult`.
- being a "positional" argument of `abs_diff` the name `x` is only known to "people who read the doc".

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,194 @@
...
+}
+
+#[quickcheck(exhaustive)]
+fn prop_x_larger_means_correction(x: u4, y: u4) -> bool {
+    if x < y {
+        !is_x_larger(abs_diff(x, y))
+    } else if y > x {
+        is_x_larger(abs_diff(x, y))
+    } else {
+        true
```
proppy, line 155: nit `is_zero`?

xls/dslx/stdlib/abs_diff.x
```diff
@@ -0,0 +1,194 @@
...
+    }
+}
+
+#[quickcheck(exhaustive)]
```
proppy, line 159: do we want a quickcheck against the `grm`?

proppy (approved):
just some nits and qs

PR #2026, opened 2025-04-01
Fix seed-dependent proc sample generator test.

schilkp:
The test `GenerateBasicProcSample` asserted that the proc generated does not feature any channels. This is the case for the (fixed) default seed but not true in general. For example, a seed of `4` causes the test to fail:

```cpp
std::mt19937_64 rng(4);
```

This removes the seed-dependent assertions.

Also explicitly set the sample type to `proc`, instead of the default (unknown).

Pinging @hzeller since you added these assertions :)


hzeller (approved):
LGTM

dplassgit (approved):

PR #2030, opened 2025-04-03
fuzzer: gather total crashers and skipped sample stats.

schilkp:
While working on fuzzer stuff, I found it useful to track the actual rate of skipped/`known_failure` samples generated. This adds bookkeeping to count the total number of samples, skipped samples, and crashes. The multiproc runner is updated to accumulate these statistics from all workers and print the totals.


ericastor (approved):
This looks like a generally useful improvement! Thanks for doing all the necessary plumbing.

PR #2032, opened 2025-04-04
Remove dead links from README/docs.

schilkp:
Removes links that are broken and, from my understanding, no longer relevant:

- `docs`: Generated docs are no longer checked-in.
- `uncore_rtl`: Seems to not be super relevant anymore (has been moved to `contrib/ice40/uncore_rtl`)


ericastor (approved):
Thanks again!

PR #2051, opened 2025-04-23
Fix incorrect rounding in apfloat::downcast

allight:
When downcasting to a subnormal previously we first did a downcast of just the fractional part followed by the downcast of the exponent, where subnormals are generated. This led to the fraction being rounded twice, once for the downcast of just the fractional part and again for the subnormal fractional. This could lead to some values being rounded up twice, ending up 1 ulp larger than intended.


allight:
@cdleary  fyi

allight:
#2056 as review/fyi pull instead.

PR #2064, opened 2025-04-28
 Allow if without else in DSLX

### LINKED ISSUE #1980: [enhancement] Allow `if` without `else` in DSLX

Currently, you can't have an `if` without an `else` part, which is different from Rust, where an `if` without an `else` always returns a unit type:

```
error[E0317]: `if` may be missing an `else` clause
 --> src/main.rs:2:5
  |
2 | /     if true {
3 | |         3
  | |         - found here
4 | |     };
  | |_____^ expected integer, found `()`
  |
  = note: `if` expressions without `else` evaluate to `()`
  = help: consider adding an `else` block that evaluates to the expected type
```

This is a small addition with well-defined behavior that can move DSLX closer to Rust syntax.
The change is not crucial by any means. We usually write empty `else` blocks when printing data with `trace_fmt!`.

Write `if` with an empty `else`

As stated before, this is not a crucial addition to the language, but in my opinion, it could make code more readable. One thing to consider is the usage of IO operations in such statements, but this IO operations would have to ignore the returned token. Once again, this should be well-defined behavior.

rw1nkler:
This PR addresses the changes requested in #1980, enabling the use of `if` expressions without requiring an `else` branch. Additionally, it provides more detailed error messages for failing cases, helping users better understand potential issues with their code. The error message is taken from the rust compiler and resembles the one posted in the issue description (https://github.com/google/xls/issues/1980#issue-2899942121)

Here is an example of the error message, which provides explanation to the user:

* Incorrect DSLX code:
```rust
fn if_without_else_incorrect (predicate: bool) {
    if (predicate) {
        trace_fmt!("This is if with else (if branch)")
    }
}

#[test]
fn test_if_without_else_incorrect() {
    assert_eq(if_without_else_incorrect(true), ());
    assert_eq(if_without_else_incorrect(false), ());
}
```
* Error message:
```
xls/examples/if_no_else.x:42:5-45:1
0040:
0041:   fn if_without_else_incorrect (predicate: bool) {
0042:       if (predicate) {
       _____^
0043: |         trace_fmt!("This is if with else (if branch)")
0044: |     }
0045: | }
      |^ XlsTypeError: token vs (): Conditional consequent type (in the 'then' clause) did not match alternative type (in the 'else' clause); note that if expression without else from @ xls/examples/if_no_else.x:42:20-44:6 evaluates to (). Consider adding an else block that evaluates to the expected type.
0046:
0047:   #[test]
Error parsing and type checking DSLX source file: xls/examples/if_no_else.x
Target //xls/examples:if_no_else_dslx_test failed to build
```

Resolves #1980 
FYI @richmckeever

### DIFF
--- a/xls/dslx/diagnostics/maybe_explain_error.cc
+++ b/xls/dslx/diagnostics/maybe_explain_error.cc
@@ -110,6 +110,19 @@ absl::Status ExplainIfBlockWithTrailingSemi(const TypeMismatchErrorData& data,
                             file_table);
 }
 
+absl::Status ExplainIfConditionalBlockWithoutElse(
+    const TypeMismatchErrorData& data, const Conditional* conditional,
+    const FileTable& file_table) {
+  const StatementBlock* culprit =
+      dynamic_cast<const StatementBlock*>(data.lhs_node);
+  std::string message = absl::StrFormat(
+      "%s; note that if expression without else from @ %s evaluates to (). "
+      "Consider adding an else block that evaluates to the expected type.",
+      data.message, culprit->span().ToString(file_table));
+  return XlsTypeErrorStatus(data.error_span, *data.lhs, *data.rhs, message,
+                            file_table);
+}
+
 absl::Status ExplainIfConditionalBlockWithTrailingSemi(
     const TypeMismatchErrorData& data, const Conditional* conditional,
     bool lhs_is_unit, bool rhs_is_unit, const FileTable& file_table) {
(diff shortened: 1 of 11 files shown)


dplassgit:
Note: the formatter will need to be updated as well. See https://github.com/google/xls/blob/53c0dcab9c657514d0a6687e280df4aad977d9ec/xls/dslx/fmt/ast_fmt.cc#L1645-L1672

erinzmoore (changes requested):
Please update the formatter as @dplassgit mentioned in the conversation and we can retrigger the internal review. As written, I think the formatter will break on if-without-else.

rw1nkler:
> Please update the formatter as @dplassgit mentioned in the conversation and we can retrigger the internal review. As written, I think the formatter will break on if-without-else.

I updated the formatter

erinzmoore (changes requested):
Can you please add a test case without `else` to `ast_fmt_test.cc`?

rw1nkler:
> Can you please add a test case without `else` to `ast_fmt_test.cc`?

I added the requested tests

erinzmoore (approved):

proppy:
can we also updated https://google.github.io/xls/dslx_reference/#if-expression?

rw1nkler:
> can we also updated https://google.github.io/xls/dslx_reference/#if-expression?

Yes sure, I will open a separate PR for this

Issue #1867, opened 2025-01-17
[DSLX] `Cannot convert expression to parametric: SIGNED > u32:0`


dank-openai:
**Describe the bug**

I'm unable to do something with parametrics; I think it should be legal. I can't use constexprs in a struct's parametric definitions.

**To Reproduce**

```
pub struct FixedPoint<SIGNED:u32, NUM_BITS: u32,  SIGNED_BOOL: bool = {SIGNED > u32:0} > {
    data: xN[SIGNED_BOOL][NUM_BITS],
}

#[test]
fn test_instantiate_FixedPoint() {
    let a = FixedPoint<u32:0, u32:5> { data: u5:1};
    assert_eq(a.data, u5:1);
}
```

Build and _test_ the above. The error shows up during the test.

**Expected behavior**
It should work, the test should pass.

**Additional context**

If you're wondering why I made SIGNED a u32 instead of bool, see [this bug.](https://github.com/google/xls/issues/1841)

PR #2066, opened 2025-04-28
[DSLX:stdlib] Add fixed point library.

dank-openai:


xls/dslx/stdlib/fixed_point.x
```diff
@@ -0,0 +1,1633 @@
...
+//
+// TODO when https://github.com/google/xls/issues/1841 is resolved, undo the workaround
+// that changed BINARY_EXPONENT:s32 to (EXPONENT_IS_NEGATIVE:u32, BINARY_UEXPONENT: u32).
+//
+// BINARY_UEXPONENT means unsigned exponent. It is the magnitude of the binary exponent.
+//
+// TODO when https://github.com/google/xls/issues/1848 is resolved, delete the two unused
+// fields
+//
+// TODO when https://github.com/google/xls/issues/1861 is resolved, make the type
```
allight, line 46: This was fixed 10 months ago. Is it still an issue?

dank-openai: I have the wrong ID in here. The actual problem (can't make this struct sign parametric) remains AFAIK.

dank-openai: Could you please correct this? I think [this is the one I intended](https://github.com/google/xls/issues/1867)

dank-openai: Regardless, once type system v2 (and potentially some other features) land, this struct should be 
```
pub struct FixedPoint<BINARY_EXPONENT: s32, S: bool, NUM_BITS: u32> {
    significand: xN[S][NUM_BITS],  // concatenation of integer and fraction bits
}
```
so maybe add a TODO on the type system v2 issue.

PR #2067, opened 2025-04-28
[DSLX:stdlib] Add Ordering and compare

dank-openai:


xls/dslx/stdlib/std.x
```diff
@@ -608,6 +608,40 @@ fn mask_bits_test() {
...
+    Equal = 0,
+    Greater = 1,
+}
+
+// Compares two integers of the same sign and width.
+//
+// The reason to use this over a comparison operator such as `<` is that this exhaustively
+// handles all 3 cases; you can match on the result.
+pub fn compare<S: bool, N: u32>(lhs: xN[S][N], rhs: xN[S][N]) -> Ordering {
+    if lhs < rhs {
```
hongted, line 623: This isn't that efficient -- needing three comparisons plus a priority select.  Suggest utilizing subtraction and handling the sign-ness.

dank-openai: working on it...

hongted: After giving it some thought, I think reordering the conditions would be close enough to a fully custom implementation.


```
let is_equal = lhs == rhs;
let less_than = lhs < rhs;

if ( is_equal ) {
  Ordering::Equal
} else if ( less_than ) {
  Ordering::Less
} else {
  Ordering::Greater
}
```

Reduces to one subtraction, one compare, and a 2x2 logic block.

dank-openai: I came up with the following. Requires subtraction and not-equal. Let me know what version you prefer. I'll just push what I have so you can see it in whole.

ericastor: Apologies for the drive-by, but FWIW, I prefer the version @hongted came up with; two comparisons seems likely to be (slightly) cheaper than a subtraction & comparison, and the code is clearer without comments required.

It also might be a bit neater to write as:
```
if lhs == rhs {
  Ordering::Equal
} else if lhs < rhs {
  Ordering::Less
} else {
  Ordering::Greater
}
```
I'm not sure the intermediate variables are helping us. But that's VERY MUCH a nit, and a matter of personal preference alone.

hongted: I'd say the cost for the two versions should come out to a wash -- both are a subtraction and a compare.  What would hold one or the other back would be the optimization opportinitues  they each hide or reveal.   However, the one with an explicit subtraction does have the benefit of being able to trade off latency and area -- by replacing the compare with the an or-tree of 

```
let not_equal = or_reduce(diff_ext) 
```

Let's go with the more explicit implementation and we can adjust based off of analysis.

dank-openai: Side note:

> let not_equal = or_reduce(diff_ext) 

Yeah, I noticed that but my reading indicated that a dedicated comparator (!=) would probably be lowered directly to a tree of XNOR+AND gates and optimized by well by the synthesis tool. Maybe g8r would shed some light on this little digression :)

dank-openai: > Let's go with the more explicit implementation and we can adjust based off of analysis

Okay, that's what I [pushed](https://github.com/google/xls/pull/2067/commits/5ca062214cb5857b12a2bb1fa649c01ce7ef8993) the other day. PTAL. Resolve this when you're satisfied.

ericastor: I'm confused, I thought the "explicit implementation" would be the "check equality, then check a single comparison" approach, rather than the explicit subtraction. (I don't believe that that version will emit a subtractor, and comparisons are theoretically slightly cheaper due to being able to discard information.)

Which approach did you have in mind, @hongted ? (If I get a vote, I'll still vote for that one.)

dank-openai: I tried all 3 implementations (including the original which had less than and greater than). I used `s32` arguments for all and compiled them using g8r (the `[ir_to_gates](https://github.com/xlsynth/rules_xlsynth/blob/5b5f81fdd8ef46cde6a7ae2b79491b06571475cf/README.md#ir_to_gates--convert-ir-to-gate-level-analysis)` rule).

- [compare using less than and greater than](https://gist.github.com/dank-openai/6a5d6194cd805524a781cb07a4315faf#file-compare-lt-gt) (the original). Deepest Path 101, Live Node Count 533
- [compare using subtraction and equal](https://gist.github.com/dank-openai/90f67f304cd396775d742c280e6997c8#file-compare-using-subtraction). DP 99, LNC 358
- [compare using equal and less than](https://gist.github.com/dank-openai/f1da60b67f52dc66e27a87472d74aca3#file-compare-eq-lt). DP 99, LNC 361

dank-openai: Note that I _just_ added 1 more implementation and updated all the links.

* If you believe in g8r, then [compare using subtraction and not equal](https://gist.github.com/dank-openai/1202eabb63bbeccd0c557845835451a3) with DP 99, LNC 358 is the clear winner, by a little bit
* Interesting that the original, with `<` and `> ` is not Pareto-worse than `-` and `or_reduce`
* using `=` and `<` is more intuitive (IMO but I think you'd agree), and only a bit worse on live node count, than `-` and `!=`

hongted: Merging is in progress.  Ideally we'd be able to combine the equality and comparison together and build up the comparator recursively a tree or ripple fashion like adders depending on the architecture.

Even then I think whatever is going to be the best way will depend on exposing the right ops so that the synthesis datapath optimizer can work it's best.

Issue #1352, opened 2024-03-15
[Enhancement]: add option to convert assert IR nodes into async-fatal-signal hardware


cdleary:
### What's hard to do? (limit 100 words)

Right now `fail!`/`assert!` in the DSL only become assert ops in the IR that turn into Verilog/SV assertions. On some chips there are async fatal signals that a block can propagate. A useful transform could be to identify assertion nodes in the IR and turn them into async fatal hardware automatically.

### Current best alternative workaround (limit 100 words)

Currently users could create an explicit fatal signal as output using the same predicates they use in assertion operations.

### Your view of the "best case XLS enhancement" (limit 100 words)

Having the option to transform the assert IR operations automatically for either "all assert ops" or "some identified subset of assert ops" in the design.

PR #2109, opened 2025-05-03
Allow #[test_proc] to expect specific assertion failures

### LINKED ISSUE #1071: [DSLX] Feature: Add `#[test_fail]` and `#[test_proc_fail]`
Right now `#[test_proc]` and `#[test]` always has to succeed.

Adding new language testing feature will allow for negative testing.
I think it would only require adding  `#[test_fail]` and `#[test_proc_fail]` to the language parser.
Test procs or fn should keep their current type format, and the bulk of the changes would be
applied to the DSLX interpreter.

rw1nkler:
This commit introduces the ability to specify that a test proc is expected to fail on a particular `assert!` or `fail!` in the DSLX interpreter. The `#[test_proc]` attribute accepts an optional `expected_fail` parameter. with the label marking the failure. If a failure occurs with the matching label, the test is treated as successful.

The choice to implement this as an attribute, rather than a built-in, as discussed in https://github.com/google/xls/issues/1071#issuecomment-1642453871 was motivated by cases where there's no clear location to “catch” the failure. For example, a test that spawns a proc containing only a single `fail!` statement.

The functionality is implemented by storing the `expected_fail` value in the `TestProc` object. When an `assert!` or `fail!` is handled in the DSLX interpreter, the label of the failure is attached to the error using an `absl::Status` payload. This is then compared in `RunDslxTestProc` against the stored label in the `TestProc`.

Resolves #1071

### DIFF
--- a/docs_src/tutorials/how_to_use_procs.md
+++ b/docs_src/tutorials/how_to_use_procs.md
@@ -217,6 +217,12 @@ test proc is very similar to a normal proc with the following changes:
     output channel for terminating interpretation. When the test is complete,
     the proc should send the test's status (`true` on success, `false` on
     failure) on that channel (commonly called the "terminator" channel).
+*   The `#[test_proc]` attribute also supports an optional `expected_fail`
+    parameter. This allows writing tests that are expected to fail due to
+    a specific `assert!` or `fail!` condition. The value must match the label
+    on the failing assertion. If a failure with the matching label occurs,
+    the test is treated as successful, and if no such failure happens,
+    the test is treated as failed.
 
 A skeletal example:
 
--- a/xls/dslx/errors.cc
+++ b/xls/dslx/errors.cc
@@ -33,6 +33,8 @@
 namespace xls::dslx {
 namespace {
 
+constexpr std::string_view kLabelPayloadName = "label";
+
 template <typename TypeOrAnnotation>
 absl::Status TypeInferenceErrorStatusInternal(
     const Span& span, const TypeOrAnnotation* type_or_annotation,
@@ -61,6 +63,25 @@ absl::Status FailureErrorStatus(const Span& span, std::string_view message,
       "FailureError: %s The program being interpreted failed!%s%s",
       span.ToString(file_table),
       message.empty() || message[0] == '\n' ? "" : " ", message));
+}
+
+absl::Status FailureErrorStatusForAssertion(const Span& span,
+                                            std::string_view label,
+                                            std::string_view message,
+                                            const FileTable& file_table) {
+  absl::Status error = FailureErrorStatus(span, message, file_table);
+  error.SetPayload(kLabelPayloadName, absl::Cord(label));
+  return error;
+}
+
+std::optional<std::string> GetAssertionLabelFromError(
+    const absl::Status& status) {
+  std::optional<absl::Cord> label_opt = status.GetPayload(kLabelPayloadName);
+  if (label_opt.has_value()) {
+    return std::string(*label_opt);
+  } else {
+    return std::nullopt;
+  }
 }
 
 absl::Status ProofErrorStatus(const Span& span, std::string_view message,
--- a/xls/dslx/frontend/parser.cc
+++ b/xls/dslx/frontend/parser.cc
@@ -815,8 +815,29 @@ Parser::ParseAttribute(absl::flat_hash_map<std::string, Function*>* name_to_fn,
         peek->span(), absl::StrCat("Invalid test type: ", peek->ToString()));
   }
   if (attribute_name == "test_proc") {
+    std::optional<std::string> expected_fail_label = std::nullopt;
+    XLS_ASSIGN_OR_RETURN(const Token* peek, PeekToken());
+    if (peek->kind() == TokenKind::kOParen) {
+      XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kOParen));
+      XLS_ASSIGN_OR_RETURN(Token parameter_name,
+                           PopTokenOrError(TokenKind::kIdentifier));
+
+      if (parameter_name.GetStringValue() == "expected_fail_label") {
+        XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kEquals));
+        XLS_ASSIGN_OR_RETURN(Token parameter_value,
+                             PopTokenOrError(TokenKind::kString));
+        expected_fail_label = parameter_value.GetStringValue();
+      } else {
+        return ParseErrorStatus(
+            attribute_tok.span(),
+            absl::StrFormat(
+                "Unknown parameter name in the #[test_proc] attribute: '%s'",
+                parameter_name.ToString()));
+      }
+      XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kCParen));
+    }
     XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kCBrack));
-    return ParseTestProc(bindings);
+    return ParseTestProc(bindings, expected_fail_label);
   }
   return ParseErrorStatus(
       attribute_tok.span(),
@@ -3965,7 +3986,8 @@ absl::StatusOr<TestFunction*> Parser::ParseTestFunction(
   return tf;
 }
 
-absl::StatusOr<TestProc*> Parser::ParseTestProc(Bindings& bindings) {
+absl::StatusOr<TestProc*> Parser::ParseTestProc(
+    Bindings& bindings, std::optional<std::string> expected_fail_label) {
   XLS_ASSIGN_OR_RETURN(ModuleMember m,
                        ParseProc(GetPos(), /*is_public=*/false, bindings));
   if (!std::holds_alternative<Proc*>(m)) {
@@ -3987,7 +4009,7 @@ absl::StatusOr<TestProc*> Parser::ParseTestProc(Bindings& bindings) {
   }
 
   // Verify no state or config args
-  return module_->Make<TestProc>(p);
+  return module_->Make<TestProc>(p, expected_fail_label);
 }
 
 const Span& GetSpan(
--- a/xls/dslx/frontend/parser.h
+++ b/xls/dslx/frontend/parser.h
@@ -619,7 +619,8 @@ class Parser : public TokenParser {
   absl::StatusOr<TestFunction*> ParseTestFunction(Bindings& bindings,
                                                   const Span& attribute_span);
 
-  absl::StatusOr<TestProc*> ParseTestProc(Bindings& bindings);
+  absl::StatusOr<TestProc*> ParseTestProc(
+      Bindings& bindings, std::optional<std::string> expected_fail_label);
 
   // Parses a constant definition (e.g. at the top level of a module). Token
   // cursor should be over the `const` keyword.
--- a/xls/dslx/frontend/proc.cc
+++ b/xls/dslx/frontend/proc.cc
@@ -154,6 +154,10 @@ std::string ProcLike::ToString() const {
 TestProc::~TestProc() = default;
 
 std::string TestProc::ToString() const {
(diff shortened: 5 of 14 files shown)


docs_src/tutorials/how_to_use_procs.md
```diff
@@ -217,6 +217,11 @@ test proc is very similar to a normal proc with the following changes:
     input channel for terminating interpretation. When the test is complete, the
     proc should send the test's status (`true` on success, `false` on failure)
     on that channel (commonly called the "terminator" channel).
+*   The `#[test_proc]` attribute also supports an optional `expected_fail`
+    parameter. This allows writing tests that are expected to fail due to
+    a specific `assert!` or `fail!` condition. The value must match the label
+    on the failing assertion. If a failure with the matching label occurs,
+    the test is treated as successful.
```
proppy, line 224: "and if no failure happens: the test is treated as failed" (or add "if and only if")

proppy:
curious if that's something that regular SystemVerilog/UVM support, and what this would ultimately map to if we were generating SV test benches from DSLX tests.

rw1nkler:
> curious if that's something that regular SystemVerilog/UVM support, and what this would ultimately map to if we were generating SV test benches from DSLX tests.

I believe there is no built-in option to ignore the `$fatal` statement in SystemVerilog. However, if support for `fail!` and `assert!` is extended to allow the generation of a signal upon failure (as proposed in [google/xls#1352](https://github.com/google/xls/issues/1352)), the testbench could include logic to respond accordingly by either triggering a failure or allowing the test to pass based on the context.

meheff:
@richmckeever mind having a look at this as it touches on the frontend. The feature is very nice to have.

docs_src/tutorials/how_to_use_procs.md
rw1nkler: fixed

xls/dslx/frontend/parser.cc
```diff
@@ -815,8 +815,30 @@ Parser::ParseAttribute(absl::flat_hash_map<std::string, Function*>* name_to_fn,
...
+    if (peek->kind() == TokenKind::kOParen) {
+      XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kOParen));
+      XLS_ASSIGN_OR_RETURN(Token parameter_name,
+                           PopTokenOrError(TokenKind::kIdentifier));
+
+      if (parameter_name.GetStringValue() == "expected_fail") {
+        XLS_RETURN_IF_ERROR(DropTokenOrError(TokenKind::kEquals));
+        XLS_ASSIGN_OR_RETURN(Token parameter_value,
+                             PopTokenOrError(TokenKind::kString));
+        expected_fail =
```
richmckeever, line 829: I think you can just assign this without explicitly constructing the optional.

xls/dslx/frontend/parser.cc
```diff
@@ -815,8 +815,30 @@ Parser::ParseAttribute(absl::flat_hash_map<std::string, Function*>* name_to_fn,
         peek->span(), absl::StrCat("Invalid test type: ", peek->ToString()));
   }
   if (attribute_name == "test_proc") {
+    std::optional<std::string> expected_fail = std::nullopt;
```
richmckeever, line 818: This should probably be a std::optional<bool> and we convert the string (if present) to a bool in here.

xls/dslx/frontend/proc.cc
```diff
@@ -154,6 +154,10 @@ std::string ProcLike::ToString() const {
 TestProc::~TestProc() = default;
 
 std::string TestProc::ToString() const {
+  if (expected_fail_.has_value()) {
+    return absl::StrFormat("#[test_proc(expected_fail=\"%s\")]\n%s",
+                           expected_fail_.value(), proc_->ToString());
```
richmckeever, line 159: Prefer `*expected_fail_` rather than `expected_fail.value()` generally when accessing optionals.

xls/dslx/frontend/proc.h
```diff
@@ -203,7 +203,9 @@ class TestProc : public AstNode {
  public:
   static std::string_view GetDebugTypeName() { return "test proc"; }
 
-  TestProc(Module* owner, Proc* proc) : AstNode(owner), proc_(proc) {}
+  TestProc(Module* owner, Proc* proc,
+           std::optional<std::string> expected_fail = std::nullopt)
```
richmckeever, line 207: Let's take this as a std::optional<bool> here and store it that way; propagating the string seems more error prone. If someday it's more than a boolean, we can use an enum or struct at that point.

xls/dslx/run_routines/run_routines.cc
```diff
@@ -202,9 +202,24 @@ absl::Status RunDslxTestProc(ImportData* import_data, TypeInfo* type_info,
       hierarchy_interpreter->GetInterfaceChannel(0);
 
   // Run until a single output appears in the terminal channel.
-  XLS_RETURN_IF_ERROR(
+  absl::Status status =
       hierarchy_interpreter->TickUntilOutput({{terminal_channel_name, 1}})
-          .status());
+          .status();
+  auto label_opt = GetAssertionLabelFromError(status);
+  if (status != absl::OkStatus() && tp->expected_fail().has_value() &&
```
richmckeever, line 209: Prefer `!status.ok()` instead of comparing to OkStatus().

xls/dslx/run_routines/run_routines.cc
```diff
@@ -202,9 +202,24 @@ absl::Status RunDslxTestProc(ImportData* import_data, TypeInfo* type_info,
...
+      label_opt.has_value()) {
+    std::string label = label_opt.value();
```
richmckeever, line 211: Here and below, use * instead of .value().

xls/dslx/errors.cc
```diff
@@ -33,6 +33,8 @@
 namespace xls::dslx {
 namespace {
 
+static constexpr std::string_view label_payload_name = "label";
```
richmckeever, line 36: The conventional name here would be kLabelPayloadName. Also drop `static` because being in an anonymous namespace has the same effect.

xls/dslx/errors.cc
```diff
@@ -63,6 +65,25 @@ absl::Status FailureErrorStatus(const Span& span, std::string_view message,
...
+                                            std::string_view message,
+                                            const FileTable& file_table) {
+  absl::Status error = FailureErrorStatus(span, message, file_table);
+  error.SetPayload(label_payload_name, absl::Cord(label));
+  return error;
+}
+
+std::optional<std::string> GetAssertionLabelFromError(
+    const absl::Status& status) {
+  auto label_opt = status.GetPayload(label_payload_name);
```
richmckeever, line 79: Prefer writing the actual type instead of `auto`, except for iterators, destructured tuples, and objects that don't have a practically expressible type.

xls/dslx/run_routines/run_routines_test.cc
```diff
@@ -533,6 +533,37 @@ proc doomed {
               IsTestResult(TestResult::kParseOrTypecheckError, 0, 0, 0));
 }
 
+TEST_P(RunRoutinesTest, TestProcWithExpectedFail) {
```
richmckeever, line 536: Let's test this with both fail! and assert! in 2 separate tests (this one is really the assert! one).

richmckeever (changes requested):

xls/dslx/frontend/parser.cc
rw1nkler: I'm not sure if I understand you correctly.

The string is used later when comparing the label of the expected failure with the actual label from the assertion. This way we can still fail on assertions different then the expected one. Substituting it with a bool can cause false positives in tests.

The comparison takes place in: 
https://github.com/google/xls/pull/2109/files#diff-4b75394b54655a9c1d63cc7c6e63bee976945b37697c3a268df6a51e5dae59b6R213

xls/dslx/frontend/proc.h
rw1nkler: The explanation for putting here a string instead of bool is described in https://github.com/google/xls/pull/2109/files#r2088996980

xls/dslx/frontend/parser.cc
richmckeever: Yeah, I think I was thrown off by the variable name. It might be good to name it "expected_fail_label" both here and in the AST.

rw1nkler: I changed the name to ` expected_fail_label` as it was suggested

xls/dslx/frontend/proc.cc
rw1nkler: fixed

xls/dslx/run_routines/run_routines.cc
rw1nkler: fixed

xls/dslx/run_routines/run_routines.cc
rw1nkler: fixed

xls/dslx/errors.cc
rw1nkler: fixed

xls/dslx/run_routines/run_routines_test.cc
rw1nkler: I added a separate test for the `fail!` macro

rw1nkler:
I also modified the DSLX formater accordingly

xls/dslx/run_routines/run_routines_test.cc
```diff
@@ -533,6 +533,68 @@ proc doomed {
               IsTestResult(TestResult::kParseOrTypecheckError, 0, 0, 0));
 }
 
+TEST_P(RunRoutinesTest, TestProcExpectedToFailOnAssert) {
```
richmckeever, line 536: Let's add one more test, where we assert with an unexpected label and the test result is a failure.

rw1nkler: I added separate tests for both `assert!` and `fail!` macros that check whether the failure on incorrect label is treated correctly.

richmckeever (approved):

xls/dslx/errors.cc
```diff
@@ -63,6 +65,25 @@ absl::Status FailureErrorStatus(const Span& span, std::string_view message,
...
+                                            std::string_view message,
+                                            const FileTable& file_table) {
+  absl::Status error = FailureErrorStatus(span, message, file_table);
+  error.SetPayload(kLabelPayloadName, absl::Cord(label));
+  return error;
+}
+
+std::optional<std::string> GetAssertionLabelFromError(
+    const absl::Status& status) {
+  absl::optional<absl::Cord> label_opt = status.GetPayload(kLabelPayloadName);
```
proppy, line 79: std::optional?

rw1nkler: fixed, but we cannot return `status.GetPayload(kLabelPayloadName) directly`

xls/dslx/frontend/parser.h
```diff
@@ -619,7 +619,9 @@ class Parser : public TokenParser {
   absl::StatusOr<TestFunction*> ParseTestFunction(Bindings& bindings,
                                                   const Span& attribute_span);
 
-  absl::StatusOr<TestProc*> ParseTestProc(Bindings& bindings);
+  absl::StatusOr<TestProc*> ParseTestProc(
+      Bindings& bindings,
+      std::optional<std::string> expected_fail_label = std::nullopt);
```
dplassgit, line 624: Optional: ` = std::nullopt` isn't needed, since all the uses already specify the fail label.

rw1nkler:
Please let me know if I need to update anything else

dplassgit (approved):

proppy (approved):

proppy:
should we also add support for `excepted_fail` for regular function `#[test]`?

PR #2110, opened 2025-05-03
[c_api] More IR builder APIs

cdleary:
This was a change on xlsynth branch for a while to enable some Rust crate APIs -- wanted to make sure we get it into upstream instead of letting it stay as a divergence. cc @meheff


meheff (approved):
Thanks!

meheff:
I get the following error building it.  I'll fix it on the inside, but we should add this as an error in our oss build:

```
[third_party/xls/public/c_api_test.cc:1708](https://cs.corp.google.com/piper///depot/google3/third_party/xls/public/c_api_test.cc?l=1708&ws=tap-presubmit-global-server/19673474&snapshot=2):13: error: unused variable 'tuple_members' [-Werror,-Wunused-variable]
 1708 |   xls_type* tuple_members[] = {u8_arr3, u8,
```

Issue #1524, opened 2024-07-31
[enhancement]  Support test utility methods


dplassgit:
### What's hard to do? (limit 100 words)

Let me write a test helper method that does not get lowered to SV, e.g.,

```
#[cfg(test)]
fn assert_value_is_0<N: u32>(a: uN[N]) {
  assert_eq(0, a);
}

#[test]
fn test_zeros() {
  assert_value_is_0(u1:0);
  assert_value_is_0(u2:0);
}
```

### Current best alternative workaround (limit 100 words)

The helper code needs to be in-line in the `#[test]` method. It can be repetitive.

### Your view of the "best case XLS enhancement" (limit 100 words)

A new annotation `#[cfg(test)]` (which is what Rust has, evidently) could be used to tag functions that are test-support but not run as part of a test suite.

PR #2168, opened 2025-05-13
Add support for #[cfg(test)] attribute

rw1nkler:
# Description

This commit adds a `#[cfg(test)]` attribute that you can use to mark functions or procs that are only meant for testing. When marked this way, they won't be lowered to IR.  The original request in #1524 was just for functions, but I applied the same idea to procs as well.

# Implementation details

This feature uses a `test_only` flag in the `Function` class to mark functions and procs that are only used for tests. When converting code to IR, functions and procs  marked as test-only are skipped, unless the `convert_tests` option is enabled. This allows test helpers to be included in the IR if needed.

IR conversion now provides clearer error messages. If a test-only function or proc is selected for conversion without the `convert_tests` option, a dedicated error is shown. Same happens when a test-only function is called from a regular function or proc and IR is being dumped.

In DSLX, the interpreter will fail if a test-only function is called from a regular one. This check is not yet in place for procs, as communication between regular and test-only procs involves more complexity, but I can add it too if needed.

Resolves https://github.com/google/xls/issues/1524

### DIFF
--- a/xls/dslx/fmt/ast_fmt.cc
+++ b/xls/dslx/fmt/ast_fmt.cc
@@ -2209,7 +2209,7 @@ DocRef Formatter::Format(const Function& n, bool is_test) {
                             arena_.MakeText("\")]"),
                             arena_.hard_line(),
                         }));
-  } else if (n.test_only() && !is_test) {
+  } else if (n.is_test_utility() && !is_test) {
     fn_pieces.push_back(
         ConcatN(arena_, {
                             arena_.MakeText("#"),
@@ -2248,7 +2248,7 @@ DocRef Formatter::Format(const ProcMember& n) {
 
 DocRef Formatter::Format(const Proc& n, bool is_test) {
   std::vector<DocRef> attribute_pieces;
-  if (n.test_only() && !is_test) {
+  if (n.is_test_utility() && !is_test) {
     attribute_pieces.push_back(
         ConcatN(arena_, {
                             arena_.MakeText("#"),
--- a/xls/dslx/fmt/ast_fmt.cc
+++ b/xls/dslx/fmt/ast_fmt.cc
@@ -48,6 +48,7 @@
 #include "xls/dslx/frontend/ast.h"
 #include "xls/dslx/frontend/ast_cloner.h"
 #include "xls/dslx/frontend/ast_node.h"
+#include "xls/dslx/frontend/ast_utils.h"
 #include "xls/dslx/frontend/comment_data.h"
 #include "xls/dslx/frontend/module.h"
 #include "xls/dslx/frontend/pos.h"
@@ -2133,7 +2134,7 @@ DocRef Formatter::Format(const ParametricBinding* n) {
   return Format(*n);
 }
 
-DocRef Formatter::Format(const Function& n) {
+DocRef Formatter::Format(const Function& n, bool is_test) {
   std::vector<DocRef> signature_pieces;
   if (n.is_public()) {
     signature_pieces.push_back(arena_.Make(Keyword::kPub));
@@ -2208,6 +2209,15 @@ DocRef Formatter::Format(const Function& n) {
                             arena_.MakeText("\")]"),
                             arena_.hard_line(),
                         }));
+  } else if (n.test_only() && !is_test) {
+    fn_pieces.push_back(
+        ConcatN(arena_, {
+                            arena_.MakeText("#"),
+                            arena_.obracket(),
+                            arena_.MakeText(std::string(kCfgTestAttr)),
+                            arena_.cbracket(),
+                            arena_.hard_line(),
+                        }));
   }
 
   if (n.body()->empty()) {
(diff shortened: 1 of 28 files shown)


rw1nkler:
FYI @richmckeever @dplassgit

proppy:
@rw1nkler can you add DSLX examples w/ tests functions?

proppy:
> #[cfg(test)]

This does seems slightly inconsistent w/ our current annotations `#[test] and #[test_proc]`, should we switch everything to `#[cfg(something)]` or add something like `#[test_aux/helper]` and `#[test_proc_aux/helper]`.

allight:
> > #[cfg(test)]
> 
> This does seems slightly inconsistent w/ our current annotations `#[test] and #[test_proc]`, should we switch everything to `#[cfg(something)]` or add something like `#[test_aux/helper]` and `#[test_proc_aux/helper]`.

This is consistent with how rust does things (sort of, #[cfg(whatever)] acts more like a #ifdef and style is to place it on the 'testing' module but its pretty close): [rust docs](https://doc.rust-lang.org/book/ch11-01-writing-tests.html)

proppy:
> This is consistent with how rust does things (sort of, #[cfg(whatever)] acts more like a #ifdef and style is to place it on the 'tests' module but its pretty close): rust docs

Ah you're right that's close to what Rust does:
```
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn it_works() {
        let result = add(2, 2);
        assert_eq!(result, 4);
    }
}
```
I guess one difference is that because they have `mod` they can enable `cfg[test)` for a whole hierarchy and don't have to annotate everything single helper functions/procs.

Maybe worth logging as a issue if we want something similar as a follow-up.

xls/dslx/frontend/ast.h
```diff
@@ -2282,6 +2283,8 @@ class Function : public AstNode {
     disable_format_ = disable_format;
   }
   bool disable_format() const { return disable_format_; }
+  void set_used_in_tests(bool used_in_tests) { used_in_tests_ = used_in_tests; }
```
cdleary, line 2286: I think this is not "used in tests" conceptually more like "test only"

cdleary (commented):
Looks good, is there a test for what happens when we refer to an entity that's test-only though in a non-test scenario? Seems like it'd be nicer to make a DSLX level error message instead of having a linker-style error at IR conversion/opt time.

xls/dslx/frontend/ast.h
rw1nkler: Yes, you're right. I changed that

rw1nkler:
> Looks good, is there a test for what happens when we refer to an entity that's test-only though in a non-test scenario? Seems like it'd be nicer to make a DSLX level error message instead of having a linker-style error at IR conversion/opt time.

There's a test that checks if a test utility function was called from a regular function or proc:
https://github.com/google/xls/pull/2168/files#diff-c24b2f53453f91215eec0d6d76daf6a9e03f9c780d0f9710472388af1fe1ce11R1686

Right now, there are no checks for spawning utility procs inside regular procs.
If we want to catch those cases too, I can work on adding that.

mikex-oss:
FYI @rw1nkler looks like you'll need to rebase to get this in.

xls/dslx/fmt/ast_fmt.cc
```diff
@@ -2208,6 +2208,11 @@ DocRef Formatter::Format(const Function& n) {
                             arena_.MakeText("\")]"),
                             arena_.hard_line(),
                         }));
+  } else if (n.test_only() && !is_test) {
+    fn_pieces.push_back(ConcatN(arena_, {
+                                            arena_.MakeText("#[cfg(test)]"),
```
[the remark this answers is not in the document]
rw1nkler: I added the constant

xls/dslx/fmt/ast_fmt.cc
```diff
@@ -2208,6 +2209,15 @@ DocRef Formatter::Format(const Function& n) {
                             arena_.MakeText("\")]"),
                             arena_.hard_line(),
                         }));
+  } else if (n.test_only() && !is_test) {
+    fn_pieces.push_back(
+        ConcatN(arena_, {
+                            arena_.MakeText("#"),
```
dplassgit, line 2215: The proc version inserts the whole attribute at the same time. Is there a benefit either way?

xls/dslx/frontend/ast.h
```diff
@@ -2376,6 +2376,8 @@ class Function : public AstNode {
     disable_format_ = disable_format;
   }
   bool disable_format() const { return disable_format_; }
+  void set_test_only(bool test_only) { test_only_ = test_only; }
```
dplassgit: If at all possible this should be done in the constructor instead of in a mutator

rw1nkler: I wanted to be consistent with other attributes like `dslx_format_disable` or `extern_verilog`. If using a constructor  is preferred I can change that.

dplassgit: Yes, having the bit in the constructor is preferred. The other mutator methods are only present out of necessity due to quirks in the parser.

rw1nkler: I moved the parameter to a constructor. Let me know if this is what you expected.

dplassgit: Looks good now

rw1nkler:
Please let me know if there's anything else I should address to move this forward

rw1nkler:
I rebased this code. PTAL @proppy

proppy (approved):

PR #2177, opened 2025-05-14
[cleanup] Switch uses of L and LL suffixes to be explicitly typed for x-platform portability

cdleary:
This was breaking the OS X build on integrate from upstream.

I think the recommendation in the style guide is to avoid integer type suffxes but they can easily creep in since there's not a tidy for them.


meheff (approved):
Thanks for fixing these.

Issue #2209, opened 2025-05-18
[codegen] multi-driver / multi-decl: suspected symbol hygiene violation in generated verilog


cdleary:
**Describe the bug**
If you switch the multiprocess fuzzer to use `p` as the gensym prefix instead of `x` you observe a lot of problems quickly.

With this .x:
```
fn main(p0: u29, p1: u33, p2: u44, p3: u32, p4: u41, p5: u33, p6: u43) -> (u29, u45, u45, u29, u15, u32, u45, u33) {                                                                       
    { 
        let p7: u15 = u15:0x2000;                                                                                                                                                                                 
        let p8: u44 = !p2;
        let p9: u33 = -p5;
        let p10: u29 = ctz(p0);
        let p11: u44 = -p8;
        let p12: u29 = bit_slice_update(p0, p8, p7);                                                                                                                                                              
        let p13: u29 = signex(p1, p10);
        let p14: u45 = u45:0x682_1b1c_101d;
        let p15: u32 = p3[:];
        let p16: u45 = p14[0+:u45];                                                                                                                                                                               
        (p10, p14, p14, p12, p7, p15, p16, p9)                                                                                                                                                                    
    }
} 
```

We get the following opt IR:
```
package sample
    
file_number 0 "/tmp/temp_directory_8R9TAz/sample.x"                                                                                                                                                               
        
top fn __sample__main(p0: bits[29] id=1, p1: bits[33] id=2, p2: bits[44] id=3, p3: bits[32] id=4, p4: bits[41] id=5, p5: bits[33] id=6, p6: bits[43] id=7) -> (bits[29], bits[45], bits[45], bits[29], bits[15], bits[32], bits[45], bits[33]) {
  one_hot.11: bits[30] = one_hot(p0, lsb_prio=true, id=11, pos=[(0,5,26)])
  literal.29: bits[24] = literal(value=0, id=29, pos=[(0,5,26)])
  encode.12: bits[5] = encode(one_hot.11, id=12)
  p8: bits[44] = not(p2, id=9, pos=[(0,3,22)])
  p7: bits[15] = literal(value=8192, id=8, pos=[(0,2,22)])
  p10: bits[29] = concat(literal.29, encode.12, id=30, pos=[(0,5,26)])
  p12: bits[29] = bit_slice_update(p0, p8, p7, id=15, pos=[(0,7,39)])                                                                                                                                             
[...]

PR #2210, opened 2025-05-19
[vast] Error if a duplicate name is used for Wire/Reg/Integer/etc in module scope.

cdleary:
See #2209 -- doing this first to prevent silent errors that make it down through synthesis.

This is not the root cause fix, we will need that as a followup, but this will prevent symbol conflicts from making it though codegen since downstream tools are often happy to accept arbitrary malformed-ness.


cdleary:
> Thanks for the bug fix!

Note this is just the detection to prevent the situation from getting successful codegen, we still need a root cause fix on top of this.

PR #2217, opened 2025-05-20
Document that `else` in `if` is optional

rw1nkler:
This commit updates the docs to note that `else` can be omitted in `if` expressions.
These changes are a follow-up to https://github.com/google/xls/pull/2064, and extends the documentation as requested in https://github.com/google/xls/pull/2064#issuecomment-2876794413

FYI @proppy


ericastor (approved):
Thanks for following up with documentation!

PR #2219, opened 2025-05-20
[DSLX, stdlib] add sat_sub, prefer const_assert

dank-openai:
- Add saturating subtraction for unsigned arguments and unsigned result.
  - bounded_minus_1 now uses sat_sub
- Replace assert in {split_msbs, split_lsbs} with const_assert. This matches the original intention: ensures well-formed type-level arguments.

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -1214,7 +1214,9 @@ fn test_vslice() {
 // This function ensures that all bits of the argument are used.
 pub fn split_msbs<N: u32, X: u32, Z: u32 = {X - N}, FROM_START: s32 = {Z as s32}>
     (x: bits[X]) -> (bits[N], bits[Z]) {
-    assert!(N <= X, "split_msbs_requires_n_less_equal_x");
+    // Can't split more bits than exist
+    const_assert!(N <= X);
+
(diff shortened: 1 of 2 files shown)


dank-openai:
Should bounded_minus_1 be kept? Or should all users migrate their calls to `sat_sub(x, uN[N]:1)`?

xls/dslx/stdlib/std.x
```diff
@@ -170,6 +170,56 @@ fn sadd_test() {
     assert_eq(s4:-8, sadd(s3:-4, s3:-4));
 }
 
+// Saturating subtraction: returns x - y, saturating at 0 if x < y.
+pub fn sat_sub<N: u32>(x: uN[N], y: uN[N]) -> uN[N] { if x < y { uN[N]:0 } else { x - y } }
```
proppy, line 174: alt: should we name this `bounded_sub` ?

ericastor: This operation actually has several standard mathematical names we could consider: https://en.wikipedia.org/wiki/Monus#Natural_numbers

I'd suggest `difference_or_zero` or `truncated_subtraction`, but I also don't care THAT much.

PR #2241, opened 2025-05-22
Cdleary/2025 05 22 type get flat bit count

cdleary:


xls/public/c_api_test.cc
```diff
@@ -286,17 +286,6 @@ TEST(XlsCApiTest, ParseTypedValueAndFreeIt) {
   xls_value_free(value);
 }
 
-// Helper for invoking APIs where we want to immediately free the C-API-provided
```
proppy, line 289: q: was this dead code?

PR #2251, opened 2025-05-23
[ir] strict verification for comparison operations

cdleary:
* non-equality cmp comparisons (ult, slt, etc.) now strictly require `bits` type operands, aligning verifier with `ir_semantics.md`
*  equality comparisons (eq, ne) now disallow `Token` type operands in the verifier

### DIFF
--- a/xls/ir/verify_node.cc
+++ b/xls/ir/verify_node.cc
@@ -551,31 +551,31 @@ class NodeChecker : public DfsVisitor {
 
   absl::Status HandleEq(CompareOp* eq) override {
     XLS_RETURN_IF_ERROR(ExpectOperandCount(eq, 2));
-    XLS_RETURN_IF_ERROR(ExpectOperandsSameBitsType(eq));
+    XLS_RETURN_IF_ERROR(ExpectOperandsSameTypeNoToken(eq));
     return ExpectHasBitsType(eq, /*expected_bit_count=*/1);
   }
 
   absl::Status HandleUGe(CompareOp* ge) override {
     XLS_RETURN_IF_ERROR(ExpectOperandCount(ge, 2));
-    XLS_RETURN_IF_ERROR(ExpectOperandsSameBitsType(ge));
+    XLS_RETURN_IF_ERROR(ExpectOperandsAreBitsAndSameWidth(ge));
     return ExpectHasBitsType(ge, /*expected_bit_count=*/1);
   }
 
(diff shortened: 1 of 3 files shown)


xls/ir/verify_node.cc
```diff
@@ -1465,12 +1465,41 @@ class NodeChecker : public DfsVisitor {
...
+  absl::Status ExpectOperandsAreBitsAndSameWidth(Node* node) const {
     if (node->operand_count() == 0) {
       return absl::OkStatus();
     }
+    XLS_RETURN_IF_ERROR(ExpectOperandHasBitsType(node, 0));
     Type* type = node->operand(0)->GetType();
+    // Already checked operand 0 is BitsType. Now check other operands.
+    for (int64_t i = 1; i < node->operand_count(); ++i) {
+      XLS_RETURN_IF_ERROR(ExpectOperandHasBitsType(node, i));
+      if (node->operand(i)->GetType()->AsBitsOrDie()->bit_count() !=
```
proppy, line 1479: nit: couldn't you use `node->operand(i)->BitCountOrDie()` ?

xls/ir/verify_node.cc
```diff
@@ -1465,12 +1465,41 @@ class NodeChecker : public DfsVisitor {
...
+          type->AsBitsOrDie()->bit_count()) {
+        return absl::InternalError(absl::StrFormat(
+            "Expected operand %d of %s to have bit count %d (same as operand "
+            "0), has bit count %d: %s",
+            i, node->GetName(), type->AsBitsOrDie()->bit_count(),
+            node->operand(i)->GetType()->AsBitsOrDie()->bit_count(),
```
proppy, line 1485: nit: couldn't you use `node->operand(i)->BitCountOrDie()` ?

proppy (approved):

xls/ir/verify_node.cc
cdleary: I have a slight preference for this way, since it's is showing you the type is indeed BitsType after the call to `ExpectOperandHasBitsType` succeeds. Tbh it's kind of weird BitCountOrDie assumes the node has a BitsType output, looking at it we might want to remove that since it's not really a sound polymorphic assumption.

xls/ir/verify_node.cc
cdleary: Ditto.

PR #1654, opened 2024-10-09
modules/zstd: Rework ZSTD Decoder

lpawelcz:
This PR reworks the existing ZSTD Decoder to memory-based approach instead of relying on stream-based architecture.

The PR relies on peripherals (`MemReader` and `MemWriter`) for accessing external memory through AXI interface that are introduced in: https://github.com/google/xls/pull/1613. It is also based on https://github.com/google/xls/pull/1616.

The Decoder is able to decompress ZSTD frames that consist of RAW and RLE blocks. This is tested in a verilog simulation with a [cocotb](https://github.com/cocotb/cocotb) testbench.

The changes in this PR include:

* Rewriting `ZstdDecoder` proc to work with memory-based approach
* Adding verilog-level tests of the `ZstdDecoder`
  * Introducing python library for generating ZSTD frames with `decodecorpus`
  * Adding python bindings for ZSTD reference library ([zstandard](https://pypi.org/project/zstandard/)) used for crosschecking decoding results in the simulation
  * Adding cocotb testbench that generates multiple ZSTD frames with RAW and RLE blocks and decodes those with simulated ZstdDecoder
* Reworking `FrameHeaderDecoder`, `BlockHeaderDecoder`, `Raw-` and `RleBlockDecoders` to use memory-based approach instead of streaming input data from the input of the decoder
* Adding `CsrConfig` and `AxiCsrAccessor` procs that implement the internal registers of the `ZstdDecoder` and provide external (AXI) and internal (native) facing interfaces for accessing those
* Simplifying:
  * `RleBlockDecoder` proc
  * `BlockHeader` library
* Removing components not used in memory-based approach:
   * magic number functions
   * `BlockDecoder` proc
   * `DecDemux` proc
* Added `common_codegen_args` list to the BUILD file

This PR is currently a `WIP`. It is opened to showcase the work done and to create a space for discussion on certain topics. The following list enumerates task that are currently being addressed and which are required for this PR to be completed before it is marked as `ready for review`.

Next steps:

- [x] Update documentation - describe the new architecture of the decoder and how to interact with the IP from the software (CSRs)
- [x] Add DSLX tests for the ZstdDecoder proc - create facilities for preparing test data and write DSLX tests
- [x] Improve error handling - write error codes to the `Status` CSR
- [x] Improve verilog tests - reorganize testbench to handle decoding of multiple ZSTD frames in one test case
- [x] Change the output format from stream-based to memory-based (write decoded frame to memory)
[...]

Issue #1897, opened 2025-01-30
Huge RAM consumption in DSLX interpreter


rw1nkler:
**Describe the bug**

Running a DSLX interpreter can consume a huge amount of RAM for larger designs like [ZSTD decoder](https://github.com/google/xls/pull/1857). 
Additionally, we noticed that the resources are not released between different test cases, and RAM consumption increases steadily over time.

**To Reproduce**

1. Checkout to the [zstd_compressed_block_dec](https://github.com/antmicro/xls/tree/zstd_compressed_block_dec) branch
2. Run any program for monitoring RAM usage (`top`, `htop`) 
3. Run `bazel run -- //xls/modules/zstd:zstd_dec_dslx_test --logtostderr`
4. Observe growing RAM consumption

**Expected behavior**

The interpreter should not consume that much RAM on larger designs.
Ideally, for a correct design that reads data from all its channel queues, it should be possible to run the DSLX interpreter simulation infinitely with a (more or less) constant RAM consumption.

I will try to provide more debug/profiling information and append to this issue.

rw1nkler:
It seems that the `BytecodeInterpreter::EvalCreateTuple` is responsible for huge allocation of memory
![Image](https://github.com/user-attachments/assets/7dcab7e3-7775-4b67-8d09-a5c43b901d2a)

allight:
Looks like the elements array is uselessly copied instead of moved.

allight:
Yeah making the vector and elements be moved seems to fix this: 

![Image](https://github.com/user-attachments/assets/204f7f6c-916f-4c58-bbd3-ce07ca0faed8)

rw1nkler:
Thank you @allight for reducing the number of re-allocations. However, the overall RAM consumption is similar to the previous values. Correct me if I'm wrong, but to my understanding, the `move` prevents the creation of copies when operating on the`elements` vector, but previously, the temporary vector was freed anyway and should not contribute to the overall RAM consumption.

I wonder why the memory is growing linearly, and why it's not freed between running different test cases.

allight:
Sorry I was on vacation so I didn't see this.

I think that this is just WAI (or at least working-as-implemented. At a minimum I don't see any obvious places we are leaking things.

The interpreter needs to make copies of values for many operations to avoid having to implement a full GC. Also as long as the values are relatively small the overhead for a gc would be much greater than the overhead of copying.

[...]

Issue #1042, opened 2023-06-21
DSLX: bug: bytecode interpreter is very slow and quickly runs out of memory when using RamModel in tests


rdob-ant:
To implement #995, DSLX procs need to use RAM blocks, each with size in range 8 - 128 KBytes.
When running bytecode interpreter-based tests that make use of RAM blocks (using `xls_dslx_test` Bazel rule with `compare: none`) I've noticed that the progression of the test is extremely slow - it literally executes 1 or 2 iterations of `next()` of my _test_proc_ per second. In addition to that, after running the test for ~180 seconds my PC runs out of RAM and the interpreter is killed by Linux kernel OOM. I've confirmed that the RAM consumption comes from `interpreter_main` executable of XLS, I have 32G RAM + 32G swap and the interpreter consumes around 55G at the moment when it is killed. 

@proppy @cdleary Do you perhaps have some idea why this may be happening and whether there is some workaround? Currently it's a blocker as it prevents us from testing code that uses RAMs.

I can reproduce the behavior using following minimalistic _test_proc_. It fills 64KByte RAM with zero bytes starting from address 0 and incrementing it by 1 till the full RAM is filled. It writes one byte each 10 iterations of `next()`.

```
import xls.examples.ram as ram

type RamWriteReq = ram::WriteReq;
type RamWriteResp = ram::WriteResp;
type RamReadReq = ram::ReadReq;
type RamReadResp = ram::ReadResp;

#[test_proc]
proc tst_ram {
    o_term: chan<bool> out;

    o_write_req: chan<ram::WriteReq<16, 8, 8>> out;
    i_write_resp: chan<ram::WriteResp> in;
    o_read_req: chan<ram::ReadReq<16, 8>> out;
    i_read_resp: chan<ram::ReadResp<8>> in;

    init {(u32:0, u16:0)}

    config(o_term: chan<bool> out) {
        let (o_write_req, i_write_req) = chan<ram::WriteReq<16, 8, 8>>;
        let (o_write_resp, i_write_resp) = chan<ram::WriteResp>;
        let (o_read_req, i_read_req) = chan<ram::ReadReq<16, 8>>;
        let (o_read_resp, i_read_resp) = chan<ram::ReadResp<8>>;

        spawn ram::RamModel<u32:8, {u32:1<<u32:16}, u32:1>
            (i_read_req, o_read_resp, i_write_req, o_write_resp);

        (
            o_term, 
            o_write_req, i_write_resp,
[...]

PR #2296, opened 2025-05-30
modules/zstd: Add support for decoding compressed blocks (DSLX part)

rw1nkler:
This PR adds the DSLX parts of #1857 in order not to depend on cocotb-related python packages supersedes #1857

This PR extends the ZstdDecoder with support for decoding compressed blocks.

The decoder is capable of decoding RAW and RLE literals as well as sequences with predefined FSE tables. A suite of DSLX tests comprising unit tests of all underlying procs and an integration test was prepared. The integration test, similarly as in #1654, first generates a random valid ZSTD frame with compressed blocks and expected decoded output. Test data is then converted to a DSLX file ([example](https://github.com/antmicro/xls/blob/zstd_compressed_block_dec_dslx_part/xls/modules/zstd/data/comp_frame.x)) that is imported by the [integration tests file](https://github.com/antmicro/xls/blob/zstd_compressed_block_dec_dslx_part/xls/modules/zstd/zstd_dec_test.x). At the beginning of the test, the default FSE decoding tables are filled with default distributions taken from [RFC 8878 section 3.1.1.3.2.2. Default Distributions ](https://datatracker.ietf.org/doc/html/rfc8878#name-default-distributions). Next, the encoded frame is loaded to the system memory and the decoder is configured through a set of CSRs to start the decoding process. The decoder starts the operation and writes the decoded frame back into the output buffer in the system memory. Once it finishes, it sends a pulse on the notify channel signaling the end of the decoding. The output of the decoder is compared against the decoding result from the [reference library](https://github.com/indygreg/python-zstandard).

The PR introduces among others:

* `SequenceDecoder` - responsible for decoding sequence sections of the compressed blocks
* `FseDecoder` - introduced as the core part of the `SequenceDecoder`
* `RefillingShiftBuffer` - used for storing and outputting in forward and backward fashion an arbitrary amount of bits required by the FSE decoder
* `LiteralsDecoder` - capable of decoding RAW, RLE and Huffman-coded literals
* `HuffmanDecoder` - used in decoding huffman-coded literals. Decoded Huffman trees are then used to decode one or four Huffman-coded streams.
* `CommandConstructor` - this proc is responsible for sending packets with decoded sequences and literals to the `SequenceExecutor` proc
* `RamMux` and `RamDemux` - procs used for handling requests/responses to multiple memory models. The procs interface with 3 separate memory buffers for FSE decoding tables.


richmckeever (approved):

xls/modules/zstd/BUILD
```diff
@@ -14,14 +14,14 @@
 
 # Build rules for XLS ZSTD codec implementation.
 
+# pytype binary, library
```
proppy: can you remove this since the python code is not included in this PR anymore?

proppy:
Looks like the following target blow up over the internal memory limit for test (~10GB):
```
//xls/modules/zstd:huffman_weights_dec_dslx_test
//xls/modules/zstd:sequence_dec_dslx_test
//xls/modules/zstd:zstd_dec_dslx_test
```
Is that expected? (possibly related to #1042?)

rw1nkler:
> Looks like the following target blow up over the internal memory limit for test (~10GB):
> 
> ```
> //xls/modules/zstd:huffman_weights_dec_dslx_test
> //xls/modules/zstd:sequence_dec_dslx_test
> //xls/modules/zstd:zstd_dec_dslx_test
> ```
> 
> Is that expected? (possibly related to #1042?)

@proppy yest the issue is caused by huge RAM consumption in DSLX interpreter in general, which was also reported in https://github.com/google/xls/issues/1897

proppy:
Is there a way to reduce the number of test cases so that it fits within 10G?

xls/modules/zstd/zstd_dec_test.x
```diff
@@ -0,0 +1,900 @@
...
+import xls.modules.zstd.zstd_dec;
+import xls.modules.zstd.comp_block_dec;
+import xls.modules.zstd.sequence_dec;
+import xls.modules.zstd.memory.mem_reader;
+import xls.modules.zstd.huffman_literals_dec;
+import xls.modules.zstd.parallel_rams;
+import xls.modules.zstd.literals_buffer;
+import xls.modules.zstd.fse_table_creator;
+import xls.modules.zstd.ram_mux;
+// import xls.modules.zstd.zstd_frame_testcases as comp_frame;
```
proppy, line 31: is that commented on purpose? is the idea here to allow different type of frame to be tested depending on what you uncomment? (if yes maybe those should be split in different tests?)

rw1nkler:
> Is there a way to reduce the number of test cases so that it fits within 10G?

@proppy I will try to reduce the number of test cases or move them to different test procs. 
The ability to filter tests could help with this separation. Please take a look at https://github.com/google/xls/pull/2312

proppy:
@rw1nkler as discussed, let's just `tag` the target as `local` for now w/ a TODO that point to #1897 and/or #1042.

rw1nkler:
I noticed that OpenROAD failed with a segmentation fault in the CI. We've actually seen quite a bit of instability with this tool internally. In many cases, simply rerunning the flow has resolved the issue.

proppy:
> I noticed that OpenROAD failed with a segmentation fault in the CI. We've actually seen quite a bit of instability with this tool internally. In many cases, simply rerunning the flow has resolved the issue.

@QuantamHD / @mikesinouye fyi.

proppy (approved):

QuantamHD:
We might need to bump openroad. There were some tsan issues resolved recently.

xls/modules/zstd/BUILD
```diff
@@ -2679,17 +2679,18 @@ benchmark_synth(
...
-    clock_period = "750",
-    core_padding_microns = 2,
-    min_pin_distance = "0.5",
-    placement_density = "0.30",
-    stop_after_step = "global_routing",
-    synthesized_rtl = ":literals_decoder_synth_asap7",
-    tags = ["manual"],
-    target_die_utilization_percentage = "10",
-)
+# TODO: Uncomment after resolving problems with too long P&R
```
proppy: shouldn't `manual` be enough to avoid triggering it?

xls/modules/zstd/BUILD
```diff
@@ -3016,17 +3017,18 @@ benchmark_synth(
...
-    clock_period = "750",
-    core_padding_microns = 2,
-    min_pin_distance = "0.5",
-    placement_density = "0.30",
-    stop_after_step = "global_routing",
-    synthesized_rtl = ":huffman_data_preprocessor_synth_asap7",
-    tags = ["manual"],
-    target_die_utilization_percentage = "10",
-)
+# TODO: Uncomment after resolving problems with too long P&R
```
proppy: shouldn't `manual` be enough to avoid triggering it?

xls/modules/zstd/BUILD
```diff
@@ -3093,17 +3095,18 @@ benchmark_synth(
...
-    core_padding_microns = 2,
-    min_pin_distance = "0.5",
-    placement_density = "0.30",
-    stop_after_step = "global_routing",
-    synthesized_rtl = ":huffman_decoder_synth_asap7",
-    tags = ["manual"],
-    target_die_utilization_percentage = "10",
-)
+# TODO: Uncomment after resolving problems with too long P&R
+# place_and_route(
```
proppy: shouldn't `manual` be enough to avoid triggering it?

xls/modules/zstd/BUILD
```diff
@@ -3349,17 +3352,18 @@ benchmark_synth(
...
-    clock_period = CLOCK_PERIOD_PS,
-    core_padding_microns = 2,
-    min_pin_distance = "0.5",
-    placement_density = "0.30",
-    stop_after_step = "global_routing",
-    synthesized_rtl = ":huffman_literals_dec_synth_asap7",
-    tags = ["manual"],
-    target_die_utilization_percentage = "10",
-)
+# TODO: Uncomment after resolving problems with too long P&R
```
proppy: shouldn't `manual` be enough to avoid triggering it?

mikesinouye:
@QuantamHD's current work is finishing the Qt bazel build hookup for upstream OpenROAD. Once that's finalized, we can clean rules_hdl's deps and bump the pointer to head for OR. There have been many stability improvements in the last 15 months so hopefully that addresses whatever issues you're observing. I think this will happen soonish, once we do let's revisit this to confirm it does.

xls/modules/zstd/BUILD
proppy: bump, this need to be removed as it makes internal check fail because it's now unused (no more python deps)

richmckeever (approved):

xls/modules/zstd/BUILD
rw1nkler: I removed those comments

xls/modules/zstd/BUILD
rw1nkler: The `manual` tag is used to exclude a target from the `//xls/modules/zstd:all` and `//xls/modules/zstd/...` targets.  We use it for all targets except `xls_dslx_library` so that the main CI can typecheck the zstd part without running other expensive checks.

Hoverwer, there is a dedicated CI for zstd. To run all `manual` targets, it explicitly lists them using `bazel query`, which searches for groups of targets with common suffixes.  I made these rules more strict to allow re-enabling some previously commented-out targets.  To skip them safely, I used the `_skip` suffix in the target names so the CI won't match them.

xls/modules/zstd/zstd_dec_test.x
rw1nkler: This would be a perfect solution, however currently the interpreted does not free memory between running separate test cases, which may be a problem for this test.

rw1nkler: I also tried using `ir-jit` evaluator for dslx tests, but it works just for 29/55 of them. The other are failing due to the following reasons:
* Incorrect token usage in the for loop which cannot be converted to ir (these tests should use `unroll_for!` instead)
* `IrConversionError: Could not find name for invocation: 'enumerate'`, indicating that the `enumerate` has no representation in IR?
* Just a SegmanetationFault on many tests

PR #2312, opened 2025-06-02
Expose `test_filter` option from interpreter_main in build rules

rw1nkler:
This commits exposes the `test_filter` option of the `interpreter_main` in the build rules:
https://github.com/google/xls/blob/f69d07cf2ded54a7dd41ad07eeaca0823b5e8d51/xls/dslx/interpreter_main.cc#L69

This enables creating DSLX test targets that run only a specific test case:
```
xls_dslx_test(
    name = "my_proc_dslx_test",
    library = ":my_proc_dslx",
    dslx_test_args = {
        "test_filter": "MyProcTest",
    },
)
```

FYI @proppy

### DIFF
--- a/xls/build_rules/xls_dslx_rules.bzl
+++ b/xls/build_rules/xls_dslx_rules.bzl
@@ -109,6 +109,7 @@ def _get_dslx_test_cmdline(ctx, src, all_srcs, append_cmd_line_args = True):
         "enable_warnings",
         "max_ticks",
         "format_preference",
+        "test_filter"
     )
 
     dslx_test_args = dict(_dslx_test_args)


mikex-oss (approved):

allight:
I'm not sure we really want to do this.

--test_filter is used directly by bazel and other test-run tools. This seems likely to conflict with it.

@meheff @cdleary WDYT?

cdleary:
@allight The intent for that flag was definitely for it to be used from the command line like we do for gunit tests with `bazel test` -- I don't think this addition has an analog in C++ targets. Since I don't think it has an analogy in other test rules for other languages I'd think we want very strong rationale for why DSLX is different.

allight:
Huh and it seems @grebe merged this already.

This does break the bazel --test_filter flag on targets with this option. It seems to work fine if the option is not present.

We might want to consider rolling it back for now...

allight:
Yeah I'm rolling this back.

@rw1nkler Would you mind just giving us some info on what problem you were trying to solve with this change. We might be able to find a better way to support it. Was it just to enable faster turn-around with refining a single test?

rw1nkler:
My goal was to add more test cases for the zstd module. However, I ran into issues with high RAM usage, which made that difficult. These problems were already reported in the following issues:

* https://github.com/google/xls/issues/1897
*  https://github.com/google/xls/issues/1042

On top of that, the interpreter doesn’t free memory between test cases, which makes things worse. To work around this, I added a filter so the long-running tests can be run separately. This helped avoid running out of memory.

PR #2329, opened 2025-06-04
[codegen] flag to avoid inserting "invariant assertions" via codegen passes

cdleary:
Sometimes passes want to add in assertions for properties they "believe they have proven to be true but would like additional validation for at simulation-time".

We put these assertions behind a flag because they can slow down assert-condition coverage closure.

### DIFF
--- a/xls/codegen/priority_select_reduction_pass_test.cc
+++ b/xls/codegen/priority_select_reduction_pass_test.cc
@@ -17,7 +17,6 @@
 #include <memory>
 #include <vector>
 
-#include "absl/status/statusor.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "xls/codegen/block_conversion.h"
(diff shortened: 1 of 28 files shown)


xls/codegen/priority_select_reduction_pass_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  BValue case0 = fb.Literal(xls::UBits(11, 32));
+  BValue case1 = fb.Literal(xls::UBits(22, 32));
+  BValue def = fb.Literal(xls::UBits(0, 32));
+  BValue prio = fb.PrioritySelect(selector, {case0, case1}, def);
+  XLS_RETURN_IF_ERROR(fb.BuildWithReturnValue(prio).status());
+  XLS_RETURN_IF_ERROR(package->SetTopByName("f"));
+  return package;
+}
+
+void RunPassAndCheck(bool enable_asserts, bool expect_assert) {
```
mikex-oss: Does it make sense to make this a parameterized test instead? So all the meaningful logic stays inside the test body?

Or factor out into smaller helpers that the tests call to keep arrange/act/assert separate?

https://abseil.io/resources/swe-book/html/ch12.html#example_onetwo-twozerodot_tests_should

mikex-oss (commented):
LGTM, just one minor test comment.

xls/codegen/priority_select_reduction_pass_test.cc
cdleary: Switched to parameterized test, PTAL.

mikex-oss: LGTM, thanks!

mikex-oss (approved):

cdleary:
Hold off on the merge if you can -- I see something we can do better here.

mikex-oss:
> Hold off on the merge if you can -- I see something we can do better here.

Is this latest push ready to be merged?

cdleary:
> > Hold off on the merge if you can -- I see something we can do better here.
> 
> Is this latest push ready to be merged?

Yes thanks! I realized that we could forget to propagate the field so I made a method to access the encapsulated CodegenOptions and made the codebase consistently initialize things "in one shot" instead of setting fields one at a time on a mutable default instance.

meheff (approved):

PR #2353, opened 2025-06-08
LSP: if stdlib path is not provided in full, also search dslx_path.

hzeller:
The search path, which is typically set via an environment variable for the language server, then also allows to find the stdlib path.

### DIFF
--- a/xls/dslx/lsp/dslx_ls.cc
+++ b/xls/dslx/lsp/dslx_ls.cc
@@ -128,14 +128,28 @@ void TextChangeHandler(const LspUri& file_uri,
   }
 }
 
-// Attempt to canonicalize path and return that if successful; else keep as-is.
+// Attempt to canonicalize "original_path" and return that if successful.
+// If it can not be resolved as absolute path or relative to CWD, attempt
+// to resolve relative to the elements in "search_path".
+// If all fails, return as-is (don't crash to not bring down the LS).
 static std::filesystem::path FailSafeCanonicalize(
+    const std::vector<std::filesystem::path>& search_path,
     const std::filesystem::path& original_path) {
   std::error_code ec;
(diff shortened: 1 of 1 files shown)


ericastor:
Seems sensible to me, but I'll want the frontend team to take a look. @richmckeever - who should make this call?

hzeller:
Context: the default path is just a relative path from project root [`xls/dslx/stdlib` ](https://github.com/google/xls/blob/main/xls/dslx/default_dslx_stdlib_path.inc#L19), but of course the `CWD` where the editor happens to launch the language server is rarely that at the root of the dslx project, so it would typically fail to find it.

The user who is aware of that can override that with the `--stdlib_path` when invoking the language server (knowing how to provide command line options depending on the editor) and fix the path that way.

This PR just adds another way of finding the path in a fallback situation if the above fails: if the user has set the [`DSLX_PATH` environment variable (or flag)](https://github.com/google/xls/blob/main/xls/dslx/lsp/dslx_ls.cc#L53-L56)  to find it relative to that search path.

xls/dslx/lsp/dslx_ls.cc
```diff
@@ -130,13 +130,24 @@ void TextChangeHandler(const LspUri& file_uri,
 
 // Attempt to canonicalize path and return that if successful; else keep as-is.
```
dplassgit: Please update comment for new parameter

dplassgit (approved):
Approval with comment.

xls/dslx/lsp/dslx_ls.cc
hzeller: Done.

hzeller:
crossed streams got auto-closing confused. Manually closing.

PR #2369, opened 2025-06-10
[dslx] Add an out-of-tree interpreter test to regression test the usecase

cdleary:
Ran `$ python3 -m yapf --style="{based_on_style: google, indent_width: 2}" -i xls/dslx/interpreter_test.py` to try to get style compliant

### DIFF
--- a/xls/dslx/interpreter_test.py
+++ b/xls/dslx/interpreter_test.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 """Tests for the DSLX interpreter.
 
 Note that generally we want DSLX test to be in `xls/dslx/tests`, but this can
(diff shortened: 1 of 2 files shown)


xls/dslx/interpreter_test.py
```diff
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
dplassgit: Revert removal of extra newline

xls/dslx/interpreter_test.py
```diff
@@ -137,9 +136,9 @@ def test_fail_incomplete_match(self):
     """)
     # warnings-as-errors off because we have a useless match expression just to
     # trigger the type error.
-    stderr = self._parse_and_test(
-        program, warnings_as_errors=False, want_error=True
-    )
+    stderr = self._parse_and_test(program,
```
dplassgit, line 159: Nit: should have formatting/refactorings in a separate PR/CL from new code, but, shrug, it's easy enough to eyeball that they're the same.

cdleary: There's not a good open source formatting solution that gives exactly the same result for python like there is with clang-format. If there is, LMK and I'll run it. Otherwise I have to just guess what will pass presubmits and the draconian linter.

xls/dslx/interpreter_test.py
cdleary: Done

I also remembered I had that pylintrc set up in pre-commit so I added this file to that checked set and fixed the things it rejected so should have a better chance of passing the internal linter.

cdleary:
Rebased this to head, is it ready to land @dplassgit ?

dplassgit (approved):

PR #2397, opened 2025-06-13
[modules] Add dual path floating point adder.

cdleary:


xls/modules/add_dual_path/README.md
allight, line 1: Our lint checks are getting mad that this has lines longer than 80 chars. Can you format this.

cdleary: Done, and added it to the pre-commit.

meheff (approved):

PR #2404, opened 2025-06-14
Make sure the Python interpreter can be found.

hzeller:
Some rules use Python from the system but with assumptions that don't work on systems with diferent paths such as NixOS or BSD. Make sure it can be found by using the `use_default_shell_env` option.


dplassgit:
Did this originate in a CL or do you want this to go through the regular OSS flow?

hzeller:
Right now I am just a regular external hobby-user, so PR it is.

Issue #2379, opened 2025-06-11
[dslx:tiv2] type mismatch: Any[element_count<Any>() + element_count<u1>()]


cdleary:
**To Reproduce**

```dslx
#![feature(type_inference_v2)]

import std;

// Attempt to distill the match-with-tuple pattern that triggers the TypeInferenceError
// There are three match arms, each returning a (u8, ?).  Two arms give an explicit u2,
// one arm constructs its value via `std::lsb(..) ++ u1:0`.
// In the original bug the type unifier cannot solve the width of that concatenation.

pub fn repro(x: u3) -> u2 {
    let (_dummy, upper): (u8, u2) = match x {
        u3:0b000 | u3:0b001 => (u8:0, u2:0b00),
        u3:0b010 | u3:0b011 => (u8:0, std::lsb(x) ++ u1:0),
        _ => (u8:0, x[0 +: u2]),
    };
    upper
} 
```

$XLSYNTH_TOOLS/dslx_interpreter_main ./minimal_repro3.x  --dslx_stdlib_path $XLSYNTH_TOOLS/xls/dslx/stdlib

richmckeever:
Fixing this requires a more real top-sort of type data flow in ConversionOrderVisitor, which I am starting now and will probably finish after vacation.

What happens is the type of arm 0 of the match becomes dependent on resolving the function being invoked in arm 1. A similar dep could be set up in an if/else or binop. ConversionOrderVisitor at head is only smart enough to do this in the right order if the top-level expr of arm 1 is the invocation, so a workaround would be to wrap that in a helper function.

PR #2424, opened 2025-06-17
[dslx:tiv2] Process match arms containing invocations first.

cdleary:
Fix for example in #2379


erinzmoore (approved):

ericastor:
For the record, I think this is an incomplete solution, which is (I presume) why we're not closing #2379.

PR #2476, opened 2025-06-25
[viz] Fix regex that looked for ret as return, would also match on ret at start of identifiers

cdleary:


xls/visualization/ir_viz/ir_to_proto.cc
```diff
@@ -400,8 +400,7 @@ absl::StatusOr<std::string> MarkUpIrText(Package* package) {
     //
     //   <span>bar</span>: bits[32] = op(<span>x</span>, <span>y</span>, ...)
     std::string node_name;
-    if (RE2::PartialMatch(line, R"(^\s*(?:ret)?\s*([_a-zA-Z0-9.]+)\s*:)",
-                          &node_name)) {
+    if (RE2::PartialMatch(line, R"(^\s*(?:ret\s+)?([_a-zA-Z0-9.]+)\s*:)", &node_name)) {
```
dplassgit: This line is too long, max is 80 chars. :/

.pre-commit-config.yaml
```diff
@@ -13,6 +13,19 @@ repos:
...
         exclude: ^xls/contrib/
         types: [text]
+      # Automatically format staged C/C++ changes with git-clang-format.
+      - id: git-clang-format
+        name: git clang-format
+        # "git clang-format" formats only the staged changes when run without
+        # commit range arguments and with --staged. We add --quiet so that it
+        # prints output only when changes are made and pre-commit will then
+        # fail allowing the user to `git add` the re-formatted files.
+        entry: bash -c 'git clang-format --quiet --staged main'
```
dplassgit, line 23: This is failing:

fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

dplassgit (changes requested):

cdleary:
Note on closing PR: this still should be done since the IR viz is busted on certain identifiers, but should redo this PR instead of reviving it at this point.

PR #2488, opened 2025-06-26
Mark inherited methods with `final`

hzeller:
The `final` or `override` keyword was missing for these; since that class is never derived from, make them `final` to offer devirtualization opportunities for the compiler and better readability to the human.


dplassgit (changes requested):
There were no `override`s added in this PR. Please update PR description.

hzeller:
Done. Also expand in the CL description.

PR #2490, opened 2025-06-26
Change forward declaration of a `struct` (was `class`)

hzeller:
Forward-declaring a `struct` type as `class` might or might not work.


hzeller:
Done.

PR #2493, opened 2025-06-26
Fix a few low-hanging clang-tidy warnings.

hzeller:
```
2x [modernize-use-equals-default]
2x [performance-inefficient-vector-operation]
2x [readability-inconsistent-declaration-parameter-name]
1x [modernize-redundant-void-arg]
1x [performance-for-range-copy]
```


xls/passes/resource_sharing_pass.cc
```diff
@@ -2443,8 +2443,7 @@ NaryFoldingAction::NaryFoldingAction(
   }
 }
 
-std::vector<std::pair<Node *, uint32_t>> NaryFoldingAction::GetFrom(
-    void) const {
```
dplassgit, line 2447: WHAT YEAR IS IT!?

Issue #453, opened 2021-06-29
Add array slice optimizations


meheff:
Currently we don't have any array slice optimizations. Here are a couple which might be useful:

1. ArraySlice of an Array operation where the ArraySlice has a literal start. In this case you should be able to replace the ArraySlice with an Array operation which gathers the statically-known element of the slice from the original Array operation's operands.
2. ArraySlice of an ArraySlice in some instances might be replaceable with a single ArraySlice.

A (potentially crazy) idea might be to decompose ArraySlice into a bunch of ArrayIndex's followed by an Array operation to gather them up, We already have lots of optimizations for ArrayIndex which would then automatically kick in. This would avoid duplicating a lot of complicated index logic. And it doesn't fundamentally affect the scaling of XLS for big arrays because we already have to emit array operations as element-by-element in Verilog.

PR #2539, opened 2025-07-05
Array slice optimizations.

psivaraj:


ericastor:
Sounds great! Unfortunately we can't accept anything without a CLA - are you willing to fill that out?

psivaraj-openai:
> Sounds great! Unfortunately we can't accept anything without a CLA - are you willing to fill that out?

Yep. I work at OpenAI, and was just added to [openai-auth-google-contributors](https://groups.google.com/d/forum/openai-auth-google-contributors), so I'm going to re-open the PR under my OpenAI alias (annoying quirk from when I first set things up here).

xls/passes/array_simplification_pass.cc
```diff
@@ -215,6 +215,66 @@ bool IsSmallArray(Node* node) {
...
 }
 
+// Simplify select operations over array operations into select over array
+// operations (e.g. array_index, array_slice) by creating the appropriate
+// array operation for each case and then creating a new select over
+// those results.
+//
+// Example transformation:
+//   select(p, array_index(A0, {idx}), array_index(A1, {idx}))
+//     => array_index(select(p, A0, A1), {idx})
```
ericastor: Is this example written backwards? It seems like the code is taking an array_op(select()) and turning it into a select(array_op()).

psivaraj:
> Yep. I work at OpenAI, and was just added to [openai-auth-google-contributors](https://groups.google.com/d/forum/openai-auth-google-contributors), so I'm going to re-open the PR under my OpenAI alias (annoying quirk from when I first set things up here).

NVM, just needed to sign it with the right e-mail address. Ready for review 🙏

xls/passes/array_simplification_pass.cc
```diff
@@ -1026,10 +1041,184 @@ absl::StatusOr<SimplifyResult> SimplifyArrayUpdate(
...
+    auto* inner_array = array_slice->array()->GetType()->AsArrayOrDie();
+    if (bits_ops::UEqual(start_bits.value(), 0) &&
+        array_slice->width() == inner_array->size()) {
+      XLS_RETURN_IF_ERROR(array_slice->ReplaceUsesWith(array_slice->array()));
+      return SimplifyResult::Changed({array_slice->array()});
+    }
+  }
+
+  // Replace a slice with a known start with the operands. This optimization
+  // is more restrictive than the identity case above beause it requires the
```
ericastor: spelling nit: "beause" -> "because"

xls/passes/array_simplification_pass.cc
```diff
@@ -1026,10 +1041,184 @@ absl::StatusOr<SimplifyResult> SimplifyArrayUpdate(
...
-// kArray op which gathers the elements written into the array. Returns a vector
-// of the optimized away update operations or nullopt is no optimization was
-// performed.
+// Try to simplify the given array slice operation.
+absl::StatusOr<SimplifyResult> SimplifyArraySlice(
+    ArraySlice* array_slice, const QueryEngine& query_engine,
+    int64_t opt_level) {
+  // Flatten nested array slices by summing start indices.
+  //   array_slice(array_slice(A, s1, w1), s2, w2)
+  //     → array_slice(A, s1 + s2, w2)
```
ericastor, line 1050: Might be an odd question, but have you seen evidence that this is net-beneficial? It's not obvious to me that an adder tree in the selector will always be cheaper. (Ideally a simple example with metrics would be awesome, but I'm open to discussing it more broadly.)

xls/passes/array_simplification_pass.cc
```diff
@@ -1026,10 +1041,184 @@ absl::StatusOr<SimplifyResult> SimplifyArrayUpdate(
...
+  //
+  // For array slice i, we assign max_start_i as:
+  //  - its exact constant if fully known, or
+  //  - the maximal unsigned value of the start node type.
+  // We apply the transform only when
+  //     Σ(max_start_i) + (array_slice.width − 1) < size(A)
+  // meaning every accessed element is provably in-bounds.
```
ericastor, line 1057: Why is this only valid/good if all accessed elements are in-bounds? It seems at first glance like it'd be just as valid as long as the start can't go past the end.

xls/passes/array_simplification_pass.cc
```diff
@@ -1026,10 +1041,184 @@ absl::StatusOr<SimplifyResult> SimplifyArrayUpdate(
...
+  // the type to be an array.
+  //
+  // Out-of-bounds access returns the last element.
+  //
+  // Examples:
+  //   array_slice(array(a, b, c, d), start=1, width=2) → array(b, c)
+  //   array_slice(array(a, b, c, d), start=2, width=3) → array(c, d, d)
+  if (array_slice->width() > 0 && array_slice->array()->Is<Array>() &&
+      start_bits.has_value() && start_bits->FitsInUint64()) {
+    uint64_t start = start_bits->ToUint64().value();
```
ericastor: nit: You might be able to use `UnsignedBitsToSaturatedInt64` here? (Found in `bits_ops.h`.) Since we don't support arrays with size that doesn't fit in an int64_t, saturating to the max value is often a good way to handle the past-the-end case.

xls/passes/array_simplification_pass.cc
```diff
@@ -215,6 +215,66 @@ bool IsSmallArray(Node* node) {
...
+// Simplify select operations over array operations into select over array
+// operations (e.g. array_index, array_slice) by creating the appropriate
+// array operation for each case and then creating a new select over
+// those results.
+//
+// Example transformation:
+//   select(p, array_index(A0, {idx}), array_index(A1, {idx}))
+//     => array_index(select(p, A0, A1), {idx})
+template <typename NewOpT>
+absl::StatusOr<SimplifyResult> SimplifySelectOfArrayOperation(
```
ericastor, line 227: This is a really clean refactor - thanks!

xls/passes/array_simplification_pass.cc
psivaraj: Good question. Unfortunately I don't have metrics off hand, but my thinking is that the final Verilog for `kArraySlice` implements the slice as `width` parallel `N` to 1 muxes, with each mux having a width equal to the array element type. Then there's also the cost the comparators for the selects. Flattening removes an entire layer of these muxes and replaces it with a small adder tree (width log₂N). 

If you prefer, I could instead make this more conservative and only apply the folding if `start` is known, in which case it's simply a constant fold and no actual adder tree is materialized. 

Let me know what you think.

xls/passes/array_simplification_pass.cc
psivaraj: I think flattening two slices into one keeps results identical only when every possible element the chain can touch is already inside the base array. Otherwise, the inner slice might pad with the last valid element that would need to be propagated through the flattening and concatenated to the final output.  

For example, with size(A) = 5:
- slice(A, 2, 2) → [A[2], A[3]]
- slice(slice(A, 2, 2), 3, 1) clamps and returns A[3],

but flattening to `slice(A, 5, 1)` clamps and returns A[4].

xls/passes/array_simplification_pass.cc
psivaraj: Fixed.

xls/passes/array_simplification_pass.cc
psivaraj: Whoops, you're right. Got myself tangled up. Fixed.

xls/passes/array_simplification_pass.cc
psivaraj: If we wanted to avoid the in-bounds check, then I think we’d need to explicitly track and propagate the last valid element from each intermediate slice.

xls/passes/array_simplification_pass.cc
psivaraj: Done. Let me know if that's sufficient.

xls/passes/array_simplification_pass.cc
ericastor: Not arguing there! I think that's probably a good enough argument for me, and we'll see if our benchmarks bear it out.

xls/passes/array_simplification_pass.cc
ericastor: I see - you're right, I'd missed the cascading effect. Thanks for clarifying!

ericastor:
Having checked internally, looks like we're not seeing any QoR impact on a variety of test scenarios... but this still seems like a good idea in general.

@psivaraj-openai, does this result surprise you on your end?

psivaraj:
Hey @ericastor , thanks for checking. 

I'll preface this by saying that I mainly opened this PR to get my feet wet with XLS and picked it up because of the “good-first-issue” tag on issue #453!

That said, I’m not too surprised about the QoR impact. The new rules only trigger when fairly specific patterns show up which I imagine are  rare in hand-written RTL and even in most DSLX (maybe they'd show up in something like a sliding window DSP kernel?). Also, when the rule does fire it's likely a micro-optimisation that most modern synthesizers already manage. 

Anyways, I figure just having an entry point to future array slice optimizations makes it worthwhile enough to include?

ericastor:
Absolutely agreed! Just wasn't sure what we were expecting here. Thanks for clarifying. I'll get a second LGTM & we'll get this merged!

allight:
FYI This has been merged internally but a slight bug with our anti-leak software means that the cl wasn't mirrored back. We'll get it fixed shortly but it might not close this pr and might mess up git-blame. Sorry.

Looks like copybara was able to handle it.

ericastor:
Afraid it looks like we have a bug here: https://github.com/google/xls/issues/2692

The issue appears to be incorrect handling of two slices in sequence where the second runs past the end of the first; e.g., `array_slice([0, 1], 0, 1) == [0]`, and `array_slice([0], 0, 2) == [0, 0]`, but this optimization instead turns `array_slice(array_slice([0, 1], 0, 1), 0, 2)` into `[0, 1]`.

psivaraj:
>Afraid it looks like we have a bug here: https://github.com/google/xls/issues/2692

Gah, thanks for catching it and for the patch.

PR #2550, opened 2025-07-08
Add a few C APIs

benquike:
-  `xls_type_get_kind`
- `xls_type_get_leaf_count`
- `xls_package_get_functions`


benquike:
Can someone take a look at this PR?

@ingomueller-net Can you take a pass first?

xls/public/c_api.cc
```diff
@@ -892,6 +916,30 @@ bool xls_package_get_function(struct xls_package* package,
...
+  CHECK(package != nullptr);
+  CHECK(error_out != nullptr);
+  CHECK(result_out != nullptr);
+  CHECK(count_out != nullptr);
+  xls::Package* xls_package = reinterpret_cast<xls::Package*>(package);
+  absl::Span<const std::unique_ptr<xls::Function>> functions = xls_package->functions();
+  *count_out = functions.size();
+  *result_out = new struct xls_function*[*count_out];
+
+  for (size_t i = 0; i < *count_out; ++i) {
```
grebe, line 941: In some other cases (e.g. `ToOwnedCStrings`) I believe we avoid allocating the empty array and set the output to nullptr. This seems like another situation where that would make sense, WDYT?

grebe (changes requested):

xls/public/c_api.cc
benquike: updated.

grebe (approved):

cdleary (approved):

PR #2070, opened 2025-04-28
[DSLX:stdlib] Add split_msbs, split_lsbs

dank-openai:

PR #2569, opened 2025-07-10
[DSLX; std] implement msbs & lsbs

dank-openai:
- document new functions in dslx_std.md
- fix: annotate literal with type (how did this work before?)
- move split_msbs & split_lsbs up, just after msbs
  - move split_lsbs before split_msbs, to match lsb & msb ordering

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -403,6 +403,66 @@ fn msbs_test() {
     assert_eq(msbs<u32:3>(u5:30), u3:7);
 }
 
+// Splits bits into (N most significant bits, the remaining least significant
+// bits).
+//
+// This function ensures that all bits of the argument are used.
+pub fn split_msbs<N: u32, X: u32, Z: u32 = {X - N}, FROM_START: s32 = {Z as s32}>
+    (x: bits[X]) -> (bits[N], bits[Z]) {
+    // Can't split more bits than exist
+    const_assert!(N <= X);
+
+    let msbs = x[FROM_START:];
+    let lsbs = x[0:FROM_START];
+    (msbs, lsbs)
+}
+
+#[test]
+fn test_split_msbs() {
+    assert_eq(split_msbs<u32:0>(u5:0b10101), (uN[0]:0, u5:0b10101));
+    assert_eq(split_msbs<u32:1>(u5:0b10101), (u1:0b1, u4:0b0101));
+    assert_eq(split_msbs<u32:2>(u5:0b10101), (u2:0b10, u3:0b101));
+    assert_eq(split_msbs<u32:3>(u5:0b10101), (u3:0b101, u2:0b01));
+    assert_eq(split_msbs<u32:4>(u5:0b10101), (u4:0b1010, u1:0b1));
+    assert_eq(split_msbs<u32:5>(u5:0b10101), (u5:0b10101, uN[0]:0));
+}
+
+#[quickcheck(exhaustive)]
+fn prop_split_msbs(n: uN[4], o: uN[3]) -> bool {
+    let (n2, o2) = split_msbs<4>(n ++ o);
+    n == n2 && o == o2
+}
+
+// Splits bits into (the remaining most significant bits, N least significant
+// bits).
+//
+// This function ensures that all bits of the argument are used.
+pub fn split_lsbs<N: u32, X: u32, Y: u32 = {X - N}, FROM_START: s32 = {N as s32}>
+    (x: bits[X]) -> (bits[Y], bits[N]) {
+    // Can't split more bits than exist
+    const_assert!(N <= X);
+
+    let msbs = x[FROM_START:];
+    let lsbs = x[0:FROM_START];
+    (msbs, lsbs)
+}
+
+#[test]
+fn test_split_lsbs() {
+    assert_eq(split_lsbs<u32:5>(u5:0b10101), (uN[0]:0, u5:0b10101));
+    assert_eq(split_lsbs<u32:4>(u5:0b10101), (u1:0b1, u4:0b0101));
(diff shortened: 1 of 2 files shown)


dank-openai:
I'm going to seed the inevitable naming discussion with [this.](https://github.com/google/xls/pull/2070#issuecomment-2862193141)

xls/dslx/stdlib/std.x
```diff
@@ -374,6 +374,95 @@ fn msb_test() {
...
+    assert_eq(msbs<u32:1>(u2:1), u1:0);
+    assert_eq(msbs<u32:2>(u3:5), u2:2);
+    assert_eq(msbs<u32:3>(u5:30), u3:7);
+}
+
+// Splits bits into (the remaining most significant bits, N least significant
+// bits).
+//
+// This function ensures that all bits of the argument are used.
+pub fn split_lsbs<N: u32, X: u32, Y: u32 = {X - N}, FROM_START: s32 = {N as s32}>
```
proppy: nit (as this is not really something introduced by this change): we should consider making `FROM_START` an const binding internal to the function since it's not really something meant for the user to override (nor visible in the signature)

proppy: > s32

couldnt that be `sN[std::clog2() + 1]`

xls/dslx/stdlib/std.x
```diff
@@ -374,6 +374,95 @@ fn msb_test() {
     assert_eq(u1:1, msb(s2:0b11));
 }
 
+
+// Returns the R LSbs of x.
+// `x` must be at least R bits wide.
+pub fn lsbs<R: u32, N: u32>(x: uN[N]) -> uN[R] {
```
proppy: q: should we alias `lsb` and `msb` to those? (does it get optimized to the same thing?)

proppy: noticed that `lsb` takes a parametric on the sign-ness, should we be consistent and do the same thing here?

dank-openai: Yes, good idea. Will do.

I think `msbs` should also be sign parametric.

xls/dslx/stdlib/std.x
dank-openai: done

xls/dslx/stdlib/std.x
dank-openai: done (for both msbs and lsbs)

dank-openai: Although for `lsbs` when argument is signed, should the result always be unsigned? I think "yes". WDYT?

dank-openai: I tried aliasing lsb to `lsbs<u32:1>(x)` but:

```const_assert! failure: `N >= R` constexpr environment: {N: u32:0, R: u32:1, S: u1:0}```

I was stumped by that, as I could not find anywhere that lsb was being called on a `u0`

dank-openai: I'll let you resolve, to determine if you're satisfied with lsbs and msbs now

proppy: /cc @richmckeever, fyi

proppy: > Although for `lsbs` when argument is signed, should the result always be unsigned? I think "yes". WDYT?

Theorically, only if the bit sign end up being discarded; but might be cumbersome to required all callers to also be parameterized on signed-ness: curious what others think?

Another option would be to have those functions takes `xN`s but returns `bits`s, wdyt?

dank-openai: That's what I have `lsbs` doing right now, returning `uN` (which is the same as `bits`, yes?):

`pub fn lsbs<R: u32, S: bool, N: u32>(x: xN[S][N]) -> uN[R] {`

proppy: Yes, I was thinking `msbs` could do the same thing, wdyt? (and `bits` might be a little more explicit that the return value should be interpreter as raw bits).

dank-openai: * I think msbs is different from lsbs, as the sign bit is always going to be returned. I.e. a negative signed argument will always return a negative signed result. A positive signed argument will always return a positive signed result. So I think there's a valid interpretation/usage of `msbs` as "truncate this signed value to fewer bits". Or more mathematically we can interpret `msbs` as "divide a signed integer by a power of two". For example, -32 : s8 is `0b1110_0000`. Take 5 msbs means discard 3 lsbs means divide by 2^3=8. Result is `0b1_1100` = -4. However, if you insist, I can make msbs always return bits.
* lsbs can return bits, as you ask. Changing...

proppy: > Or more mathematically we can interpret msbs as "divide a signed integer by a power of two". For example, -32 : s8 is 0b1110_0000

Makes sense, sorry for the lengthy back and forth.

ericastor: Apologies for stepping in, but I'm going to have to agree with the original take here. I think `msbs` (as a bit-level function) should probably return `bits[R]`, rather than a signed integer.

The fact that you can implement division of `x: s8` by 8 using `msbs<5>(x)` doesn't mean that the result should be inherently signed. I'd suggest that if you were doing that, you'd probably want to write `(x >> 3) as s5` instead; that way, it's easy to read off that it's division by 8 without knowing that `x` happens to be exactly 8 bits wide. Even better, you could write `(x / 8) as s5` and trust the optimizer to do the right thing. 😄 

If this is a strong disagreement, I'm happy to either discuss it further, or call in another opinion we all agree on for the tie-break.

dank-openai: > I think msbs (as a bit-level function) should probably return bits[R]

done

ericastor:
LGTM, but I'll defer to @richmckeever

dank-openai:
Polite nudge: can we shepherd this along?

richmckeever (approved):

PR #2591, opened 2025-07-14
Relax CHECK in inline_bitmap from equality to greater-or-equal.

benquike:
As the title says


benquike:
@grebe can you take a look at this ?

hongted (approved):

PR #2594, opened 2025-07-14
[c_api] Expose query for whether function requires implicit token calling convention.

cdleary:
Note: based on top of #2589


hongted (approved):

cdleary:
Looks like this landed in the referenced commit it just didn't close the PR, closing manually.

PR #2636, opened 2025-07-18
Fix a few clang-tidy warnings.

hzeller:
Also, disable the `-Wnullability-completeness`, as it is mostly noise currently.


ericastor (approved):
Thanks!

PR #2641, opened 2025-07-19
Add more C-APIs

benquike:


xls/public/BUILD
```diff
@@ -262,6 +262,7 @@ cc_library(
         ":runtime_build_actions",
         "//xls/common:init_xls",
         "//xls/common/file:filesystem",
+	"//xls/data_structures:inline_bitmap",
```
benquike: spacing issue

xls/public/c_api.cc
```diff
@@ -421,6 +422,58 @@ struct xls_value* xls_value_make_false() {
   return reinterpret_cast<xls_value*>(value);
 }
 
+// Populate the xls_value from a byte array in place. The bit_count must be
+// greater than or equal to the flat bit count of the value.
+// The byte array is expected to be padded with 0s to the end of the value.
+// Returns false if the bit_count is less than the flat bit count of the value
+// or if the value cannot be populated from the bytes, in which case the error
+// message is written to error_out. error_out must be freed by the caller with
+// xls_c_str_free.
```
cdleary, line 431: I think this comment should be in the header not in the impl

xls/public/c_api.cc
```diff
@@ -421,6 +422,58 @@ struct xls_value* xls_value_make_false() {
...
+bool xls_value_populate_from_bytes(struct xls_value* value,
+                                   const uint8_t* bytes, size_t byte_count,
+                                   size_t bit_count, char** error_out) {
```
cdleary, line 434: the real question for interfaces like these is how the final partial byte at the end of the array is expected to be populated and how it corresponds to the populated BitVector's value, e.g. you could imagine MSbs of the last byte or LSbs of the last byte or padding requirements or etc -- we should make sure it's well documented in the C API documentation since it's committing to it as a public interface

xls/public/c_api.cc
```diff
@@ -421,6 +422,58 @@ struct xls_value* xls_value_make_false() {
...
+  CHECK(value != nullptr);
+  CHECK(bytes != nullptr);
+  CHECK(error_out != nullptr);
+  auto* cpp_value = reinterpret_cast<xls::Value*>(value);
```
cdleary, line 438: I think mutating XLS values in place is not great -- I think my overall question is can we prefer a constructor API or you really need/want this populate API for maximum speed?

xls/public/c_api.cc
```diff
@@ -1036,6 +1089,172 @@ int64_t xls_type_get_leaf_count(struct xls_type* type) {
...
+bool xls_type_is_token(struct xls_type* type) {
+  CHECK(type != nullptr);
+  xls::Type* xls_type = reinterpret_cast<xls::Type*>(type);
+  return xls_type->IsToken();
+}
+
+// convert a type to a bits type, return false if the type is not a bits
+// type and set the error_out to the error message, error_out should be freed
+// with xls_c_str_free; otherwise return true and set the result_out to the
+// bits type, result_out *does not* need to be freed.
```
cdleary, line 1126: comments should all be in the .h

you may also want to note the retval is expected that it can (and currently will) alias the input `type` pointer

xls/public/c_api.cc
```diff
@@ -1036,6 +1089,172 @@ int64_t xls_type_get_leaf_count(struct xls_type* type) {
...
+    return false;
+  }
+  *result_out = reinterpret_cast<struct xls_bits_type*>(bits_type.value());
+  return true;
+}
+
+// convert a type to a tuple type, return false if the type is not a tuple
+// type and set the error_out to the error message, error_out should be freed
+// with xls_c_str_free; otherwise return true and set the result_out to the
+// tuple type, result_out *does not* need to be freed.
```
cdleary, line 1145: I think a block comment will be better than duplicating on each function since then the likelihood they get out of sync goes up, the one benefit of not having a javadoc like system

xls/public/c_api.cc
benquike: I agree that maintaining a constructor style API to retain immutability in values is a good idea.

However at the same time, it might also be very inconvenient, especially considering that value can be nested.  e.g., mutating a single bit in a tree of values requires reconstructing the whole tree.
The idea to introduce this API is to make it easier in such cases.

I think the byte ordering issue you mentioned here is not unique to this specific API, it is also present in APIs like  xls_bits_make_ubits. I agree that we should document it better.

proppy:
converted this back to a draft, feel free to comment when it's ready to review again.

Issue #1549, opened 2024-08-19
Adding Formal Verification for IR [as extension to #1344 and discussion #1537]


dassbrothers:
### What's hard to do? (limit 100 words)

Making the generated Z3 code independent and provable standalone. There are three things I would like to address:

1. The assertion node is unhandled while converting an IR to Z3. 
2. Function parameters also do not show up in the Z3 code while converting an IR to Z3
3. Can the "**#test**" part of the DSLX code be used to generate an equivalent testbench in System Verilog?
 
![image](https://github.com/user-attachments/assets/a3174631-b8d2-43ee-b043-59eba5643628)

### Current best alternative workaround (limit 100 words)

The second point of adding the function parameters to the Z3 can be easy, as the function signature is already present in the Package class but is not public and cannot be accessible from the "smtlib_emitter_main". For point 1, we need to modify the parser to handle the assertion node and convert it to equivalent Z3 code. For each line of the IR, we have an equivalent Z3 code that needs to be addressed. However, I am still not sure how to address point 3. 

### Your view of the "best case XLS enhancement" (limit 100 words)

We can independently verify the IR without performing logical equivalence checking with the above advancements. Secondly, if the "#test" block can be converted to System Verilog in the same manner, then we can quickly verify the generated Verilog code via simulating.

proppy:
@dassbrothers was asking for a few pointers:

# how to  get the `Function` params name:
- https://github.com/google/xls/blob/aa165490129cdb522c13d2369f9b8aabb32e2cbe/xls/ir/function_base.h#L84
- https://github.com/google/xls/blob/aa165490129cdb522c13d2369f9b8aabb32e2cbe/xls/ir/nodes.h#L724

# how params are translated to z3
- https://github.com/google/xls/blob/aa165490129cdb522c13d2369f9b8aabb32e2cbe/xls/solvers/z3_ir_translator.cc#L644-L669

# assert IR semantic
- https://google.github.io/xls/ir_semantics/#assert
- https://github.com/google/xls/blob/aa165490129cdb522c13d2369f9b8aabb32e2cbe/xls/ir/nodes.h#L172

allight:
So it seems like it would be somewhat helpful to get some more details about what exactly you think a good result would be.

Reading this and your other bugs I assume you'd like a few different things

### Z3 Improvements - input declarations

Currently if you have some function like 

```
fn muladd(a: u8, b: u8, c: u8) -> u8 { (a * b) + c }
```

you'll get a z3 program like

[...]

PR #2642, opened 2025-07-21
[xls][z3] Add support for assert in Z3 IrTranslator

benquike:
assert nodes are addressed by not doing any translation

### DIFF
--- a/xls/solvers/z3_ir_translator.cc
+++ b/xls/solvers/z3_ir_translator.cc
@@ -374,6 +374,11 @@ absl::Status IrTranslator::HandleBinary(OpT* op, FnT f) {
 absl::Status IrTranslator::HandleAdd(BinOp* add) {
   return HandleBinary(add, Z3_mk_bvadd);
 }
+
+absl::Status IrTranslator::HandleAssert(Assert* assert) {
+  return absl::OkStatus();
+}
+
 absl::Status IrTranslator::HandleSub(BinOp* sub) {
   return HandleBinary(sub, Z3_mk_bvsub);
 }
--- a/xls/solvers/z3_ir_translator.h
+++ b/xls/solvers/z3_ir_translator.h
@@ -149,6 +149,7 @@ class IrTranslator : public DfsVisitorWithDefault {
   absl::Status DefaultHandler(Node* node) override;
   absl::Status HandleAdd(BinOp* add) override;
   absl::Status HandleAfterAll(AfterAll* after_all) override;
+  absl::Status HandleAssert(Assert* assert) override;
   absl::Status HandleMinDelay(MinDelay* min_delay) override;
   absl::Status HandleArray(Array* array) override;
   absl::Status HandleArrayIndex(ArrayIndex* array_index) override;


allight:
I'm not sure I'd call ignoring asserts 'support' TBH.

Is there something not working with the 'ignore_unsupported' constructor option which does the same thing?

benquike:
> Is there something not working with the 'ignore_unsupported' constructor option which does the same thing?

'ignore_unsupported' would ignore other nodes not supported yet. 
for `assert`, I think it is OK to translate it to no z3 nodes, as dslx would do under production builds

I have updated the description

allight:
I mean the thing is naively I'd expect that assert would translate into a z3 assert of some type to prove statements about the whole design. Simply ignoring it seems wrong. This is admittedly a big change to handle however.

The other unsupported ops are things like registers, cover, etc which don't seem relevant to cases like this.

For other cases like this we handle this by simply removing the asserts from the ir before analysis (eg: http://google3/third_party/xls/dev_tools/check_ir_equivalence_main.cc?l=198).

Maybe we should take a step back; What is the use case for this? How are you invoking the z3 translator and what are you trying to use it for?

benquike:
> What is the use case for this? How are you invoking the z3 translator and what are you trying to use it for?

I am trying to get the z3 expression of functions for other analysis,  using `xls_function_to_z3_smtlib`. However, this C API raises a runtime error complaining that assert not supported. 

> For other cases like this we handle this by simply removing the asserts from the ir before analysis (eg: 

http://google3/third_party/xls/dev_tools/check_ir_equivalence_main.cc?l=198).
maybe I can enable this pass instead.

benquike:
> I mean the thing is naively I'd expect that assert would translate into a z3 assert of some type to prove statements about the whole design.

that's a nice idea, do you have any idea? maybe I can follow that direction as well.

allight:
> > For other cases like this we handle this by simply removing the asserts from the ir before analysis (eg:
> > http://google3/third_party/xls/dev_tools/check_ir_equivalence_main.cc?l=198).
>
> maybe I can enable this pass instead.

I'd be much happier with that change if you're willing.

You'd probably want to just give those passes their own xls_pass targets and add them to the pass-registry in xls/passes/BUILD. Maybe give them short-names with a 'for-testing' suffix or something.

> > I mean the thing is naively I'd expect that assert would translate into a z3 assert of some type to prove statements about the whole design.
>
> that's a nice idea, do you have any idea? maybe I can follow that direction as well.

This is more a 'rewrite of how we do z3' sort of project unfortunately. There are also questions of how exactly you want the assert to be emitted which probably needs some clarification/configuration.

allight:
I actually just did the cleanup to move the passes to their own targets myself: http://cl/785516740 (https://github.com/google/xls/pull/2648)

benquike:
smt-lib2 supports assert directly, Why not translate assert statement here to it?

Or more broadly speaking, why are we targeting the translation to z3, instead of smtlib2?

dassbrothers:
@benquike The key idea was to add the support for assert in the z3_ir_translator as well. This is because if we think of verifying designs from a hardware point of view, then ideally, we would like to verify the design against a set of properties. So, these properties would be written in some assert in the code. This would help to verify the design in a standalone way without equivalence checking and also not by simulation rather formally. 

Moreover, I was trying to convert it into smtlib2 format only and not z3. The change you made in the two files (z3_ir_translator.h and z3_ir_translator.cc) would simply help to skip the assert node and not throw any error. However, it will not convert the same into smtlib2. 

Please look into this issue that I had filed earlier
https://github.com/google/xls/issues/1549

benquike:
> @benquike The key idea was to add the support for assert in the z3_ir_translator as well. This is because if we think of verifying designs from a hardware point of view, then ideally, we would like to verify the design against a set of properties. So, these properties would be written in some assert in the code. This would help to verify the design in a standalone way without equivalence checking and also not by simulation rather formally.
> 

agreed.

> Moreover, I was trying to convert it into smtlib2 format only and not z3. The change you made in the two files (z3_ir_translator.h and z3_ir_translator.cc) would simply help to skip the assert node and not throw any error. However, it will not convert the same into smtlib2.

Good, that was what I was asking about. I agree that this change is not proper, as shown in previous discussions.

mikex-oss:
>  I agree that this change is not proper, as shown in previous discussions.

If I'm not misinterpreting this comment and surrounding discussion, I'm going to move this PR back to draft status.

PR #2646, opened 2025-07-21
[c_api] More query APIs for vast

cdleary:


proppy (approved):

xls/public/c_api_vast.cc
```diff
@@ -616,4 +626,105 @@ struct xls_vast_statement* xls_vast_statement_block_add_nonblocking_assignment(
...
+}
+
+xls_vast_data_type* xls_vast_def_get_data_type(struct xls_vast_def* def) {
+  auto* cpp_def = reinterpret_cast<xls::verilog::Def*>(def);
+  return reinterpret_cast<xls_vast_data_type*>(cpp_def->data_type());
+}
+
+bool xls_vast_data_type_width_as_int64(struct xls_vast_data_type* type,
+                                       int64_t* out_width, char** error_out) {
+  CHECK(out_width != nullptr);
```
proppy, line 687: nit: are we sure we want to `CHECK` here vs error'ing out? (it seems to be the first time we do this in this file)

xls/public/c_api_vast.cc
```diff
@@ -616,4 +626,105 @@ struct xls_vast_statement* xls_vast_statement_block_add_nonblocking_assignment(
   return reinterpret_cast<xls_vast_statement*>(cpp_assignment);
 }
 
+struct xls_vast_module_port** xls_vast_verilog_module_get_ports(
+    struct xls_vast_verilog_module* m, size_t* out_count) {
+  auto* cpp_module = reinterpret_cast<xls::verilog::Module*>(m);
+  absl::Span<const xls::verilog::ModulePort> ports = cpp_module->ports();
+  *out_count = ports.size();
```
proppy, line 634: nit: should we check/error out that `out_count` is not null.

xls/public/c_api_vast.cc
```diff
@@ -616,4 +626,105 @@ struct xls_vast_statement* xls_vast_statement_block_add_nonblocking_assignment(
...
+  if (ports.empty()) {
+    return nullptr;
+  }
+  // Allocate an array owned by the caller (to be freed by
+  // xls_vast_verilog_module_free_ports).
+  auto** result = new xls_vast_module_port*[ports.size()];
```
proppy, line 640: nit: should we check/error out that `result` is not null.

xls/public/c_api_vast.cc
cdleary: Yeah I haven't been consistent about it (you can see some in the other c_api.cc file), but it's just slightly better message wise vs a segfault that usually results from the UB, so don't need to be too religious about it I think, I could eliminate them all in this file if you think the consistency is preferable.

xls/public/c_api_vast.cc
cdleary: `new` is infallible so don't need to do that

xls/public/c_api_vast.cc
cdleary: Done.

xls/public/c_api_vast.cc
proppy: I was mostly curious if that was the first case of us doing that :)

If that's already being done in other file, it's consistent to do it here as well!

proppy (approved):

proppy:
@cdleary can you rebase?

cdleary:
> @cdleary can you rebase?

Done, but note github didn't seem to think a rebase was needed (maybe was needed internally for some reason tho)

Issue #2335, opened 2025-06-05
[enhancement] Add DSLX and/or IR source annotations to generated Verilog


rw1nkler:
### What's hard to do? (limit 100 words)

It’s hard to match generated Verilog code back to the original DSLX source. DSLX code is turned into an intermediate representation and optimized before Verilog is made. Because of this, the Verilog often looks very different from the DSLX it came from. This makes it tough to debug code when the generated design does not match the expected behavior.

### Current best alternative workaround (limit 100 words)

Right now, people have to reason about the parts of the DSLX code that matches with the generated Verilog by tracking similarities between operations and generated variable names.

### Your view of the "best case XLS enhancement" (limit 100 words)

XLS could keep track of where each piece of DSLX code came from and add this info as comments in the generated Verilog, like in the following example:
```verilog
// From my_file.x: myProc, line 42, col 5
assign nor_359453 = ~(____state_0 | wr_resp_valid_reg | rd_resp_valid_inv);
// From my_file.x: myProc, line 43, col 5
assign and_359454 = and_359447 & wr_req_valid;
// From my_file.x: myProc, line 44, col 5
assign and_359455 = and_359447 & ~wr_req_valid & rd_req_valid;

```

proppy:
support for mapping verilog to DSLX in separate file could also be interesting to explore; maybe we could reuse https://tc39.es/source-map/ to allow us to leverage existing tooling like https://marketplace.visualstudio.com/items?itemName=colder.source-map-visualization

hzeller:
For Verilog, wouldn't it be neat to have the backtick-line annotation (LRM section 22.12) ?
That way even tools can easily ingest the content.

However, that directive only allows to express line numbers not line + column (or even line+column range). So maybe switchable by flag to choose between backtick-line and human readable comments.

hzeller:
Also CC @hongted as he was exploring this field before.

PR #2652, opened 2025-07-22
vast: use PreEmit to produce comments with source location

wsipak:
This is a draft with effort towards generating SystemVerilog with comments that contain information about DSLX source location, described previously in https://github.com/google/xls/issues/2335.
The aim of this draft PR is to present the idea and make space for suggestions and discussions on the proper way to implement the solution.

The definition of `ModuleMember` in `vast.h` allows adding `Comment`  objects which are generated as expected. However, adding new `Comment` with source location doesn't allow necessary per-line granularity. Also there are objects like ports emitted in `Module::Emit` which are stored in port-only collection (type `ModulePort`) which doesn't allow adding per-port comments.

Considering that the generation of comments is a generic task in terms of various types inheriting from `VastNode`, the idea here is to implement `PreEmit` that will be generic, implemented in `VastNode`. `PreEmit` should produce a comment string only if source information is available for the object. This won't cover the `ModulePort` struct which is not a `VastNode`, so objects of this and similar types need to be handled separately. This can be solved by extending `ModulePort` with additional field of `SourceInfo` or `std::string` to store the information, or by generating the comments along with ports in `Module::Emit`.


tmichalak:
@richmckeever could you take a look and comment on our initial approach

xls/codegen/vast/vast.h
```diff
@@ -2248,6 +2250,13 @@ struct ModulePort {
   std::string ToString() const;
 
   ModulePortDirection direction;
+  // TODO since ModulePort is not a VastNode,
+  // add a field for source location info
+  // use a string for a comment,
+  // or SourceInfo that will be used to format a comment
+  std::string _docstring;
+  const std::string& docstring() const { return _docstring; }
```
richmckeever: I would just make the string a public member like the direction rather than having a setter, and maybe  name it `doc_string`.

xls/codegen/vast/vast.cc
```diff
@@ -48,6 +48,13 @@
 namespace xls {
 namespace verilog {
 
+std::string VastNode::PreEmit(LineInfo* line_info) {
+    // TODO if (!loc().Empty)
```
richmckeever: Let's have some draft of the actual string in here before checkin. Also have a plan for how we avoid emitting the source info when not flagged on. I would suggest either checking in the flagging mechanism as a no-op first, or combining with this PR.

richmckeever (changes requested):

wsipak:
The current changes allow generating the selected type of annotations for IR objects that have the `pos=` argument present. However, when considering Verilog generation from DSLX, IR objects seldom have the `pos=` argument present, so I'm going to work on the completeness of this information in a follow-up PR.

tmichalak:
@richmckeever please take a look

xls/codegen/block_generator.cc
```diff
@@ -1121,6 +1122,25 @@ absl::StatusOr<std::vector<Block*>> GatherInstantiatedBlocks(Block* top) {
...
+  switch (strategy) {
+    case xls::verilog::CodegenOptions::SourceAnnotationStrategy::kNone:
+      return AnnotationType::kNone;
+    case xls::verilog::CodegenOptions::SourceAnnotationStrategy::kComment:
+      return AnnotationType::kComment;
+    case xls::verilog::CodegenOptions::SourceAnnotationStrategy::kLineDirective:
+      if (file_type == FileType::kSystemVerilog) {
+        return AnnotationType::kLineDirective;
+      } else {
+        // TODO warn as `line directive is available in SystemVerilog only
```
richmckeever: I would go ahead and warn here using VLOG. Also dropping the "else" is the usual style after an "if" that returns.

xls/codegen/vast/vast.cc
```diff
@@ -48,6 +48,45 @@
 namespace xls {
 namespace verilog {
 
+std::string getFilename(Fileno fileno, VerilogFile* file) {
```
richmckeever: Capitalize the G in GetFilename.

xls/tools/codegen_flags.cc
```diff
@@ -210,6 +210,9 @@ ABSL_FLAG(std::string, reference_residual_data_path, "",
 ABSL_FLAG(std::string, block_metrics_path, "",
           "The filename to write the metrics, including the bill of "
           "materials, for the generated Verilog file");
+ABSL_FLAG(std::string, source_annotation_strategy, "None",
```
richmckeever: I think it's more common for flags like this to have the value (like "none" or "comment") in all lowercase.

xls/codegen/block_generator.cc
tmichalak: Fixed

xls/codegen/vast/vast.cc
tmichalak: Fixed

xls/tools/codegen_flags.cc
tmichalak: Fixed

proppy (approved):

Issue #2553, opened 2025-07-08
RAM rewriting causes artificial responses on RAM write channel


rw1nkler:
**Describe the bug**

For [RamDemux](https://github.com/google/xls/blob/9c1e05d43bd453bbb8afbe71a35a13b1ebc50bc8/xls/modules/zstd/ram_demux.x#L37) proc that we created to allow access to two RAMs from one pro, applying RAM rewriting causes incorrect behavior of the system in Verilog. Specifically, after RAM rewriting is applied, the proc begins to emit artificial responses on the write response channel even though no write requests were ever sent. The problem only shows up after applying the RAM rewrite pass, as the simulation without this step behaves as expected.

We suspect the problem is related to RAM rewriting being applied to non-blocking receive statements. This case might not be handled correctly by the pass. We believe this because when we wrapped the outputs of the RamDemux with additional [RamPassthrough](https://github.com/antmicro/xls/blob/a5e063f39581be8633a2d92447353b52035355f5/xls/modules/zstd/ram_passthrough.x#L159) procs (which use blocking I/O), and then applied RAM rewriting, everything started working correctly.

I created a simple simulation showing the problem, that is available at:
https://github.com/antmicro/xls/tree/ram_rewrite_issue

**To Reproduce**
1. Check out the `ram_rewriting_issue` branch from our [XLS fork](https://github.com/antmicro/xls/tree/ram_rewrite_issue)
2. Execute the following command:
   ```bash
   bazel run -c opt //xls/modules/zstd:ram_demux_cocotb_test
   ```
3. The script will run three test cases. Their results will be printed in the terminal log. One of them should fail.

4. Each test generates an FST waveform file that can be analyzed using a waveform viewer like GTKWave. Look for lines in the logs that start with:
   ```
   FST info: dumpfile /path/to/output.fst
   ```

   These lines show where each waveform file is saved.

**Description of the test cases**

The test writes data first to the first RAM, then switches to write to the second RAM. After that, it switches again to read from the first RAM and then from the second RAM, checking that the values are correct.

- The first test shows the proc’s correct behavior without RAM rewriting. In this case, no extra traffic appears on the write response channel.

<img width="1892" height="401" alt="Image" src="https://github.com/user-attachments/assets/cd34d368-9318-4e6b-8413-f3bfc5fe26ac" />

- The second test shows the `RamDemux` proc **after** RAM rewriting. Here, artificial responses are generated even before any write requests are sent.

[...]

PR #2653, opened 2025-07-22
Improve Huffman decoding in ZSTD decoder

koblonczek:
This PR primarily includes fixes to the Huffman coding algorithm used for decoding compressed literals in ZSTD frames. The changes address issues related to the ordering of operations involving token handling and code computation.

It also includes a fix for the RAM rewriting pass, which previously failed when attempting to rewrite one of the RAM interfaces. The issue was resolved by introducing a passthrough-like process that aggregates all channels into a single proc, allowing the pass to complete successfully.

Additionally, this PR adds a proc name to the `trace_fmt!` statements so that they are visible both in the DSLX tests and in the Verilog simulation.

These fixes enabled us to run the Verilog tests internally with correct results. The corresponding test updates will be submitted in a separate pull request, as they require a reorganization of dependencies.


proppy:
> It also includes a fix for the RAM rewriting pass,

Is that a workaround for https://github.com/google/xls/issues/2553 ?

xls/modules/zstd/BUILD
```diff
@@ -1428,12 +1428,6 @@ xls_dslx_test(
     ],
 )
 
-xls_dslx_library(
```
proppy, line 1431: is that replaced w/ ram demux?

proppy:
> This PR primarily includes fixes to the Huffman coding algorithm

Can you also document the log change in the PR description?

xls/modules/zstd/BUILD
rw1nkler: with `RamPassthrough` ([code](https://github.com/google/xls/pull/2653/files/0d1fc31ec459b803af5cb041934830a2a9753bdf#diff-cb74ede7463d11191318780b775a62a1139076d590ea087065aaefe167a36c4fR1399))

rw1nkler:
> > This PR primarily includes fixes to the Huffman coding algorithm
> 
> Can you also document the log change in the PR description?

done

rw1nkler:
> > It also includes a fix for the RAM rewriting pass,
> 
> Is that a workaround for #2553 ?

This is a different problem with RAM rewriting, but we resolved it using the same solution.

PR #2691, opened 2025-07-25
Fix MODULE.bazel.lock to only contain relevant references.

hzeller:


hzeller:
In https://github.com/google/xls/commit/f0bf71b98bd32619d43feeaeb04b4078ac5a99a3 , this file accidentally got 10x as big, containing all kinds of dependencies not relevant for the build (I suspect `--lockfile_mode=update` was used ? The default way to update the MODULE.bazel.lock is to just build and then check in the updated lock file).

Anyway, this fixes it to be a less daunting blob and more the size it was before.

proppy (approved):

PR #2704, opened 2025-07-27
[c_api] Add C API to access imported module type info

lsrcz:


xls/public/c_api_dslx.h
```diff
@@ -76,6 +76,9 @@ struct xls_dslx_function;
 struct xls_dslx_quickcheck;
 struct xls_dslx_function;
 struct xls_dslx_param;
+struct xls_dslx_expr;
+struct xls_dslx_module_member;
+struct xls_dslx_type_dim;
```
cdleary, line 81: I think this opaque definition is unused?

lsrcz: I think there are usages? https://github.com/xlsynth/xlsynth/blob/abcba4930d8237876e7afcf54eddecdcda327619/xls/public/c_api_dslx.h#L343-L345

Maybe not directly used in this PR though

cdleary: I see, it's backfilling something that should have been there that was getting filled in implicitly sounds like. SGTM.

PR #2716, opened 2025-07-29
Improvements to ZSTD Decoder: 

laudominik:
These changes fix:

- off by one error in huffman weights decoding end condition that caused temporary buffers not to be zeroed at the last bytes
- endianness in reading multi-stream stream sizes
- end condition in huffman decoding so that only the last packet of the last stream is marked as "last" when passing data to the literals buffer


tmichalak:
@proppy

mgielda:
Hi @proppy could you help us get this merged? Those are necessary changes so it would let us minimize our delta to mainline.

rw1nkler:
Superseded by https://github.com/google/xls/pull/3034

tmichalak:
@proppy @dplassgit since #3034 got merged this one can be closed.

PR #2717, opened 2025-07-29
ZSTD decoder: cocotb utils and decoder testcases, verilog helper modules

laudominik:
Added functionality:

- multiple testcases covering the decoding of frames with "compressed" blocks of different type
- cocotb testing utilities for common operations on XLS channels
- collection of PnR benchmarks
- ZSTD frame generator
- tests of axi ram writer/reader procs as well as memreader & memwriter


tmichalak:
@proppy

mgielda:
It would be good to understand whether there are any major blockers to merging this - like Python deps etc. This will need the Google-internal CI to run, which I think needs our approval @proppy

rw1nkler:
Superseded by #3035

PR #2755, opened 2025-07-31
Use std::string in OptOptions

benquike:
using std::string_view is prone to memory bugs.
Since using std::string does not incur much performance penalty, swiching to std::string.


benquike:
can someone take a look at this PR?

allight (approved):

PR #2811, opened 2025-08-08
Use `inline constexpr` for toplevel constants in headers.

hzeller:
This way, they won't be reported as unused values warnings.


dplassgit:
I've never worked with this code before, reassigning to XLS build gardener.

ericastor:
I'm not the build gardener this week - but sure, I'll take a look! 😁

PR #2822, opened 2025-08-12
Remove variables that are assigned to, but never read.

hzeller:


xls/passes/optimization_pass_pipeline_test.cc
```diff
@@ -299,7 +299,7 @@ TEST_F(OptimizationPipelineTest, ProcScopedChannels) {
 
   // Create a top proc which instantiates two leaf procs and sends an input
   // value through the chain, accumulates it and then sends to the output.
-  Proc* top;
+  Proc* top = nullptr;
```
ericastor: I don't see any value in having this `top` variable. Can we remove it while we're here?

hzeller: Done.

PR #2823, opened 2025-08-12
Use `final` where possible unless `override` is needed.

hzeller:
Readability: The `final` keyword indicates to the reader of the code that an implementation is a leaf, improving readability; it is always possible to make it `override` later.

Performance: This also can result in performance improvements as the compiler can better see devirtualization opportunities.

This changes many places in leaf implementations to be annotated `final` instead of `override`. Maybe it also helps future code to adopt this good practice.


ericastor:
This is a very large change on which many software engineers are very opinionated in both directions, and amounts to a change in our style guidelines. We're not going to accept this at this time; if you'd like to open such a discussion, since you're at Google, please take it up internally.

EDIT: For anyone *not* at Google, if you'd like to suggest a style change, we're going to adopt a policy on this matter; style changes *must* originate in Issues or Discussions, not as Pull Requests.

PR #2826, opened 2025-08-12
Update version of rules_hdl

allight:
This new version of rules hdl has more sources listed for the ncurses dep which should make build breakages due to failed downloads less common.


ericastor:
Any idea what's up with the pre-commit hook failure?

allight:
existing failure I think.

allight:
Ok should be fixed.

PR #2857, opened 2025-08-14
[dslx:ir_convert] Fix overly aggressive renaming that propagated to parameters

### LINKED ISSUE #2849: Module port name does not match DSLX function parameter name
**Describe the bug**
IR function parameter (and subsequent Verilog module port) does not match the DSLX function parameter name if there is a reassignment of the parameter.

**To Reproduce**
Input:
```
pub fn my_fun(baz: u32) -> u32 {
    let foo = baz;
    foo
}
```

Generated IR:
```
$ ir_convert/ir_converter_main ~/tmp/f.x
package f

file_number 0 "/home/meheff/tmp/f.x"

fn __f__my_fun(foo: bits[32] id=1) -> bits[32] {
  ret foo: bits[32] = param(name=foo, id=1)
}
```

Note the parameter name is `foo` rather than `baz`.

cdleary:
Fixes #2849


ericastor:
Looks sensible to me, but I'll double-check it with someone from the frontend team!

PR #2912, opened 2025-08-20
Add support for generating compile_commands.json.

meheff:
Use hedronvision/bazel-compile-commands-extractor to generate the
compile_commands.json file. This file is used by clangd for autocomplete
and other IDE featues. This approach is more robust than the existing
dev_tools/make-compilation-db.sh approach and results with many fewer
error/warning false positives being flagged in the UI (unused includes,
undefined symbols, etc).

As configured the json file is large (70MB) but not enormous and does not
seem to pose problems for clangd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,9 +176,27 @@ localhost instances of [bazel-remote](https://github.com/buchgr/bazel-remote/).
 
 ### Getting Clangd completions
 
-A `compile_flags.txt` file compatible with clangd and similar tools can be
-created by running `xls/dev_tools/make-compilation-db.sh`. Follow directions for
-your editor to install clangd code completion.
+There are two ways of getting clangd completions and related features.
+
+A `compile_flags.txt` file compatible with clangd and similar
+tools can be created by running `xls/dev_tools/make-compilation-db.sh`.
+This approach is faster but does not account for differences in the way
+individual targets are built.
+
+Alternatively,
+[hedronvision/bazel-compile-commands-extractor](https://github.com/hedronvision/bazel-compile-commands-extractor)
+can be used to generate a `compile_commands.json` file that
+clangd can consume. This approach is slower to setup but the
+compiler flags are tailored for each target. The
+`compile_commands.json` file can be built by running:
+
+```bash
+bazel build -c opt //xls/... -k
+bazel run //:refresh_compile_commands
+```
+
+See the comments in the top-level
+[BUILD](https://github.com/google/xls/blob/main/BUILD) file for more details.
 
 ## Stack Diagram and Project Layout
 
--- a/BUILD
+++ b/BUILD
@@ -28,6 +28,7 @@
 # more tutorial information.
 
 load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+load("@hedron_compile_commands//:refresh_compile_commands.bzl", "refresh_compile_commands")
 load("@rules_license//rules:license.bzl", "license")
 
 package(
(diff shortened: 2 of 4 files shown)


meheff:
Here's a screen grab of the existing approach. Note the errantly flagged unused includes and undefined symbols:

image

And after:

image

ericastor:
@hzeller, I know you looked into this; was there a reason this didn't look promising?

... hopefully not, because this sounds good.

allight:
IIRC when I looked into this a few months ago the objection was mostly that its really slow and at least at that time the existing solution still worked well enough.

meheff:
Out of the box the JSON file was something like 420MB. As configured it is like 70MB. It's limited to just xls/... targets and has exclude_external_sources = True. One or both of those really reduced the size.

hzeller:
The hedronvision was not very useful last time I checked (lots and lots of missing includes, in particular seen with `clang-tidy`), but maybe it is better now. The `clang-tidy` still works well with the compilation DB generated by `xls/dev_tools/make-compilation-db.sh`

It looks like `clangd` actually is broken and does not honor relative paths anymore which seems to have broken the current editing experience. Just confirmed that this is indeed the case.

What `bant` emits, is e.g. `-Ibazel-out/../../../external/abseil-cpp~` which is always correct, independent of the current compilation mode in use (as bazel likes to create new symbolic links depend on if k8-opt or k8-dbg and stuff). If it was absolute, you'd have to re-create the compilation DB every time you change the build-mode.

Alas, It looks like current clangd can't deal with paths that contain ../ anymore or at least does not follow them correctly (so `-I/usr/local/google/home/hzeller/.cache/bazel/_bazel_hzeller/7c80d2a2a5bc94665318b231f0182730/external/abseil-cpp~
` for instance then finds the absl headers).

I'll add a workaround in `bant` to emit absolute paths, and try to find when clangd stopped working correctly (it still does work with `clang-tidy`, so there might be something special they put in there); but if in the meantime hedronvision works, please use that.

hzeller:
I have a fix for the broken `clangd` handling added to bant and so strongly suggest to instead just update bant to 0.2.2 instead (which is currently [in review on bcr](https://github.com/bazelbuild/bazel-central-registry/pull/5613)). The hedronvision compilation DB has a lot of issues.

meheff:
OOC what are the issues with the hedronvision compilation DB?

hzeller:
I was attempting to use it as the compilation db input for `clang-tidy` (for my recreational weekend cleanups), but it did not add all the includes needed, so that resulted in a lot of `clang-diagnostic-errors` (and once a file gets a diagnostic error, all bets are off with other issues it reports on that file).

Then it was extremely slow to build.
Also it was modifying something in my git client and otherwise modified the client and added files that I was not happy with.

I tried it several times on XLS and Verible and it was an exhausting experience (not only to the battery of my laptop :). And `bant` otoh works like a breeze at least for simple projecs like xls or verible. And apparently it now helped to find a bug in `clangd` :).

Since outside Google you use `bant` on a daily basis to auto-fill your dependencies, build_clean and print bazel targets etc. anyway, it is a good choice to also use for the compilation DB. If something does not work in the compilation DB, it then also helps improve the tool.

mikex-oss:
Is there a need for us to be opinionated here? Can we let folks choose whichever they prefer?

ericastor:
I think we can provide both options going forward. @meheff, would you mind editing the documentation change in this PR accordingly?

README.md
```diff
@@ -176,9 +176,18 @@ localhost instances of [bazel-remote](https://github.com/buchgr/bazel-remote/).
 
 ### Getting Clangd completions
 
-A `compile_flags.txt` file compatible with clangd and similar tools can be
-created by running `xls/dev_tools/make-compilation-db.sh`. Follow directions for
-your editor to install clangd code completion.
+The XLS project uses
```
ericastor: Let's provide both options, rather than saying which is more robust; if you don't mind, please leave the original documented option here as well, especially now that we have a fix in a separate PR.

meheff:
Hey Eric! (and Henner!) Yeah, that's a good idea to keep both of them around. Changed the documention to include both approaches.

ericastor:
"This branch has conflicts that must be resolved" - will happily review once that's fixed.

(And yep, hi Mark!)

mikex-oss:
@meheff for future reference, the documentation updates here need to go in docs_src to get migrated properly.

#2934 to add this to the developer documentation.

PR #2926, opened 2025-08-21
[DSLX:TS] Fix for TypeMissingError for bare zero! and all_ones! FunctionRefs.

cdleary:
Make bare parametric builtin macro references `zero!<T>`/`all_ones!<T>` first-class function-typed values (`() -> T`), so returning them produces a clear return-type mismatch instead of a TypeMissing error (wired in TIv1 and TIv2). Centralized the "AST-node builtins with explicit parametrics" names in a small helper and used that in the type system. Added regression tests for the parser sample and parameterized type-inference tests that validate both v1 and v2; for v1 we assert the full positional error string.

### DIFF
--- a/xls/dslx/type_system/typecheck_module_test.cc
+++ b/xls/dslx/type_system/typecheck_module_test.cc
@@ -125,16 +125,9 @@ TEST_P(TypecheckBothVersionsTest, ZeroMacroFunctionRefIsNotValue) {
       "Return type of function body for 'a' did not match the annotated return "
       "type.\nType mismatch:\n   () -> E\nvs E";
   absl::Status st = Typecheck(text).status();
-  if (GetParam() == TypeInferenceVersion::kVersion1) {
-    FileTable ft;
-    ASSERT_THAT(GetPositionalErrorData(st, "XlsTypeError", ft),
-                absl_testing::IsOk());
-    EXPECT_EQ(GetPositionalErrorData(st, "XlsTypeError", ft)->message,
-              std::string(kV1Msg));
-  } else {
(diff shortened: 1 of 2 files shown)


xls/dslx/type_system/typecheck_module_test.cc
```diff
@@ -117,6 +117,44 @@ TEST_P(TypecheckBothVersionsTest, IndexZeroSizedArray) {
...
 
+TEST_P(TypecheckBothVersionsTest, ZeroMacroFunctionRefIsNotValue) {
+  // Bare parametric macro reference should have function type `() -> E` and
+  // not be directly usable as a value of type `E`.
+  std::string_view text = R"(enum E:u2{} fn a()->E{ zero!<E> })";
+  constexpr const char* kV1Msg =
+      "Return type of function body for 'a' did not match the annotated return "
+      "type.\nType mismatch:\n   () -> E\nvs E";
+  absl::Status st = Typecheck(text).status();
+  if (GetParam() == TypeInferenceVersion::kVersion1) {
```
mikex-oss: Can these abstract away the v1/v2 diffferences in a test helper, like from xls/dslx/type_system/typecheck_test_utils.h?

> When tests do need their own logic, such logic should often be moved out of the test bodies and into utilities and helper functions

https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

cdleary: Okay I think I added the kind of matcher you were hoping for here, PTAL.

mikex-oss: Thanks!

PR #2949, opened 2025-08-26
Create and save residual metadata from codegen

meheff:
This residual data currently just includes the emission order of the nodes into verilog though it will likely expand. It can be consumed in later codegen processes to reuse the emission order and minimize the differences in the generated Verilog. For example if you dump the residual metadata from IR sample A and use it for codegen'ing IR sample A' (small diff from A) then the resulting verilog from A' will have a very similar emission order assuming a correspondence of node ids between A and A'.

### DIFF
--- a/xls/codegen/block_generator.cc
+++ b/xls/codegen/block_generator.cc
@@ -277,8 +277,8 @@ absl::StatusOr<std::vector<Stage>> SplitBlockIntoStages(
   return stages;
 }
 
-// Returns a sequence of node IDs which is the emission order from the
-// from the residual data, or an empty vector if none exists.
+// Returns a sequence of node IDs which is the emission order from the residual
+// data, or an empty vector if none exists.
 std::vector<int64_t> NodeIdOrderFromResidualData(
     std::string_view block_name, const CodegenResidualData& container) {
   std::vector<int64_t> ids;
--- a/xls/codegen/block_generator.cc
+++ b/xls/codegen/block_generator.cc
@@ -38,6 +38,7 @@
 #include "absl/strings/str_format.h"
 #include "absl/types/span.h"
 #include "xls/codegen/codegen_options.h"
+#include "xls/codegen/codegen_residual_data.pb.h"
 #include "xls/codegen/conversion_utils.h"
 #include "xls/codegen/expression_flattening.h"
 #include "xls/codegen/module_builder.h"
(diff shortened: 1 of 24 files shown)


xls/codegen/block_generator.cc
```diff
@@ -276,14 +277,32 @@ absl::StatusOr<std::vector<Stage>> SplitBlockIntoStages(
   return stages;
 }
 
+// Returns a sequence of node IDs which is the emission order from the
+// from the residual data, or an empty vector if none exists.
```
ericastor: nit: duplicated "from the"

xls/tools/codegen_flags.cc
```diff
@@ -394,6 +403,16 @@ static absl::StatusOr<bool> SetOptionsFromFlags(CodegenFlagsProto& proto) {
     any_flags_set = true;
     CHECK(proto.has_package_interface());
   }
+
+  // If a reference residual data path is provided, parse it and attach to
+  // the flags proto so downstream codegen can consume without file I/O.
+  if (FLAGS_reference_residual_data_path.IsSpecifiedOnCommandLine() &&
+      !absl::GetFlag(FLAGS_reference_residual_data_path).empty()) {
+    verilog::CodegenResidualData reference;
```
ericastor, line 412: I think this should also set `any_flags_set = true;`?

ericastor (commented):
Couple of minor things before I approve internally.

xls/tools/codegen_main_test.py
```diff
@@ -160,6 +161,12 @@ def _compare_to_golden(self, got: str) -> None:
       want: str = runfiles.get_contents_as_text(path)
       self.assertMultiLineEqual(got, want)
 
+  def _find_line_number(self, text: str, substring: str) -> int:
```
ericastor: This seems to be unused?

xls/codegen/block_generator.cc
meheff: Fixed.

xls/tools/codegen_flags.cc
meheff: Fixed.

xls/tools/codegen_main_test.py
meheff: Removed.

meheff:
Thanks! Resolved comments.

PR #2973, opened 2025-08-29
Add expression inlining decisions to codegen residual data.

meheff:
This follows the initial change which added codegen residual data which recorded the order in which nodes are emitted in the Verilog. The intent is for this data to be reused in a later codegen invocation to minimized the difference in the generated Verilog. This change adds the expression inlining decisions for each node.


ericastor:
Pinging @mikex-oss, who had some thoughts on cleaner ways to write the test files, IIUC.

xls/tools/block_to_verilog_main_test.py
```diff
@@ -66,14 +93,17 @@
...
-    b.block_name = 'my_function'
-    for name, node_id in node_pairs:
+    b.block_name = block_name
+    for spec in node_specs:
       n = b.nodes.add()
-      n.node_name = name
-      n.node_id = node_id
+      n.node_name = spec.node_name
+      n.node_id = spec.node_id
+      n.emitted_inline = spec.emitted_inline
```
mikex-oss: Just a nit that this is considered "the hard way" in Google Python best practices and that it's more concise/cleaner to build these with kwargs passed to `add`, e.g.

`b = ref.blocks.add(block_name=block_name)`

External reference: https://protobuf.dev/reference/python/python-generated/#:~:text=%23%20add()%20also%20forwards,%3D13)

meheff: TIL. Thanks Mike!

xls/tools/block_to_verilog_main_test.py
```diff
@@ -66,14 +93,15 @@
...
+                          node_specs: list[NodeMetadata], filename: str):
+    """Writes residual data for a block from a list of NodeMetadata."""
     ref = codegen_residual_data_pb2.CodegenResidualData()
     b = ref.blocks.add()
-    b.block_name = 'my_function'
-    for name, node_id in node_pairs:
-      n = b.nodes.add()
-      n.node_name = name
-      n.node_id = node_id
+    b.block_name = block_name
```
mikex-oss: Nit: you can also inline this one into the add above.

xls/tools/block_to_verilog_main_test.py
```diff
@@ -281,6 +315,69 @@ def test_pipeline_generator_rejects_reference_residual(self):
     self.assertNotEqual(proc.returncode, 0)
     self.assertIn('not supported when generating pipelines', proc.stderr)
 
+  def test_inline_residual_data_affects_inlining(self):
```
mikex-oss: Mind splitting this up so that each commented block ("# Inline everything", etc.) is a separate test? 

https://abseil.io/resources/swe-book/html/ch12.html#:~:text=%C2%A0There%E2%80%99s%20a%20better%20way%3A%20rather%20than%20writing%20a%20test%20for%20each%20method%2C%20write%20a%20test%20for%20each%20behavior.4

You can factor out `run_with_inline_specs` into a private module function and move `ir_file` to `setUp` if you want to avoid creating the same IR multiple times.

xls/tools/block_to_verilog_main_test.py
meheff: Done!

xls/tools/block_to_verilog_main_test.py
meheff: Good idea, that's cleaner.

meheff:
Anything else you need from me on this? Copybara is failing, but I don't think that is anything I have visibility into.

xls/tools/block_to_verilog_main_test.py
```diff
@@ -13,10 +13,12 @@
 # limitations under the License.
 
 import subprocess
+from dataclasses import dataclass
```
dplassgit: This is failing lint. The suggestion is to replace with `import dataclasses` and then refer to dataclass as `dataclasses.dataclass`

meheff: Done. Also enabled pylint for all python files on the precommit. Unfortunately, pylint generates a number of false positives even though it is using Googles pylintrc, so I set it to warn only by printing errors and returning error code 0.

dplassgit (approved):

PR #2975, opened 2025-08-30
[bugfix] Ensure nodes aren't cloned multiple times

lsrcz:
`AstCloner` previously attempted to prevent double-cloning by checking the `old_to_new_` mapping inside `VisitChildren`. However, this missed the case where `ReplaceOrVisit` was invoked directly on a node. In such cases, the same node could be cloned multiple times, breaking semantics.

For example, with the following code, the old logic would incorrectly produce distinct type entities for the argument and return type in the clone:

```dslx
type MyU32 = u32;
fn foo(x: MyU32) -> MyU32 {
  x
}
```

This PR fixes the issue by moving the `old_to_new_` check into `ReplaceOrVisit` and ensures nodes are only cloned once regardless of how they are visited.

This change revealed another bug: `HandleTestFunction` and `HandleTestProc` were not using `CastIfNotVerbatim` and would cause failure in the formatter test suite. This pull request make it consistent with existing implementations like `HandleFunction` and fixed the test failure.

### DIFF
--- a/xls/dslx/frontend/ast_cloner_test.cc
+++ b/xls/dslx/frontend/ast_cloner_test.cc
@@ -1655,19 +1655,19 @@ fn foo(x: MyU32) -> MyU32 {
   XLS_ASSERT_OK_AND_ASSIGN(AstNode * clone, CloneAst(foo));
   Function* cloned_foo = down_cast<Function*>(clone);
 
-  // Param type should reference the existing module-level alias, not a cloned
-  // one.
+  // Param type and the return type should reference the same alias.
   auto* param_trta = down_cast<TypeRefTypeAnnotation*>(
       cloned_foo->params()[0]->type_annotation());
   TypeRef* param_tr = param_trta->type_ref();
   ASSERT_TRUE(std::holds_alternative<TypeAlias*>(param_tr->type_definition()));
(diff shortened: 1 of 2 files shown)


cdleary (approved):

xls/dslx/frontend/ast_cloner_test.cc
```diff
@@ -1634,6 +1634,127 @@ fn bar() -> u32{
...
+  x
+}
+)";
+
+  FileTable file_table;
+  XLS_ASSERT_OK_AND_ASSIGN(auto module, ParseModule(kProgram, "fake_path.x",
+                                                    "the_module", file_table));
+  XLS_ASSERT_OK_AND_ASSIGN(Function * foo,
+                           module->GetMemberOrError<Function>("foo"));
+  XLS_ASSERT_OK_AND_ASSIGN(TypeAlias * alias,
```
richmckeever: This unused var needs to be deleted or used in order for this to be merged.

lsrcz: Thanks. Just fixed it.

proppy (approved):

proppy:
This was merged in https://github.com/google/xls/commit/f66695109c31f269311fe3a57d3494108103391c

PR #3055, opened 2025-09-12
Fix IR conversion bug in for loops.

meheff:
The implicit activation bit was improperly generated inside for loops in some cases. This resulted in spurious asserts or other side-effecting operations firing improperly. This manifested as a divergence in assert behavior in a test between DSLX interpreter and IR interpreter .

### DIFF
--- a/xls/dslx/ir_convert/function_converter.cc
+++ b/xls/dslx/ir_convert/function_converter.cc
@@ -1650,8 +1650,10 @@ absl::StatusOr<BValue> FunctionConverter::HandleForBody(
 
   XLS_ASSIGN_OR_RETURN(BValue init, Use(node->init()));
   if (implicit_token_data_.has_value()) {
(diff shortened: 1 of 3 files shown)


ericastor:
This is a great find! We may need to try to figure out why the fuzzing never found this...

meheff:
It was a tough one to suss out as the failing test was large. Most of the time was spent building infrastructure for dumping call traces in both interpreters (last bit is https://github.com/google/xls/pull/3057) which lead me to believe it had something to do with the implicit activation. Then I handed the AI (gpt5 in cursor) the call trace, user source code, and my vague notion of what might be going wrong and it one shotted the fix which was impressive.

allight:
Fuzzer didn't find it because it doesn't generate asserts: 
https://github.com/google/xls/blob/2861b55fa7c54938331f559e6b3cdab0cf25b883/xls/fuzzer/fuzz_coverage_test.py#L235C6-L240C28

PR #3060, opened 2025-09-13
Clarify logic around choosing the quickcheck evaluator.

### LINKED ISSUE #2970: Quickchecks don't execute unless `--compare=jit`
**Describe the bug**
Quickchecks don't execute unless `--compare=jit`.

Why would I want to slow down quickcheck execution by running it twice (interpreted and jit), and using the slow interpreter? What if I trust that JIT is correct and I just want to execute my quickcheck 4 billion times, as quickly as possible? 

**To Reproduce**

1. Make a file with a quickcheck in it, name it foo.x
2. `dslx_interpreter_main-arm64 --evaluator ir-jit foo.x`, observe: `[ SKIPPING QUICKCHECKS  ] (JIT is disabled)`. Also skips quickchecks: `--compare=none --evaluator ir-jit`
4. `dslx_interpreter_main-arm64 --compare=jit foo.x`, observe `RUN QUICKCHECK`
5. Note that if you supply both flags, that's rejected: `--compare flag is only supported on --evaluator=dslx-interpreter`

**Expected behavior**
I would expect that executing quickchecks as quickly as possible (i.e. with evaluator=jit and compare disabled) to be encouraged and accepted. I don't want to compare the interpreter against the JIT, that's just massively slowing things down.

meheff:
Previously, the quickcheck evaluator piggy-backed on the comparison runner which meant if you ran the interpreter with --evaluator=jit --comparison=none you'd get a confusing message about not running the quickcheck tests because the jit is disabled. Now a separate options field holds the quickcheck runner. It uses the comparison one if comparison is enabled, otherwise uses the --evaluator one. Also, allow the ir-interpreter to be used for quickchecks but only if a special flag is passed in (--run_quickcheck_when_interpreting).

Fixes #2970

### DIFF
--- a/xls/dslx/interpreter_main.cc
+++ b/xls/dslx/interpreter_main.cc
@@ -33,6 +33,7 @@
 #include "absl/strings/str_split.h"
 #include "absl/time/time.h"
 #include "absl/types/span.h"
+#include "re2/re2.h"
 #include "xls/common/exit_status.h"
 #include "xls/common/file/filesystem.h"
 #include "xls/common/init_xls.h"
@@ -47,7 +48,6 @@
 #include "xls/dslx/virtualizable_file_system.h"
 #include "xls/dslx/warning_kind.h"
 #include "xls/ir/format_preference.h"
-#include "re2/re2.h"
 
 // LINT.IfChange
 ABSL_FLAG(std::string, dslx_path, "",
@@ -95,6 +95,10 @@ ABSL_FLAG(std::string, evaluator, "dslx-interpreter",
           "the XLS-IR JIT. ir-interpreter' is the XLS-IR interpreter.");
 ABSL_FLAG(std::optional<bool>, type_inference_v2, std::nullopt,
           "Whether to use type system v2 when type checking the input.");
+ABSL_FLAG(
+    bool, run_quickcheck_when_interpreting, false,
+    "Whether to run quickchecks when using the IR interpreter. By default, "
+    "this flag is off because the IR interpreter is too slow.");
 // LINT.ThenChange(//xls/build_rules/xls_dslx_rules.bzl)
 
 namespace xls::dslx {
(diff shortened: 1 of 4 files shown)


xls/dslx/interpreter_main.cc
```diff
@@ -169,16 +173,31 @@ absl::StatusOr<TestResult> RealMain(
...
+      quickcheck_runner = std::make_unique<RunComparator>(CompareMode::kJit);
+    } else if (evaluator == EvaluatorType::kIrInterpreter) {
+      if (absl::GetFlag(FLAGS_run_quickcheck_when_interpreting)) {
+        quickcheck_runner =
+            std::make_unique<RunComparator>(CompareMode::kInterpreter);
+      }
+    } else {
+      // Quickcheck is never run with the DSLX interpreter.
+    }
+  } else if (compare_flag == CompareFlag::kJit) {
```
proppy: should we print or error out if `evaluator` is different?

meheff: Good idea. The thing to check is that the evaluator is the DSLX interpreter in this case. The comparison mode checks the result of the DSLX interpreter against the specified evaluator.

xls/dslx/interpreter_main.cc
```diff
@@ -169,16 +173,39 @@ absl::StatusOr<TestResult> RealMain(
   XLS_ASSIGN_OR_RETURN(std::string module_name, PathToName(entry_module_path));
 
   std::unique_ptr<AbstractRunComparator> run_comparator;
-  switch (compare_flag) {
-    case CompareFlag::kNone:
-      break;
-    case CompareFlag::kJit:
+  // When comparison is on, use the same mode (JIT/interpreter) as the
+  // comparator. Otherwise, use the same mode as the evaluator flag. Using the
+  // IR interpreter requires --run_quickcheck_when_interpreting.
```
proppy, line 178: q: is that because it's not a common usecase that we want to not run the quickcheck by default when the evaluator is the interpreter?

one alternative could be to have a `--run_quickcheck` flag w/ different default depending on the evaluator being set (but maybe that would be more confusing?)

PR #3061, opened 2025-09-13
Add ReplaceInvocationsInModule

lsrcz:
This PR introduces `ReplaceInvocationsInModule`. It operates on `TypecheckedModule` since type information is required to perform the transformation.

`ReplaceInvocationsInModule` rewrites all invocations that matches a given callee function and parameter list to different callees and parameter lists. This is particularly useful for tools I am building for transforming a DSLX file into a fully monomorphized form.

The result is a new `TypecheckedModule`.


xls/public/c_api_dslx.h
```diff
@@ -82,6 +82,20 @@ struct xls_dslx_type_dim;
 struct xls_dslx_parametric_env;
 struct xls_dslx_interp_value;
 struct xls_bits;
+// Rule for rewriting invocations in bulk API.
```
cdleary, line 86: Nit: newline above this

xls/dslx/replace_invocations_test.cc
```diff
@@ -0,0 +1,1232 @@
...
+#include "xls/dslx/type_system/type_info.h"
+#include "xls/dslx/virtualizable_file_system.h"
+#include "xls/ir/bits.h"
+
+namespace xls::dslx {
+namespace {
+
+using ::absl_testing::StatusIs;
+
+struct PT {
```
cdleary: Google C++ style guide requires descriptive names even though they're longer -- e.g. ParseTypecheckResult here

xls/dslx/replace_invocations.cc
```diff
@@ -0,0 +1,504 @@
...
+namespace xls::dslx {
+
+namespace {
+
+template <typename... T>
+struct Overloaded : T... {
+  using T::operator()...;
+};
+template <typename... T>
+Overloaded(T...) -> Overloaded<T...>;
```
cdleary: We use absl::visit and xls::Visitor for this; e.g. https://github.com/google/xls/blob/65aa30d29a36e34040707b7eaf4a8dbb5e06651d/xls/common/visitor.h#L23 you can pull in from that header

cdleary (approved):
LGTM with comments addressed

xls/dslx/replace_invocations_test.cc
```diff
@@ -0,0 +1,1232 @@
...
+      stdlib, /*additional_search_paths=*/std::vector<std::filesystem::path>{},
+      kAllWarningsSet, std::make_unique<RealFilesystem>()));
+  XLS_ASSIGN_OR_RETURN(
+      TypecheckedModule tm,
+      ParseAndTypecheck(text, /*path=*/"test.x", /*module_name=*/"test",
+                        import_data.get()));
+  return ParseTypecheckResult{.import_data = std::move(import_data), .tm = std::move(tm)};
+}
+
+TEST(ReplaceInvocationsTest, NonParametricSimpleReplacement) {
```
richmckeever, line 61: I would suggest parameterizing these tests to run on both typecheck versions so they don't become tech debt for the deletion of TIv1. The contract for what TI needs to include in TypeInfo is not always clear, so any new use outside of IR converter could encounter subtle differences.

Also as a side note, once we delete TIv1 we can probably expose better data from TIv2 for this use case (e.g. TIv2 keeps parametric envs in ExprOrType format natively).

richmckeever (approved):

xls/dslx/replace_invocations_test.cc
richmckeever: There are actually a couple of tests in here that fail with TIv2 and so are failing internally at Google:

[  PASSED  ] 23 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] ReplaceInvocationsTest.ParametricEnumUseImportExplicitReplacement
[  FAILED  ] ReplaceInvocationsTest.ParametricToEnvNonEnumTypeAnnotationErrors

Could you either parameterize it or do a quick hack where you change kDefaultTypecheckVersion in parse_and_typecheck.h and run it that way to debug? Let me know if you need help investigating.

lsrcz: I see. Let me have a look when I got some time.

lsrcz: @richmckeever 

I fixed the `ParametricEnumUseImportExplicitReplacement` test. The DSLX there was invalid.

For `ParametricEnumUseImportExplicitReplacement`, I feel that there might be a gap in TIv2 as it does not handle use modules well.

A minimal reproduction is:

`a.x`:

```dslx
pub enum E : u2 {
  A = 0,
  B = 1,
}
```

`b.x`:

```dslx
#![feature(use_syntax)]
use a;
```

```bash
> ir_converter_main b.x --type_inference_v2
E1006 14:56:24.977736 5811651 populate_table_visitor.cc:120] INTERNAL: XLS_RET_CHECK failure (xls/dslx/type_system_v2/populate_table_visitor.cc:120) result.imported_member != nullptr 
0x105111a80: xls::dslx::(anonymous namespace)::PopulateInferenceTableVisitor::HandleUse()
0x1051236ac: xls::dslx::(anonymous namespace)::PopulateInferenceTableVisitor::DefaultHandler()
0x105050490: xls::dslx::PopulateTable()
0x10504e184: xls::dslx::TypecheckModuleV2()
0x104fce804: xls::dslx::TypecheckModule()
0x104f4f880: xls::dslx::ConvertFilesToPackage()
0x104f4768c: main
0x182282b98: start

E1006 14:56:24.977798 5811651 command_line_utils.cc:50] Could not extract a textual position from error message: INTERNAL: XLS_RET_CHECK failure (xls/dslx/type_system_v2/populate_table_visitor.cc:120) result.imported_member != nullptr : INVALID_ARGUMENT: Provided status is not in recognized error form: INTERNAL: XLS_RET_CHECK failure (xls/dslx/type_system_v2/populate_table_visitor.cc:120) result.imported_member != nullptr 
Error: INTERNAL: XLS_RET_CHECK failure (xls/dslx/type_system_v2/populate_table_visitor.cc:120) result.imported_member != nullptr 
```

TIv2 seems to only handle use of module members, but not a module itself.

lsrcz:
I have just added another test for an edge case where the code wasn't able to handle higher order functions correctly. It is also fixed.

xls/dslx/replace_invocations_test.cc
richmckeever: There is actually no support for `use` syntax in TIv2 as of now. I think in any situations where it works, it's because the parser happens to be conveying the same thing as an equivalent `import` statement. Most tests in other places that rely on `use` syntax have an if statement which returns if TIv2 is flagged on.

lsrcz: I see. Let me just remove or mark the test case skipped. Which do you think is better?

richmckeever: If it's worthwhile coverage, you could probably use import instead of use, and add a TODO to put it back this way. Otherwise maybe mark it disabled.

lsrcz: Just changed to use import

proppy (approved):

PR #3062, opened 2025-09-13
[DSLX:TS] Check test functions are typed as `() -> ()`

cdleary:


xls/dslx/type_system_v2/typecheck_module_v2.cc
```diff
@@ -104,6 +106,26 @@ absl::StatusOr<std::unique_ptr<ModuleInfo>> TypecheckModuleV2(
       TypeInfo * type_info,
       import_data->type_info_owner().GetRootTypeInfo(module.get()));
 
+  // Enforce `#[test]` function signature `() -> ()`.
```
richmckeever, line 109: In v2 it would probably make more sense to just add a HandleTestFunction in validate_concrete_type.cc with the logic you have here (starting from where you have the TestFunction ptr). Or perhaps a HandleFunction which does this if the parent is a TestFunction (not sure if the TestFunction itself gets a type).

cdleary: @richmckeever I tried that at first but it didn't kick in at the right times. I needed to implement it in both places, and then I deleted the Handle* and this one covered the other. I think maybe because it doesn't always need to concretize? It was kind of unclear to me why it wasn't kicking in for all the samples.

PR #3076, opened 2025-09-17
Add force_implicit_token_calling_convention flag to dslx to ir converter

lsrcz:
This pull request adds a flag to the DSLX-to-IR converter that forces all functions to use the implicit-token calling convention.
We need this to enable IR transformations that insert assertions into existing programs, which requires converting the function to the implicit-token convention.
Previously, this was impossible because the activation bit could not be inferred from the converted IR alone.
To support this behavior, we introduce a new flag that instructs the converter to generate implicit tokens for all functions.

This includes a breaking change to the C API to include the flag.

### DIFF
--- a/xls/dslx/ir_convert/function_converter.cc
+++ b/xls/dslx/ir_convert/function_converter.cc
@@ -175,6 +175,9 @@ absl::StatusOr<xls::Function*> EmitImplicitTokenEntryWrapper(
 
 bool GetRequiresImplicitToken(const dslx::Function& f, ImportData* import_data,
                               const ConvertOptions& options) {
+  if (options.force_implicit_token_calling_convention) {
+    return true;
+  }
   std::optional<bool> requires_opt =
       import_data->GetRootTypeInfo(f.owner()).value()->GetRequiresImplicitToken(
           f);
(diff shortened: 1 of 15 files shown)


xls/dslx/ir_convert/function_converter.cc
```diff
@@ -804,22 +807,21 @@ absl::Status FunctionConverter::HandleExternNameRef(
...
-            return DefAlias(c->value(), /*to=*/node);
-          },
-          [&](auto) {
-            return absl::UnimplementedError(absl::StrFormat(
-                "Unsupported module member type %s for external name "
-                "reference: `%s` @ %s",
-                GetModuleMemberTypeName(*member.value()), node->identifier(),
-                node->span().ToString(file_table())));
-          }},
+      Visitor{[&](Function* f) { return DefAlias(f, /*to=*/node); },
```
cdleary: Did this change? hard to see

lsrcz: It is an auto formatter artifact and I reverted it.

lsrcz:
The CI failure does not seem to be related.

```
xls/dslx/cpp_transpiler/cpp_transpiler_test.cc: Found non-absolute includes:
  path/to/some_types.h
  path/to/other_types.h
  /tmp/fake_path.h
  path/to/some_types.h
  path/to/other_types.h
```

richmckeever (approved):

proppy:
@lsrcz can you confirm which version of xlsynth is compatible with this change?

lsrcz:
@proppy This one was on top of commit 3646c6339a2289efb663535a8e97d288b3ec6238. Should I update to the latest main branch?

proppy:
> @proppy This one was on top of commit [3646c63](https://github.com/google/xls/commit/3646c6339a2289efb663535a8e97d288b3ec6238). Should I update to the latest main branch?

Sorry I meant which version of the `xlsynth`'s rust crate is compatible w/ this change.

lsrcz:
@proppy I think it is compatible with xlsynth's rust crate since v0.6.0.

proppy:
> @proppy I think it is compatible with xlsynth's rust crate since v0.6.0.

understood, will make sure we import `v0.6.0` before landing this /cc @hongted

PR #3111, opened 2025-09-25
Function builder guard for TupleIndex with OOB index

cdleary:
Previously this would cause a std::out_of_range which would cause a terminate -- now we give back a Status consistent with the rest of the API surface area.


proppy:
curious if we had an issue reported where this materialized?

PR #3121, opened 2025-09-26
Fix dependencies in xls/public

hzeller:
Also looks like c_api.cc does not depend on c_api_vast.h anymore.

### DIFF
--- a/xls/public/c_api.h
+++ b/xls/public/c_api.h
@@ -21,6 +21,7 @@
 #include "xls/public/c_api_dslx.h"
 #include "xls/public/c_api_format_preference.h"
 #include "xls/public/c_api_ir_builder.h"
+#include "xls/public/c_api_vast.h"  // IWYU pragma: export
 
 // C API that exposes the functionality in various public headers in a way that
 // C-based FFI facilities can easily wrap.
--- a/xls/public/BUILD
+++ b/xls/public/BUILD
@@ -264,6 +264,7 @@ cc_library(
         ":c_api_format_preference",
         ":c_api_impl_helpers",
         ":c_api_ir_builder",
+        ":c_api_vast",
         ":runtime_build_actions",
         "//xls/common:init_xls",
         "//xls/common/file:filesystem",


xls/public/c_api.h
```diff
@@ -21,7 +21,6 @@
 #include "xls/public/c_api_dslx.h"
 #include "xls/public/c_api_format_preference.h"
 #include "xls/public/c_api_ir_builder.h"
-#include "xls/public/c_api_vast.h"
```
hzeller, line 24: Here I am unsure; nothing in the header needs this, but maybe it was meant to by `// IWYU pragma: export` ?

hzeller:
Before sending this to XLS internal reviewers, does this look good @cdleary ?

grebe:
Do you still want to land? Is this superseded by #3184?

hzeller:
Yes, still want to land.
In #3184 I explicitly broke out the changes in xls/public as I was waiting on this one separately.

grebe:
Well, this looks good to me. I'm inclined to land unless @cdleary opposes, it's also not hard to revert/fix later if need be.

cdleary:
I think for the header it's useful to have the central .h file roll up all the public APIs, for the cc doesn't matter though.

grebe:
At least internally, it looks like xlsynth relies on the BUILD dep else a linker error complaining about missing symbols from c_api_vast. @cdleary do you want to keep the BUILD dep or explicitly depend on `c_api_vast` in xlsynth?

cdleary:
@grebe I need a target that rolls up all the C API symbols as a DSO, that was libxls.so which is the shared library wrapper of c_api AFAIK -- it'd be nice to not break that for header cleanliness :-) we need something that is canonically "libxls with all its symbols"

grebe:
Makes sense, I think an `IWYU: pragma export` makes sense then as that will keep the bazel dep as well. @hzeller do you mind making the change?

xls/public/c_api.h
hzeller: Done.

hzeller:
Done. Added the `// IWYU: pragma export` and re-added the `:c_api_vast` to `//xls/public:c_api`

hzeller:
the ci failure looks unrelated.

grebe:
Yeah, CI is a known issue.

grebe (approved):

PR #3133, opened 2025-09-30
[ir] Support packages with keywords as names (contextual keywords)

cdleary:


hzeller:
Uh, nice, now we can have a bits package!

richmckeever (approved):

PR #3152, opened 2025-10-02
Add a specialized module cloner that allows ignoring a set of members

lsrcz:
This pull request adds a specialized module cloner that allows ignoring a set of members. In the C API, we use this to transform a type checked module into another one with some functions stripped.


proppy:
> In the C API, we use this to transform a type checked module into another one with some functions stripped.

Can you give a concrete example?

lsrcz:
> Can you give a concrete example?

Basically all my recent PRs are mostly built towards a DSLX file specializer. We monomorphize all the parametric functions, replace their call sites, and then strip the original parametric versions.

The C API currently uses the type checked module as the root for managing all the AST node lifetimes, and a module is always tied to a typed checked module. This is the reason why we need to transform a type checked module into another one.

xls/dslx/frontend/ast_cloner.cc
```diff
@@ -1335,6 +1389,106 @@ absl::StatusOr<std::unique_ptr<Module>> CloneModule(const Module& module,
   return new_module;
 }
 
+absl::StatusOr<std::unique_ptr<Module>> CloneModuleIgnoreMembers(
+    const Module& module, absl::Span<const AstNode* const> members_to_ignore) {
+  absl::flat_hash_set<const AstNode*> nodes_to_ignore;
+  nodes_to_ignore.reserve(members_to_ignore.size());
+  for (const AstNode* node : members_to_ignore) {
+    if (node != nullptr) {
```
ericastor: Any reason we want to allow `nullptr` here, rather than just returning an `InvalidArgumentError`?

xls/dslx/frontend/ast_cloner.h
```diff
@@ -112,6 +112,13 @@ CloneAstAndGetAllPairs(const AstNode* root,
 absl::StatusOr<std::unique_ptr<Module>> CloneModule(
     const Module& module, CloneReplacer replacer = &NoopCloneReplacer);
 
+// Returns a clone of `module` that omits any top-level members whose
+// identifiers appear in `member_names_to_ignore`. References to those members
+// are not validated or rewritten; the caller is responsible for handling any
+// resulting dangling references.
+absl::StatusOr<std::unique_ptr<Module>> CloneModuleIgnoreMembers(
```
ericastor: Very minor nit - but "ignore" seems confusing. Would it be worth considering a name like `CloneModuleRemovingMembers`?

PR #3171, opened 2025-10-08
Add extraction of call graph in a module

lsrcz:
This adds an extraction procedure of a call graph represented as a mapping from caller to vector of callees. Callees are deduplicated. For `map` built-in, the function argument would be used as the callee.


xls/dslx/type_system/type_info_test.cc
```diff
@@ -126,6 +127,60 @@ fn main() -> u32 { f<u32:0>() }
   EXPECT_EQ(invocations.size(), 1);
 }
 
+TEST(TypeInfoTest, FunctionCallGraphBasic) {
+  ImportData import_data = CreateImportDataForTest();
+  XLS_ASSERT_OK_AND_ASSIGN(
+      TypecheckedModule tm,
+      ParseAndTypecheck(R"(
```
dplassgit: For consistency with the other tests in this file, please extract the inlined source code into its own string. Same comment for the next test.

xls/dslx/type_system/type_info_test.cc
```diff
@@ -126,6 +127,60 @@ fn main() -> u32 { f<u32:0>() }
...
+}
+
+TEST(TypeInfoTest, FunctionCallGraphHandlesMapAndIgnoresBuiltins) {
+  ImportData import_data = CreateImportDataForTest();
+  XLS_ASSERT_OK_AND_ASSIGN(
+      TypecheckedModule tm,
+      ParseAndTypecheck(R"(
+fn inc(x: u32) -> u32 { x + u32:1 }
+
+fn apply(xs: u32[3]) -> u32[3] {
```
dplassgit, line 165: Probably should add tests for:
1. intermodule invocations
2. proc spawns

xls/dslx/type_system/type_info.cc
```diff
@@ -395,6 +396,38 @@ std::string TypeInfo::GetTypeInfoTreeString() const {
   return absl::StrJoin(pieces, "\n");
 }
 
+absl::flat_hash_map<const Function*, std::vector<const Function*>>
+TypeInfo::GetFunctionCallGraph() const {
+  const TypeInfo* top = GetRoot();
+  absl::flat_hash_map<const Function*, absl::flat_hash_set<const Function*>>
+      grouped;
+  for (const auto& [invocation, invocation_data] : top->invocations_) {
```
dplassgit: Optional: `invocation` is never read, so it can be replaced with `_`

xls/dslx/type_system/type_info.cc
```diff
@@ -395,6 +396,38 @@ std::string TypeInfo::GetTypeInfoTreeString() const {
...
+    vec.reserve(callees.size());
+    for (const Function* callee : callees) {
+      vec.push_back(callee);
+    }
+  }
+
+  for (const ModuleMember& member : module()->top()) {
+    if (std::holds_alternative<Function*>(member)) {
+      const Function* function = std::get<Function*>(member);
+      (void)result[function];
```
dplassgit: Forgive my idiocy, but what does this do? Does it check that there's a call from each function in a module? Otherwise it looks like it is a nop?

xls/dslx/type_system/type_info_test.cc
lsrcz: Added these tests

xls/dslx/type_system/type_info.cc
lsrcz: Thanks for pointing it out. Just did some rewrite to the code to make the intent clearer. Here I wanted to make sure that all top-level functions, even if they call nothing, are in the call graph.

proppy:
@lsrcz can you rebase? we see some failure against `//xls/dslx/type_system:type_info_test` internally.

lsrcz:
@proppy Just fixed. There are some inconsistencies between the behavior of v1 and v2. Since v1 will eventually be deprecated, I forced the test case to use TIv2 and only test for its behavior.

PR #3207, opened 2025-10-16
Use generate statements for multidimensional array updates.

meheff:
This significantly reduces the size of the verilog for these operations. Also reduce unnecessary duplication of expressions in some ops, and inline bits typed literals.


xls/codegen/testdata/combinational_generator_test_PrioritySelectOneHot.vtxt
```diff
@@ -7,7 +7,9 @@ module main(
 );
   wire [1:0] one_hot_15;
   wire [2:0] one_hot_18;
+  wire eq_20;
   assign one_hot_15 = {!p, p};
   assign one_hot_18 = {one_hot_15[1:0] == 2'h0, one_hot_15[1] && !one_hot_15[0], one_hot_15[0]};
+  assign eq_20 = one_hot_15 == one_hot_18[1:0];
```
ericastor, line 13: Any idea why this happened? It seems to be a new - and unused - wire?

xls/codegen/testdata/module_builder_test_ArrayUpdate4D.vtxt
```diff
@@ -0,0 +1,155 @@
+module ArrayUpdate4D(
+  input wire [3839:0] array,
+  input wire [1:0] idx0,
+  input wire [1:0] idx1,
+  input wire [319:0] value
+);
+  wire [31:0] array_unflattened[0:3][0:2][0:1][0:4];
```
ericastor, line 7: followup suggestion: Maybe we can make array unflattening happen in a generate statement as well?

xls/codegen/block_generator.cc
```diff
@@ -527,13 +527,55 @@ class BlockGenerator {
...
+      case Op::kAssert:
+        return true;
+      default:
+        return false;
+    }
+  }
+
   // Returns true if the given node should be emitted as an assignment
   // for readability purposes. This is a soft constraint / suggestion.
   bool ShouldEmitAsAssignment(Node* const n, int64_t inline_depth) {
```
ericastor: This is a pretty significant change, so I'm surprised it's in the same PR?

xls/codegen/block_generator.cc
```diff
@@ -527,13 +527,55 @@ class BlockGenerator {
...
-    return (
-        n->HasAssignedName() ||
-        (n->users().size() > 1 && !ShouldInlineExpressionIntoMultipleUses(n)) ||
-        inline_depth > options_.max_inline_depth());
+    bool has_multiple_uses = n->users().size() > 1;
+    if (!has_multiple_uses) {
+      for (Node* user : n->users()) {
```
ericastor: I think we'd be better off with a helper function that just reports whether there are multiple uses, by iterating over the users & "counting" references? This was easy to misread at first attempt.

xls/codegen/module_builder.cc
```diff
@@ -965,18 +1033,22 @@ absl::StatusOr<LogicRef*> ModuleBuilder::EmitAsAssignment(
...
-        // We use a generate loop in the simple case where the array is 1D and
-        // the element type is bits.
-        if (index_types.size() == 1 &&
-            node->GetType()->AsArrayOrDie()->element_type()->IsBits()) {
-          XLS_RETURN_IF_ERROR(EmitArrayCopyAndUpdateViaGenerate1D(
+        Type* base_type = node->GetType();
+        while (base_type->IsArray()) {
+          base_type = base_type->AsArrayOrDie()->element_type();
+        }
+        // Use `generate` to emit the assignment if the base type of the array
```
ericastor: Can we document why it'll be more work to support non-bits types here?

xls/codegen/node_expressions.cc
```diff
@@ -730,7 +730,8 @@ absl::StatusOr<Expression*> NodeToExpression(
 
 bool ShouldInlineExpressionIntoMultipleUses(Node* node) {
   return node->Is<BitSlice>() || node->op() == Op::kNot ||
-         node->op() == Op::kNeg;
+         node->op() == Op::kNeg ||
+         (node->Is<xls::Literal>() && node->GetType()->IsBits());
```
ericastor, line 734: No change needed here - just wanted to thank you for spotting this opportunity!

xls/codegen/block_generator.cc
ericastor: Having understood the function now - no need for any change, it seems reasonable.

xls/codegen/testdata/combinational_generator_test_ArrayUpdate2DUpdateArrayVariableIndex.vtxt
```diff
@@ -16,11 +16,14 @@ module ArrayUpdate2DUpdateArrayVariableIndex(
...
-  assign array_update_8[0][1] = idx == 37'h00_0000_0000 ? value_unflattened[1] : a_unflattened[0][1];
-  assign array_update_8[0][2] = idx == 37'h00_0000_0000 ? value_unflattened[2] : a_unflattened[0][2];
-  assign array_update_8[1][0] = idx == 37'h00_0000_0001 ? value_unflattened[0] : a_unflattened[1][0];
-  assign array_update_8[1][1] = idx == 37'h00_0000_0001 ? value_unflattened[1] : a_unflattened[1][1];
-  assign array_update_8[1][2] = idx == 37'h00_0000_0001 ? value_unflattened[2] : a_unflattened[1][2];
   assign out = {{array_update_8[1][2], array_update_8[1][1], array_update_8[1][0]}, {array_update_8[0][2], array_update_8[0][1], array_update_8[0][0]}};
+  genvar array_update_8__index_0;
+  genvar array_update_8__update_0;
+  generate
+    for (array_update_8__index_0 = 0; array_update_8__index_0 < 2; array_update_8__index_0 = array_update_8__index_0 + 1) begin : array_update_8__gen_0
```
grebe, line 23: I believe our preferred style is for `gen_` to be a prefix on the label and to not have generate regions `(generate, endgenerate)`. Not sure how style-opinionated you are, I'd be OK with some parameterization but it seems pretty cosmetic.

My read is that our style guide does **not** require this, but I think it's more common and more readable to write `for (genvar <id>...)`. I'm fine as-is, but thought it was worth mentioning.

grebe (changes requested):
This is nice! Thanks for doing this

xls/codegen/testdata/combinational_generator_test_PrioritySelectOneHot.vtxt
meheff: In the IR the eq_20 feeds an assert which we don't emit in plain verilog. You can see it in the system verilog:

https://github.com/google/xls/blob/e87c628235ca70e2db2ee0d5aaa986509f9d7a7b/xls/codegen/testdata/combinational_generator_test_PrioritySelectOneHot.svtxt#L13d

This appears now because an assert is considered a multiuse now and so eq_20 is no longer emitted inline (which for plain verilog where we drop asserts means we don't emit it at all).

A better approach for plain verilog might be to delete the assert IR ops and run DCE before emitting code.

It is not completely clear why eq_20 is was not inlined in the SystemVerilog at head.

xls/codegen/block_generator.cc
meheff: Agreed. Cleaned up a bit.

xls/codegen/module_builder.cc
meheff: Actually, non-bits types should work. I was thinking that if the base type was a tuple with nested arrays it would not work, but we represent such tuples as a flattened bit vector so a single assignment works fine. Deleted the old function.

xls/codegen/testdata/module_builder_test_ArrayUpdate4D.vtxt
meheff: Agreed. I can work on that followup.

xls/codegen/testdata/combinational_generator_test_ArrayUpdate2DUpdateArrayVariableIndex.vtxt
meheff: Ah, I thought plain verilog required the generate/endgenerate, but I guess not after I looked at the spec. I can clean it up in a followup. It will actually make things simpler because each loop is then more self contained.

meheff:
The failure //xls/jit:jit_channel_queue_test is unrelated and seems to be broken at head at least for github CI.

allight:
Would you mind rebasing and re-running rebuild-goldens? That script was missing a bunch of tests it should have included but it has been updated now.

meheff:
(I missed your earlier comment). Updated and fixed tests. Also fixed the golden file script which was broken in OSS.

ericastor (commented):
LGTM, but I'll let @grebe weigh in before this lands.

xls/codegen/testdata/combinational_generator_test_ArrayUpdate2DUpdateArrayVariableIndex.vtxt
grebe: Sounds good- would you mind adding a TODO?

meheff:
Add "Reviewing Internally"  label?  Thanks

PR #3215, opened 2025-10-18
[c_api:vast] VAST statements / control constructs

cdleary:
Adds VAST C APIs for procedural control: Conditional (if/else[-if]) and Case (items/default), plus blocking assignments (block-level and file-level factory).


meheff:
Just revisiting some languishing PRs. I just rebased this. I don't imagine this should cause any issues google-side.

ericastor:
This PR is still marked a draft; is it supposed to be reviewed? If so, it should probably not be marked a draft.

meheff:
Ah, I didn't notice it was a draft. That's probably why it's been sitting there so long :-) Leary is out and I'm trying to minimize the number of patches we have on our internal branch.

PR #3216, opened 2025-10-20
Add support for #(...) style parameters to modules.

meheff:
This parameter form is used pretty universally and it's been supported since Verilog-2001 and always in SystemVerilog. In addition to being much more common than the old form, the new form enables use of parameters in input/output port definition not possible with parameter defined in the module body. Example:

```
module foo #(
  parameter int WIDTH = 8
) (
  input  logic [WIDTH-1:0] data_in,
  output logic [WIDTH-1:0] data_out
);
endmodule
```


meheff:
The failing //xls/jit:jit_channel_queue_test test is unrelated and seems to be broken at head at least for github CI.

allight:
Ok so this is failing internally because the xlsynth rust tests still expect the old version. Could you push an update to those to make them accept either then we can merge this.

meheff:
This is problem for you folks too right? if you happened to make this exact change to codegen the rust tests will break I presume. Can you ignore the Rust test failures or disable them in some way? As is, the crate tests inside google3 are a blocker for changes which might alter the generated code.

meheff:
In the interest of expediency, I removed the extra space between the module name and the open paren. This change now just includes the new style parameters on modules. Hopefully, this will allow it to land much easier.

allight:
We can ignore them, I figured out how to get rid of the test by talking to the rust folks (for internal folks; add file to the list in xlsynth copy.bara.sky and run `copybara ./copy.bara.sky --same-version`). This is a much easier to review change though so thanks we'll probably just take this one. Thanks for the help.

allight (approved):

Issue #3104, opened 2025-09-22
SystemVerilog codegen should use `logic` instead `reg`


proppy:
The two types are described as equivalent in the SystemVerilog standard:

> **6.11.2 2-state (two-value) and 4-state (four-value) data types**
> 
> The keyword reg does not always accurately describe user intent, as it could be perceived to imply a
> hardware register. The keyword logic is a more descriptive term. logic and reg denote the same type. 

And in Sutherland System Verilog for Design

> ** 3.3.2 SystemVerilog 4-state variables **
> [...]
> Semantically, a variable of the logic data type is identical to the
Verilog reg type. The two keywords are synonyms, and can be
used interchangeably (except that the reg keyword cannot be
paired with net type keywords, as discussed in section 3.3.4 on page
47). 

See also https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#use-logic-for-synthesis

PR #3237, opened 2025-10-23
Improve support for `logic` variables in VAST.

meheff:
Enable `logic` module ports like so:

  module foo (
    logic a,
    logic b,
  ) ...

Expose this functionality on the C API, and add always_comb support to the C API as well.


proppy:
Related https://github.com/google/xls/issues/3104 (and wip https://github.com/google/xls/pull/3106), but your solution is probably better and easier to land!

meheff:
Ah, I didn't see those. Probably the next step is to change codegen (module_builder.cc) to use logic instead of wire and reg. when use_verilog is true.

PR #3257, opened 2025-10-28
Extend the VAST api.

meheff:
Add  unbased unsized one ('1)  and richer support for types in localparam.  Also add "int" type. Previously we just had the older "integer" type. The differences is "integer" is 4-state while "int" is 2-state (and SystemVerilog only). "int" is generally preferred because in most integer type use cases 2-state is desired.

Expose these on the VAST C API.

### DIFF
--- a/xls/codegen/vast/vast.h
+++ b/xls/codegen/vast/vast.h
@@ -267,7 +267,7 @@ class ScalarType final : public DataType {
   bool is_signed_;
 };
 
-// Represents an integer type. Example:
+// Represents an `integer` type (4-value integer type). Example:
 //   integer foo;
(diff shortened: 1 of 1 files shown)


xls/codegen/vast/vast.h
```diff
@@ -291,6 +291,30 @@ class IntegerType final : public DataType {
   bool is_signed_;
 };
 
+// Represents an integer type. Example:
```
ericastor: We should probably update this comment for `int` rather than `integer`? (Might even be worth explaining the distinction.)

meheff: Done.

PR #3268, opened 2025-10-31
Add constexpr if

rw1nkler:
This PR adds support for the constexpr `if` expression. 

Information about the `const` modifier is stored in the `Conditional` class and used later during type checking. For constexpr `if`, separate type variables are assigned to each branch of the condition. At this stage, the result is not directly bound to them. During the `TypeInfo` generation, the constexpr condition is evaluated, and one branch of the if is selected. Its type is unified, and the entire if is annotated with that type. The bytecode emitter and IR converter have been updated to inline the selected branch, skipping the unchosen one.

Proper tests for this feature are still needed and will be added soon.

 One issue not yet handled correctly in this PR is the conditional spawning of procs, for which an example has been added that shows the problem. Currently, all spawns affect the type, which should not happen as only the selected branch of constexpr `if` should be considered. One possible approach is to enable constexpr evaluator in the `PopulateInferenceTableVisitor` and skip traversing the branch that was not taken. However, the `PopulateInferenceTableVisitor` does not have access to the parametric environment, and the condition relying on the parametrics seems to be the most imprtant use-case for using constexpr `if`  when spawning procs.

### DIFF
--- a/xls/examples/BUILD
+++ b/xls/examples/BUILD
@@ -1423,12 +1423,18 @@ xls_dslx_ir(
     dslx_top = "Main",
     ir_file = "const_if.ir",
     library = ":const_if_dslx",
+    ir_conv_args = {
+        "lower_to_proc_scoped_channels": "true",
+    },
 )
 
 xls_dslx_opt_ir(
     name = "const_if_opt_ir",
     dslx_top = "Main",
     library = ":const_if_dslx",
+    ir_conv_args = {
+        "lower_to_proc_scoped_channels": "true",
+    },
 )
 
 xls_dslx_test(


xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -500,10 +500,26 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
...
+      XLS_RETURN_IF_ERROR(
+          table_.SetTypeVariable(node->consequent(), type_variable));
+      XLS_RETURN_IF_ERROR(
+          table_.SetTypeVariable(ToAstNode(node->alternate()), type_variable));
+    } else {
+      // For constexpr if we create separate variables for each if branch.
+      // Later during the generation of TypeInfo, after resolving parametrics,
+      // and evaluating the test condition, we force the type of the selected
+      // if branch on the conditional node.
+      XLS_RETURN_IF_ERROR(
```
richmckeever, line 514: I think this would be more in-model if we would define a new type of internal type annotation, like ConstConditionalTypeAnnotation, and set an instance of that on the Conditional node in the const case. The ConstConditionalTypeAnnotation could have the const test Expr as a member, and a TVTA for each possible branch.

TypeAnnotationResolver would then perform the resolution logic that you currently have inside GenerateTypeInfo (in a manner similar to MemberTypeAnnotation, ElementTypeAnnotation, etc.).

Aside from that this all looks reasonable.

xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -500,10 +500,26 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
     // the same type as 'D'. The test 'a' must be a bool, so we annotate it as
     // such.
     const NameRef* type_variable = *table_.GetTypeVariable(node);
-    XLS_RETURN_IF_ERROR(
-        table_.SetTypeVariable(node->consequent(), type_variable));
-    XLS_RETURN_IF_ERROR(
-        table_.SetTypeVariable(ToAstNode(node->alternate()), type_variable));
+    if (!node->IsConst()) {
```
richmckeever: Usually we prefer to use positive conditions in a C++ if statement, even if it leads to the less common branch being first (so flip the 2 branches here).

rw1nkler: fixed

xls/dslx/type_system_v2/populate_table_visitor.cc
rw1nkler: Let me know if this is what you expected

xls/dslx/frontend/ast.h
```diff
@@ -957,6 +959,38 @@ class GenericTypeAnnotation : public TypeAnnotation {
...
+  }
+
+  std::vector<AstNode*> GetChildren(bool want_types) const override {
+    return {};
+  }
+
+  const Expr* test() const { return test_; }
+  TypeAnnotation* consequent_type() const { return consequent_type_; }
+  TypeAnnotation* alternate_type() const { return alternate_type_; }
+  std::string ToString() const override { return "ConstConditional"; }
```
richmckeever: The output of ToString() should contain the details, as in other type annotations.

xls/dslx/frontend/ast.h
```diff
@@ -2323,13 +2360,16 @@ class Conditional : public Expr {
   // blocks of this if/else-if.../else ladder.
   std::vector<StatementBlock*> GatherBlocks();
 
+  bool IsPartOfLadder() const;
```
richmckeever: Add a comment explaining what this means. Also, a better name might be "IsInnerPartOfLadder" or "IsElseIf" because it doesn't look like it counts the outer if.

xls/dslx/frontend/ast_cloner.cc
```diff
@@ -1106,6 +1106,17 @@ class AstCloner : public AstNodeVisitor {
     return absl::OkStatus();
   }
 
+  absl::Status HandleConstConditionalTypeAnnotation(
+      const ConstConditionalTypeAnnotation* n) override {
+    XLS_RETURN_IF_ERROR(ReplaceOrVisit(n->consequent_type()));
+    XLS_RETURN_IF_ERROR(ReplaceOrVisit(n->alternate_type()));
+    old_to_new_[n] = module(n)->Make<ConstConditionalTypeAnnotation>(
+        n->span(), n->test(),
```
richmckeever: It's unusual not to clone the test() here, and could be problematic in some contexts. If we really need to preserve it in order to look up the constexpr, let's add a comment.

richmckeever (commented):
Let's add tests in typecheck_module_v2_test to prove that for example it's OK to have disparate types in the branches of the const if in a function like:

```
fn f<N: u32>() -> uN[N] {
  const if (N == 1) {
    u1:0
  } else {
    zero!<uN[N]>();
  }
}
```

xls/dslx/frontend/ast.h
wsipak: Thank you for the suggestions. I think "IsElseIf" is a good choice.
I've added a comment and renamed the function.

xls/dslx/frontend/ast.h
wsipak: Alright, right now the output includes the string `ConstConditional`, along with details about the consequent and alternate types.

xls/dslx/frontend/ast_cloner.cc
wsipak: I've modified this function to clone the `test()` node as well. Thank you.

wsipak:
I've applied changes requested in the review.
I've included a test with the function:
```
fn f<N: u32>() -> uN[N] {
  const if (N == 1) {
    u1:0
  } else {
    zero!<uN[N]>()
  }
}
```
This previously failed due to wrong `TypeInfo` object being used for the conditional evaluation - this is now fixed.

I've also included an example with conditional proc spawning. In order to spawn procs in `const if`, a change was required in the `ProcConfigIrConverter` which is also present in this PR.

FYI @richmckeever

proppy:
@richmckeever can you re-review and approve?

wsipak:
I've rebased these changes on `main` and provided the args to use proc-scoped channels in the `const_if` example.

magancarz:
I've rebased these changes on main and removed a change from the deleted `deduce.cc` file.

PR #3278, opened 2025-11-02
Update generate loop syntax.

meheff:
Use a more compact, modern generate loop syntax like this:

for (genvar i = 0; ...) begin
  ...
end

Previously loops were constructed like:

genvar i;
generate
  for (i = 0; , ... ) begin
     ...
  end
endgenerate


xls/codegen/module_builder.cc
```diff
@@ -649,30 +648,41 @@ absl::Status ModuleBuilder::EmitArrayCopyAndUpdateViaGenerate(
...
   // only if the update value is an array.
-  std::vector<LogicRef*> update_genvars;
+  std::vector<GenerateLoop*> update_loops;
+
+  GenerateLoop* innermost_loop = nullptr;
+  auto add_loop = [&](const std::string& genvar_name, Expression* init,
+                      Expression* limit, std::optional<std::string> label) {
+    GenerateLoop* loop = nullptr;
+    if (innermost_loop == nullptr) {
+      loop = module_->Add<GenerateLoop>(SourceInfo(), genvar_name, init, limit,
```
grebe, line 662: Nit: seems slightly cleaner to do something like `loop = innermost_loop; if (loop == nullptr) { loop = module_ }` and then have a single call to `Add<T>()`. Might not even need `loop` then and just operate on `innermost_loop`.

meheff: Alas, innermost_loop and module_ are different, incompatible types so you can't have a single call to Add.

meheff:
Rebased to head.

meheff:
Rebased to fix failing IR conversion tests.

mikex-oss (approved):

PR #3279, opened 2025-11-02
Add WidthCast operation to VAST.

meheff:
Examples: 14'(42), MyWidth'(x), (MyWidth+1)'(100)


xls/codegen/vast/vast.cc
```diff
@@ -672,6 +672,46 @@ std::string ModulePort::ToString() const {
...
+  if (all_digits) {
+    return true;
+  }
+
+  const char first = s.front();
+  if (!absl::ascii_isalpha(first) && first != '_') {
+    return false;
+  }
+  for (char ch : s.substr(1)) {
+    if (!absl::ascii_isalnum(ch) && ch != '_') {
```
mikex-oss: `$` is a valid character after the first, so it should be allowed here.

> A simple identifier shall be any sequence of letters, digits, dollar signs (`$`), and underscore characters (`_`)....The first character of a simple identifier shall not be a digit or `$`; it can be a letter or an underscore.

xls/codegen/vast/vast.cc
```diff
@@ -672,6 +672,46 @@ std::string ModulePort::ToString() const {
                          xls::verilog::ToString(direction), name());
 }
 
+static bool IsDigitsOrIdentifier(std::string_view s) {
```
mikex-oss, line 678: `IsDigitsOrSimpleIdentifier`

Not sure if/how VAST represents escaped identifiers, but seems better to document it's not handled here.

xls/codegen/vast/vast.cc
```diff
@@ -672,6 +672,46 @@ std::string ModulePort::ToString() const {
...
+  for (char ch : s) {
+    if (!absl::ascii_isdigit(ch)) {
+      all_digits = false;
+      break;
+    }
+  }
+  if (all_digits) {
+    return true;
+  }
+
```
mikex-oss, line 693: Is it worth a lookup against keywords to skip the identifier loop?

xls/codegen/vast/vast.cc
```diff
@@ -672,6 +672,46 @@ std::string ModulePort::ToString() const {
...
+}
+
+std::string WidthCast::Emit(LineInfo* line_info) const {
+  LineInfoStart(line_info, this);
+  LineInfoEnd(line_info, this);
+  std::string width_expr = width_->Emit(line_info);
+  if (IsDigitsOrIdentifier(width_expr)) {
+    // Parentheses are not needed for digits or identifiers.
+    return absl::StrFormat("%s'(%s)", width_expr, value_->Emit(line_info));
+  } else {
```
mikex-oss: Nit: no need for else since if branch returns (just dedent it).

xls/codegen/vast/vast.cc
meheff: Added a comment about not handling escaped identifiers.

xls/codegen/vast/vast.cc
meheff: Good point. Added.

xls/codegen/vast/vast.cc
meheff: Probably not. It's not peformance critical and this is just a cosmetic heuristic to avoid superfluous parentheses in common cases.

xls/codegen/vast/vast.cc
meheff: Fixed.

meheff:
I also added a missing function to the C API while was here.

meheff:
The test failures look unrelated and repro at head.

mikex-oss:
> The test failures look unrelated and repro at head.

This was a TIv1 vs TIv2 diff, should be clean now as a result of https://github.com/google/xls/commit/1798223b8b999d8198ef58f9e3be2127db10e624.

meheff:
Rebased to fix the failing tests. Also add TypeCast as that was also missing from VAST.

PR #3302, opened 2025-11-05
Add inout port kind to VAST module and fix PackedArrayType

meheff:
The dimension orders in PackedArrayTypes were not being emitted in the right order. For example a PackedArrayType whose array dimensions are 20 (major) and 10 (minor) where each element is a bit-vector of size 8, previously the following would be emitted

```
   logic [7:0][19:0][9:0] foo;
```

However, the correct order (emitted with this change) is:

```
   logic [19:0][9:0][7:0] foo;
```

### DIFF
--- a/xls/codegen/vast/vast.cc
+++ b/xls/codegen/vast/vast.cc
@@ -500,7 +500,7 @@ PackedArrayType* VerilogFile::PackedArrayType(int64_t element_bit_count,
   // For packed arrays we always use a bitvector (non-scalar) for the element
   // type when the element bit width is 1. For example, if element bit width is
   // one and dims is {42} we generate the following type:
-  //   reg [0:0][41:0] foo;
+  //   reg [41:0][0:0] foo;
   // If we emitted a scalar type, it would look like:
   //   reg [41:0] foo;
   // Which would generate invalid verilog if we index into an element
--- a/xls/codegen/vast/vast.h
+++ b/xls/codegen/vast/vast.h
@@ -400,11 +400,11 @@ class ArrayTypeBase : public DataType {
   DataType* element_type() const { return element_type_; }
 
   // Returns the dimensions for the type, excluding the element type. For
-  // example, the net type for "wire [7:0][42:0][3:0] foo;" has {43, 4} as the
+  // example, the net type for "wire [42:0][3:0][7:0] foo;" has {43, 4} as the
   // array dimensions. The inner-most dimension (of the contained
   // `BitVectorType`) is not included as that is considered for type purposes as
   // the underlying array element type ("wire [7:0]"). Similarly, for an
-  // unpacked array, the net type for "wire [7:0][42:0] foo [123][7];" has {123,
+  // unpacked array, the net type for "wire [42:0][7:0] foo [123][7];" has {123,
   // 7} as the unpacked dimensions. The ordering of indices matches the order in
   // which they are emitted in the emitted Verilog.
   absl::Span<Expression* const> dims() const { return dims_; }
@@ -445,7 +445,7 @@ class PackedArrayType final : public ArrayTypeBase {
 };
 
 // Represents an unpacked array of bit-vectors or packed array types. Example:
-//   wire [7:0][42:0][3:0] foo [1:0];
+//   wire [42:0][3:0][7:0] foo [1:0];
 // Where [1:0] is the unpacked array dimensions.
 class UnpackedArrayType final : public ArrayTypeBase {
  public:
@@ -2903,8 +2903,11 @@ class VerilogFile {
                                                 const SourceInfo& loc,
                                                 bool is_signed = false);
 
-  // Returns a packed array type. Example:
-  //   wire [7:0][41:0][122:0] foo;
+  // Returns a packed array type. Example, PackedArrayType(8, {42, 123})
+  // produces:
+  //   wire [41:0][122:0][7:0] foo;
+  //
+  // Dimensions are specified in major to minor order.
   verilog::PackedArrayType* PackedArrayType(int64_t element_bit_count,
                                             absl::Span<const int64_t> dims,
                                             const SourceInfo& loc,


mikex-oss:
@meheff it looks like this is missing corresponding changes to https://github.com/google/xls/blob/main/xls/codegen/vast/translate_vast_to_dslx.cc

For example:

https://github.com/google/xls/blob/68d6192d221543fbdbe3c806a4528d0c5a0848a3/xls/codegen/vast/translate_vast_to_dslx_test.cc#L1256-L1262

IIUC, this should be 9 bits packed into a size 4 array (since the literals represent max index).

But the expected value is: 

https://github.com/google/xls/blob/68d6192d221543fbdbe3c806a4528d0c5a0848a3/xls/codegen/vast/translate_vast_to_dslx_test.cc#L1279

which is a 4 bit element packed into a size 9 array.

Also, if you don't mind, please do an audit of the surrounding comments that reference the bad ordering semantics. E.g.

https://github.com/google/xls/blob/68d6192d221543fbdbe3c806a4528d0c5a0848a3/xls/codegen/vast/vast.h#L399-L403

https://github.com/google/xls/blob/68d6192d221543fbdbe3c806a4528d0c5a0848a3/xls/codegen/vast/vast.cc#L498-L500

meheff:
Fixed up the comments to reflect the new, correct dimension ordering.

For the translate_vast_to_dslx stuff, nothing needs to be changed because the bug was only in the way PackedArrayType is printed out in Verilog. The data structure itself is fine and that is what translate_vast_to_dslx uses (not the Verilog serialization). I admittedly was totally puzzled for a few minutes on how translate_vast_to_dslx_test was not broken by my change until I realized that.

Part of what was confusing is that our dimension order in DSLX is reversed from verilog so it *looks* like the DSLX test is using the old, wrong dimension order for PackedArrayType, but it's just the way DSLX is.

mikex-oss:
I spent awhile trying to make sure I didn't miss anything, and I'm still not convinced what you have is right. The problem is the Verilog order is the same as the DSLX order now.

https://github.com/google/xls/blob/68d6192d221543fbdbe3c806a4528d0c5a0848a3/xls/codegen/vast/translate_vast_to_dslx_test.cc#L1256-L1262

The `bit_vector_type` is the thing with `size_expr` of `8` with `size_expr_is_max_ = true`. The `dims` with `dims_are_max_ = true` has a single element of `3`. Your PR fixes the SV representation to `<element_type> [3:0][8:0]`, i.e. emit the `bit_vector_type->DimensionsString()` last.

The DSLX that passes in the test is `bits[4][9]`, which is the same order. Per the DSLX documentation:

> Multidimensional array sizes are declared from the innermost dimension to the outermost dimension (arr: u32[INNER_DIMENSION][OUTER_DIMENSION]) and indexed from the outermost dimension to the innermost dimension (arr[OUTER_DIMENSION][INNER_DIMENSION]).

The inner `bit_vector_type` should come first.

Note that this is causing a bunch of internal code to fail type checking because of this.

meheff:
This change only changes the way PackedArrayTypes are emitted in verilog not the internal representation. AIUI translate_vast_to_dslx does not use the verilog emission except in some error messages so my change should not affect any translate_vast_to_dslx behavior.

That said, the translate_to_dslx functionality is currently wrong at head. I believe it would translate `logic [10:0][9:0][8:0] foo` to `foo: bits[9:0][10:0][8:0]` which is wrong. It should be `foo: bits[8:0][9:0][10:0]`. I think the offending line is here where the bit vector element type is pushed to the *end* of the dims (should be put at the beginning): https://github.com/google/xls/blob/main/xls/codegen/vast/translate_vast_to_dslx.cc#L413

I suspect the failing internal tests must be relying on the incorrect verilog emission of the packed array types. Are these tests round-tripping through verilog? Perhaps there is a corresponding bug on the verilog->VAST translation side (the nonpublic bit) which cancels out the bug this change fixes (incorrect emission of packedarraytype into verilog)?

Also, this looks a little weird: https://github.com/google/xls/blob/main/xls/dslx/translators/dslx_to_verilog.cc#L128-L154 For bitvector verilog types it is constructing an array of scalars rather than a bitvector type.

Edit: Corrected dimension order in the example.

mikex-oss:
>  I suspect the failing internal tests must be relying on the incorrect verilog emission of the packed array types. Are these tests round-tripping through verilog?

Yes, there are internal round-tripping tests, and because we pull the first Verilog dimension as the `bit_vector_type` instead of the last, it doesn't round trip. When I patched the PR to fix the internal-only issues, that's what exposed many more failures due to the xls/codegen/vast/translate_vast_to_dslx.cc component that's in OSS.

Ultimately, this is a broader mistake for relative ordering of the element type size vs. array dimensions that needs to be fixed together. Thanks for identifying the vast_to_dslx issue.

Re: dslx_to_verilog, I think it effectively does the same thing?

https://github.com/google/xls/blob/fe1af71adb39335d952adc170d860ed3058d5eb0/xls/dslx/translators/testdata/dslx_to_verilog_test_MultiDimTest.vtxt#L12-L16

mikex-oss (approved):

mikex-oss:
Hmm, I'm not exactly sure how to coordinate this change with the xlsynth crate which breaks presubmits due to the following test:

```
thread 'tests::test_slice_on_both_sides_of_assignment' (896) panicked at [third_party/rust/xlsynth/v0_6/tests/vast_test.rs:165]:
assertion `left == right` failed
  left: "module my_module;\n  wire [2:0][4:0][1:0] a;\n  wire [1:0] b;\n  wire [2:0] c;\n  assign a[1][2][3:4] = b[1:0];\n  assign a[3:4] = c[2:1];\nendmodule\n"
 right: "module my_module;\n  wire [1:0][2:0][4:0] a;\n  wire [1:0] b;\n  wire [2:0] c;\n  assign a[1][2][3:4] = b[1:0];\n  assign a[3:4] = c[2:1];\nendmodule\n"
```

It looks like @meheff fixed this test in https://github.com/xlsynth/xlsynth-crate/commit/759691ef26450bc2c9c407a5ef8926ed6759e7a7#diff-4bda422219af53bbd9db823389d59fe3196870d719092af290924f000b497e60.

@proppy ?

meheff:
At least judging from the verilog types in the comments, it looks like the dslx_to_verilog tests are correct so the brokenness-canceling-out-brokenness must be in verilog->VAST layer. https://github.com/google/xls/blob/main/xls/dslx/translators/testdata/dslx_to_verilog_test_MultiDimTest.vtxt#L9-L16

Importing the rust crate into Google and using the tests as a presubmit for XLS makes many changes which affect the generated RTL impossible to land. The usual flow would be make an XLS then we pull in those XLS changes into xlsynth and update the Rust code to work with it, but that doesn't work for the way things are setup inside Google. The rust-crate and the XLS change would have to happen atomically. Can you disable the rust tests?

One other thing I noticed about translate_vast_to_dslx is that all types, etc, are put in VAST modules (ie verilog modules) which isn't right. They should be in packages. Likely the VAST->verilog translator has this similar problem. I think I was the one who reviewed it and had more Verilog experience than the author and should have caught that earlier. It works, but is odd and may be awkward if more functionality is added.

mikex-oss:
Landed in https://github.com/google/xls/commit/4279936f0de71b473408848b086937a7101237cc.

PR #3317, opened 2025-11-06
Fix issue where --type_inference_v2 is ignored in ir_converter_main

meheff:


xls/dslx/ir_convert/ir_converter.cc
```diff
@@ -620,11 +620,11 @@ absl::Status AddContentsToPackage(
...
   XLS_RETURN_IF_ERROR(
       module->SetConfiguredValues(convert_options.configured_values));
-  absl::StatusOr<TypecheckedModule> typechecked_module =
-      TypecheckModule(std::move(module), path_value, import_data,
-                      convert_options.type_inference_v2
-                          ? std::make_optional(TypeInferenceVersion::kVersion2)
-                          : std::nullopt);
+  absl::StatusOr<TypecheckedModule> typechecked_module = TypecheckModule(
+      std::move(module), path_value, import_data,
+      convert_options.type_inference_v2
```
richmckeever, line 625: Changing this from a one-way force-on flag to 2-way is a little bit tricky. If we really want to do it, I think we should make it an optional in ConvertOptions as well, or it's too easy to lose coverage by creating an ad hoc ConvertOptions without it set properly.

Note that there is a type_inference_v1 module attribute that can be put in a top-level .x file to use v1, which may be easier for your use case.

meheff: If it's a bit hairy then let's not do this given the limited life span of v1. I can use the v1 feature attribute. I didn't know about that. Thanks!

PR #3324, opened 2025-11-06
Allow more constructs in a generate loop in VAST

meheff:
Previously the generate loop contained a StatementBlock which limited what could be put inside of loops to statements. However, other constructs can be put in loops like always blocks, etc. This change expands what may be placed in the body of a GenerateLoop.


xls/codegen/vast/vast.h
```diff
@@ -2266,6 +2234,45 @@ class VerilogFunctionCall final : public Expression {
   std::vector<Expression*> args_;
 };
 
+using GenerateLoopMember =
+    std::variant<LocalParam*, Statement*, AlwaysComb*, AlwaysFf*, AlwaysFlop*,
```
mikex-oss: Isn't `Statement` too general? Shouldn't this be some subset of `ModuleMember` and explicitly include things like `GenerateLoop` and `ContinuousAssignment`?

meheff: Yeah you're right, it's too general or more precisely kinda wrong. I think actually having it hold ModuleMember is probably the right thing to do since the body of the generate loop is effectively at the module scope. The only potentially weird thing is ModuleSection, but that is not really part of the AST and is instead a transparent container which makes it easier to construct VAST as you can add stuff to different points in the module rather than having to build everything sequentially. And there is no reason why the loop body couldn't  use ModuleSection for the same purpose. I'll start on that change.

meheff:
Updated so generate loop bodies hold module members.

meheff:
Anything else you need from me on this PR?

mikex-oss (approved):

Issue #3293, opened 2025-11-04
[enhancement] solvers::z3::ScopedVerifyEquivalent (and Proc and check_ir_equivalence_main) should check assert equivalence


allight:
### What's hard to do? (limit 100 words)

If you have passes or transforms which add asserts changes done by opt should not change the circumstances that asserts trigger on. Currently none of these can be verified since our z3 translator does not support asserts.

~~This would have let us catch [3288](https://github.com/google/xls/issues/3288) earlier for example.~~ Edit: Nope that seems to be a different error. In any event this would still generally be useful.

### Current best alternative workaround (limit 100 words)

Manual/none

### Your view of the "best case XLS enhancement" (limit 100 words)

ScopedVerifyEquivalent (and maybe the devtools check_ir_equivalence_main) should do pre-transforms on their inputs to match up asserts with the same label on both sides and ensure they remain identical. Asserts with no corresponding assert on the other side should be checked they can never trigger.

PR #3361, opened 2025-11-12
Gracefully support multiple observers in IR evaluators

DeeveshChowdary:
I added a small `CompositeEvaluationObserver` so the interpreter/JIT can fan out `NodeEvaluated` calls for me. The CLIs now just collect every observer the flags enable and either pass it straight through (one observer) or wrap the set in the composite.

For proc/block flows I also ditched the old variant helper and introduced a `ScopedObserver` class that owns the scoped observers, hands the right pointer to the runtimes, and keeps the coverage pause/resume behavior. 

I can run soak up tracing + coverage (or whatever comes next) in a single pass now without touching the evaluator APIs.

### DIFF
--- a/xls/tools/eval_ir_main.cc
+++ b/xls/tools/eval_ir_main.cc
@@ -556,11 +556,10 @@ absl::Status Run(Package* package, absl::Span<const ArgSet> arg_sets_in) {
     }
   }
 
-  EvaluationObserver* observer =
-      composite_observer.empty() ? nullptr : &composite_observer;
   std::optional<EvaluationObserver*> observer_opt =
-      observer == nullptr ? std::nullopt
-                          : std::optional<EvaluationObserver*>(observer);
+      composite_observer.empty()
+          ? std::nullopt
+          : std::optional<EvaluationObserver*>(&composite_observer);
 
   // Copy the input ArgSets because we want to write in expected values if they
   // do not exist.


xls/interpreter/observer.h
```diff
@@ -57,6 +58,34 @@ class CollectingEvaluationObserver : public EvaluationObserver {
   absl::flat_hash_map<Node*, std::vector<Value>> values_;
 };
 
+// Forwards callbacks to multiple observers.
+class CompositeEvaluationObserver : public EvaluationObserver {
```
allight, line 63: A lot of this PR would be simpler if there was an 'AddObserver' & 'RemoveObserver' methods on this as well as an 
'observers' function to get the list of current observers.

Once you have this you can use `!observers.empty()` to see if any observers need to be registered. It also means you can just add and remove things as required.

xls/tools/eval_ir_main.cc
```diff
@@ -528,20 +528,16 @@ absl::Status Run(Package* package, absl::Span<const ArgSet> arg_sets_in) {
...
   ScopedRecordNodeCoverage cov(
       coverage_enabled ? absl::GetFlag(FLAGS_output_node_coverage_stats_proto)
                        : std::nullopt,
       coverage_enabled
           ? absl::GetFlag(FLAGS_output_node_coverage_stats_textproto)
           : std::nullopt);
   std::unique_ptr<ScopedTracingObserver> tracing_observer;
+  std::unique_ptr<CompositeEvaluationObserver> composite_observer;
   std::optional<EvaluationObserver*> observer;
+  std::vector<EvaluationObserver*> observer_ptrs;
```
allight: Unneeded if the CompositeEvaluationObserver can be modified.

xls/tools/eval_ir_main.cc
```diff
@@ -528,20 +528,16 @@ absl::Status Run(Package* package, absl::Span<const ArgSet> arg_sets_in) {
...
-  }
-
   ScopedRecordNodeCoverage cov(
       coverage_enabled ? absl::GetFlag(FLAGS_output_node_coverage_stats_proto)
                        : std::nullopt,
       coverage_enabled
           ? absl::GetFlag(FLAGS_output_node_coverage_stats_textproto)
           : std::nullopt);
   std::unique_ptr<ScopedTracingObserver> tracing_observer;
+  std::unique_ptr<CompositeEvaluationObserver> composite_observer;
```
allight: Make `CompositeEvaluationObserver` modifable as mentioned above then you can just leave this as a stack object and add stuff to it.

xls/tools/eval_proc_main.cc
```diff
@@ -252,31 +249,79 @@ absl::Status LogInterpreterEvents(std::string_view entity_name,
...
-            CHECK(o->observer().has_value())
-                << "The ScopedObserver should only ever hold a "
-                   "ScopedRecordNodeCoverage when it holds an observer.";
-            return *o->observer();
-          }},
-      observer);
-}
+// Helper class that keeps observer lifetimes tied to evaluation routines while
+// allowing callers to register multiple observers at once.
+class ScopedObserver {
```
allight, line 254: This can be massively simplified by just holding the (modifable) CompsiteEvaluationObserver and a vector of std::unique_ptr<EvaluationObserver> to keep things live.

xls/interpreter/observer.h
DeeveshChowdary: Sounds good! I modified it. CompositeEvaluationObserver now exposes AddObserver, RemoveObserver, observers(), and empty() so callsites can treat it as an incrementally fillable container.

xls/tools/eval_ir_main.cc
DeeveshChowdary: Agreed! The composite is now mutable... we just add observers in place and pass a single optional pointer derived from composite.empty(). The composite lives on the stack for the whole Run() invocation.. observers are added directly via the new helpers

xls/tools/eval_proc_main.cc
DeeveshChowdary: I replaced the custom variant/refresh logic with a vector of owning unique_ptrs plus the shared composite, so adding/removing observers is just pushing into those vectors. Coverage pause handling now simply iterates the owned coverage observers.

xls/tools/eval_ir_main.cc
```diff
@@ -554,10 +548,19 @@ absl::Status Run(Package* package, absl::Span<const ArgSet> arg_sets_in) {
...
   }
+  if (coverage_enabled) {
+    if (auto cov_observer = cov.observer(); cov_observer.has_value()) {
+      composite_observer.AddObserver(*cov_observer);
+    }
+  }
+
+  EvaluationObserver* observer =
+      composite_observer.empty() ? nullptr : &composite_observer;
+  std::optional<EvaluationObserver*> observer_opt =
```
allight, line 559: Nit: just do std::optional<...> observer = composite_observer. empty() ? std::nullopt : std::optional<...>(& composite_observer)

allight:
Just one other little cleanup then this looks good.

There aren't really any other missing pieces of the observer system where the missing functionality is a similar level of complexity. The only other major hole in observers is they can't be used with ahead-of-time (aot) compiled code since those don't have the in-memory IR package to tell the obserers what they're looking at. If you want to take a look at it the jit/aot compiler can be found in xls/jit. You'd need to modify the llvm code-generator to support calling the observer callbacks with ids that aren't the node's pointer but the more difficult part is reconstituting these into nodes on the other side. This would be a really big project.

If you're willing to try some z3 stuff https://github.com/google/xls/issues/3293 would be a fun thing to try your hand at.

xls/tools/eval_ir_main.cc
DeeveshChowdary: Done!

DeeveshChowdary:
@allight Thanks! This looks interesting. Please feel free to assign the current issue and #3293 to me.. I'll start working on it.

grebe:
I just want to add on and say thanks for doing this!

allight:
I would also like to say thanks for this. It's great!

We generally use the assignee field for members of the core xls team only. Nobody is working on that at the moment though.

PR #3363, opened 2025-11-12
Add conversion from ast nodes to module member and clone removing member C API

lsrcz:


xls/public/c_api_dslx.cc
```diff
@@ -88,6 +88,23 @@ struct InvocationCalleeDataArray {
   std::vector<xls::dslx::InvocationCalleeData> entries;
 };
 
+template <typename T>
+xls::dslx::ModuleMember* FindModuleMemberForNode(T* node) {
```
richmckeever, line 92: It seems like this would be better as `bool IsModuleMember(const AstNode* node)` The current name kind of implies that it's going to dig up the ModuleMember that the node is in.

richmckeever (approved):

xls/public/c_api_dslx.cc
```diff
@@ -382,13 +399,47 @@ bool xls_dslx_typechecked_module_clone_removing_functions(
     char** error_out, struct xls_dslx_typechecked_module** result_out) {
   CHECK(error_out != nullptr);
   CHECK(result_out != nullptr);
+  auto fail = [&](absl::string_view message) {
```
proppy, line 402: we need to switch this to `std::string_view` for internal checks to pass.

PR #3388, opened 2025-11-16
Enforce assert equivalence in Z3 verifier before claiming passes are safe

DeeveshChowdary:
While looking into https://github.com/google/xls/issues/3293, I realized the Z3 equivalence checker never even sees assert behavior.. it either ignores them or strips them if you pass `--ignore_asserts`. That meant passes could tighten/loosen asserts, or add new ones, and the proof would still come back green because we never asked whether the asserts themselves still fired on the same inputs... I fixed this by making `TryProveEquivalence` pre-process both versions: it now pairs asserts by label/original_label, proves each pair’s condition matches, and forces any unmatched assert to be a tautology. Those checks are ANDed into the predicate we hand to Z3, so `ScopedVerify*` and `check_ir_equivalence_main` pick up the safety net automatically.

Fixes: https://github.com/google/xls/issues/3293

### DIFF
--- a/xls/solvers/z3_ir_equivalence.cc
+++ b/xls/solvers/z3_ir_equivalence.cc
@@ -228,14 +228,14 @@ absl::StatusOr<ProverResult> TryProveEquivalence(Function* a, Function* b,
   }
   XLS_ASSIGN_OR_RETURN(Node * new_ret, CombineChecks(to_test_func, checks));
   XLS_RETURN_IF_ERROR(to_test_func->set_return_value(new_ret));
-  // Remove asserts if requested.
-  if (ignore_asserts) {
-    OptimizationContext ctx;
-    PassResults res;
-    RemoveAssertsPass rap;
-    XLS_RETURN_IF_ERROR(rap.Run(to_test.get(), {}, &res, ctx).status())
-        << "Unable to remove asserts from function!";
-  }
+  // Remove asserts prior to Z3 translation (the solver does not understand
+  // them yet). If assert semantics are being checked, those checks have been
+  // encoded into the return value already.
+  OptimizationContext ctx;
+  PassResults res;
+  RemoveAssertsPass rap;
+  XLS_RETURN_IF_ERROR(rap.Run(to_test.get(), {}, &res, ctx).status())
+      << "Unable to remove asserts from function!";
   // Run prover
   XLS_ASSIGN_OR_RETURN(
       ProverResult base_result,


xls/solvers/z3_ir_equivalence.cc
```diff
@@ -46,6 +49,92 @@ namespace xls::solvers::z3 {
...
+    }
+  }
+
+  std::vector<Node*> checks;
+  checks.reserve(original_asserts.size() + transformed_asserts.size());
+  for (const auto& [label, original_assert] : original_asserts) {
+    auto transformed_it = transformed_asserts.find(label);
+    if (transformed_it != transformed_asserts.end()) {
+      Node* cond_a =
+          original_assert->operand(Assert::kConditionOperand);
```
ericastor: nit: Since `original_assert` is an `Assert*`, you should be able to just write `original_assert->condition()`. Similar below!

xls/solvers/z3_ir_equivalence_test.cc
```diff
@@ -427,5 +427,72 @@ TEST_F(EquivalenceTest, MultiFunctionIgnoresAssert) {
               IsOkAndHolds(IsProvenTrue()));
 }
 
+TEST_F(EquivalenceTest, DetectsMismatchedAssertCondition) {
```
allight, line 430: These would both be significantly easier to understand if they just built 2 independent Function*s instead of using the 'run_pass' version of the TryProveEquivalence function.

xls/solvers/z3_ir_equivalence_test.cc
```diff
@@ -427,5 +427,72 @@ TEST_F(EquivalenceTest, MultiFunctionIgnoresAssert) {
...
+          Node * assert_node,
+          func->MakeNodeWithName<Assert>(SourceInfo(), token, condition,
+                                         "added", "new_label", std::nullopt,
+                                         "new_assert"));
+      (void)assert_node;
+      return absl::OkStatus();
+    };
+  };
+
+  EXPECT_THAT(TryProveEquivalence(f, add_assert(/*tautology=*/false)),
```
allight: 2 different test cases for these please. See: https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

xls/solvers/z3_ir_equivalence.cc
```diff
@@ -46,6 +49,92 @@ namespace xls::solvers::z3 {
...
+      Node* cond_b =
+          transformed_it->second->operand(Assert::kConditionOperand);
+      checks.push_back(f->AddNode(std::make_unique<CompareOp>(
```
allight: We usually add new nodes like

```
XLS_ASSIGN_OR_RETURN(std::back_inserter(checks), f->MakeNode<CompareOp>(SourceInfo(), cond_a, cond_b, Op::Eq));
```

or

```
XLS_ASSIGN_OR_RETURN(std::back_inserter(checks), f->MakeNodeWithName<CompareOp>(SourceInfo(), cond_a, cond_b, Op::Eq, absl::StrCat("assert_", label, "_match")));
```

here and below.

The MakeNode path will also invoke the verifier to do some basic structural checks of the node being added which can catch some bugs.

xls/solvers/z3_ir_equivalence.cc
```diff
@@ -46,6 +49,92 @@ namespace xls::solvers::z3 {
...
+  }
+  for (const auto& [_, transformed_assert] : transformed_asserts) {
+    checks.push_back(
+        transformed_assert->operand(Assert::kConditionOperand));
+  }
+  return checks;
+}
+
+Node* CombineChecks(Function* f, absl::Span<Node* const> checks) {
+  XLS_DCHECK(!checks.empty());
```
allight: There is no XLS_DCHECK.

This should probably just return a StatusOr<Node*> and use XLS_RET_CHECK.

xls/solvers/z3_ir_equivalence.cc
```diff
@@ -46,6 +49,92 @@ namespace xls::solvers::z3 {
...
+  if (checks.size() == 1) {
+    return checks.front();
+  }
+  return f->AddNode(std::make_unique<NaryOp>(
```
allight: As above use MakeNode

xls/solvers/z3_ir_equivalence.cc
DeeveshChowdary: Done. BuildAssertChecks now uses Assert::condition() for both the original and transformed nodes.

xls/solvers/z3_ir_equivalence.cc
DeeveshChowdary: Done!

xls/solvers/z3_ir_equivalence.cc
DeeveshChowdary: Updated. CombineChecks now returns StatusOr<Node*> and uses XLS_RET_CHECK(!checks.empty()).

xls/solvers/z3_ir_equivalence_test.cc
DeeveshChowdary: Done.

xls/solvers/z3_ir_equivalence_test.cc
DeeveshChowdary: Done. I split the coverage into ExtraAssertThatCanFireFails and ExtraAssertThatIsTautologyPasses so each test asserts exactly one behavior.

DeeveshChowdary:
Thanks @ericastor and @allight! I updated the code. Please take another look whenever possible.

> Also the top comment of the PR thread becomes the merge commits message. If you add `Fixes: https://github.com/google/xls/issues/3293` to the end of it then that bug will be automatically closed when merged.

I see.. I've added it to the comment now. Thanks again!

ericastor:
What @allight said - this is great work, and thanks!

PR #3430, opened 2025-11-26
A few small VAST improvements.

meheff:
Specific changes:
* macro statements
* macros with arguments,
* support adding file-scoped blank lines and comments in the C API
* externally defined types with out without package qualifiers. Remove the superfluous representative type from ExternType as well.

PR #3432, opened 2025-11-26
Various VAST C API enhancements

meheff:
Specific changes:
* support emitting an arbitrary expression to string
* expose replicated concatenation
* expose blank line, comments, and inline verilog constructs on statement blocks and modules


meheff:
@scampanoni Thanks, Simone!

There is one more small outstanding PR: https://github.com/google/xls/pull/3430. Thanks!

PR #3597, opened 2025-12-21
Propagate annotated widths into range literals in nested arrays (Re: #3564)

### LINKED ISSUE #3564: bad type inference for range expression
**Describe the bug**

It seems that range expression don't type-coerce into properly annotated typed array but instead redefine the type binding to be the max size of the range.

**To Reproduce**

```
#[test]
fn test_array2d_tiv2() {
    let data: u32[8][2] = [u32:0..u32:8, u32:8..u32:16];
    assert_eq(u32:0 ++ u32:8, data[0][0] ++ data[1][1]);
}
```
```
0313: fn test_array2d_tiv2() {
0314: let data: u32[8][2] = [0..8, 8..16];
0315: assert_eq(u32:0 ++ u32:8, data[0][0] ++ data[1][0]);
~~~~~~~~~~~~~~~~~~~^----------------------------------------^ FailureError: The program being interpreted failed!
lhs: u64:8
rhs: u9:8
were not equal
0316: }
```

**Expected behavior**

I would expect `[0..8, 8..16]` to get coerced into `u32[8][2]`.

Note that annotating the range expression will work around this issue
```
#[test]
fn test_array2d_tiv2() {
    let data: u32[8][2] = [u32:0..u32:8, u32:8..u32:16];
    assert_eq(u32:0 ++ u32:8, data[0][0] ++ data[1][0]);
}
```

DeeveshChowdary:
Hey team! 

While looking into #3564, I tracked the bug down to the DSLX type inference for array literals. A range like 0..8 infers the narrowest bit width needed for its endpoints, and when I wrote `let data: u32[8][2] = [0..8, 8..16];` the outer annotation only enforced the shape, not the inner element width. Because the type checker only propagated the annotated element type into top‑level literal numbers, the `Range` members kept their narrow widths, so data[0][0] wasn’t a u32 and the assertion failed with mismatched widths. 

I fixed this by extending the existing propagation... when an array literal has an annotation, we now also push that annotated element type down into the start and end of any unannotated `Range` members (resolving through one level if the element type itself is an array). With the endpoints coerced to the target bits type, the `Range` deduces to the right element width and the nested array becomes u32[8][2] as annotated. 

I added a regression test mirroring the original repro to lock in the behavior. 

Thanks in advance for the review. Wishing everyone a bug free and relaxing holiday season!

Fixes: [#3564](https://github.com/google/xls/issues/3564)

### DIFF
--- a/xls/dslx/type_system/deduce_expr.cc
+++ b/xls/dslx/type_system/deduce_expr.cc
@@ -191,7 +191,7 @@ static absl::StatusOr<std::unique_ptr<ArrayType>> DeduceArrayInternal(
     // Likewise, propagate the annotated element type to range endpoints so
     // nested arrays of ranges adopt the intended width.
     const Type* target_range_type = &annotated->element_type();
-    if (auto* array_elem = dynamic_cast<const ArrayType*>(target_range_type)) {
+    while (auto* array_elem = dynamic_cast<const ArrayType*>(target_range_type)) {
       target_range_type = &array_elem->element_type();
     }
     if (GetBitsLike(*target_range_type)) {
--- a/xls/dslx/type_system/deduce_expr.cc
+++ b/xls/dslx/type_system/deduce_expr.cc
@@ -188,6 +188,29 @@ static absl::StatusOr<std::unique_ptr<ArrayType>> DeduceArrayInternal(
             NoteTypeForNumber(*number, annotated->element_type(), ctx));
       }
     }
+    // Likewise, propagate the annotated element type to range endpoints so
+    // nested arrays of ranges adopt the intended width.
+    const Type* target_range_type = &annotated->element_type();
+    if (auto* array_elem = dynamic_cast<const ArrayType*>(target_range_type)) {
+      target_range_type = &array_elem->element_type();
+    }
+    if (GetBitsLike(*target_range_type)) {
+      for (Expr* member : node->members()) {
+        auto* range = dynamic_cast<Range*>(member);
+        if (range == nullptr) {
+          continue;
+        }
+        for (Expr* endpoint : {range->start(), range->end()}) {
+          if (auto* number = dynamic_cast<Number*>(endpoint);
+              number != nullptr && number->type_annotation() == nullptr) {
+            XLS_RETURN_IF_ERROR(
+                TryEnsureFitsInType(*number, *target_range_type));
+            XLS_RETURN_IF_ERROR(
+                NoteTypeForNumber(*number, *target_range_type, ctx));
+          }
+        }
+      }
+    }
   }
 
   std::vector<std::unique_ptr<Type>> member_types;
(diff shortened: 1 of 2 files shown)


proppy:
Thanks for tackling this!

xls/dslx/type_system/deduce_expr.cc
```diff
@@ -188,6 +188,29 @@ static absl::StatusOr<std::unique_ptr<ArrayType>> DeduceArrayInternal(
             NoteTypeForNumber(*number, annotated->element_type(), ctx));
       }
     }
+    // Likewise, propagate the annotated element type to range endpoints so
+    // nested arrays of ranges adopt the intended width.
+    const Type* target_range_type = &annotated->element_type();
+    if (auto* array_elem = dynamic_cast<const ArrayType*>(target_range_type)) {
```
proppy: would that also work for array of array of array?

DeeveshChowdary: I changed this to a while loop now. So arrays of arrays of arrays will also coerce range endpoints all the way down to the base bits type.

xls/dslx/type_system/typecheck_module_test.cc
```diff
@@ -3371,13 +3371,12 @@ fn main() {
 
 TEST_F(TypecheckV2Test, RangeAnnotatesNestedArrays) {
   constexpr absl::string_view kProgram = R"(
-#![test]
+#[test]
 fn test_array2d_tiv2() {
```
proppy, line 3375: can we add a test for >2d array as well?

DeeveshChowdary: Done. I added another test for 3d array.

xls/dslx/type_system/deduce_expr.cc
proppy, line 182: @richmckeever should this do something similar and propagate over >1 dimension?

DeeveshChowdary: Hey everyone! Sorry I'm late. Checking-in again on this thread. Hope everyone had a great vacation ( and a pleasant next two months ). 

I wanted to know if this issue has already been resolved or if it is still relevant now? If not, we can close this PR and I can probably start working on some other issue :)
Please let me know.

cc: @richmckeever @proppy

richmckeever: I think it's already resolved with the switch to type inference v2. deduce_expr.cc is gone now, but it's probably worth committing your test, and it should pass now.

proppy:
@DeeveshChowdary did you get a chance to look at @richmckeever's suggestion?

proppy:
marking this as draft, @DeeveshChowdary feel free to re-mark it as active it if you still want to contribute the test as @richmckeever suggested.

PR #3626, opened 2025-12-30
Add `const match`

wsipak:
This PR adds support for the constexpr `match` expression.

This is work in progress.
The aim of this PR is to add support for `const match` expressions, where the arm selection is performed in compilation time.
* [x] allow parsing `const match`, as a `Match` node with a boolean property indicating the `const`.
* [x] evaluate the matched value and enforce that it must be possible to evaluate in compilation time for `const match`
* [x] select the right arm for a `const match` in compilation time
* [x] test DSLX and IR
* [x] allow disparate types in the arms of a match / do not typecheck the unused arms 
* [x] test generating SystemVerilog 
* [x] test the formatter

### DIFF
--- a/xls/dslx/frontend/ast_utils_test.cc
+++ b/xls/dslx/frontend/ast_utils_test.cc
@@ -173,56 +173,53 @@ fn f() -> u32 {
   auto* match = absl::down_cast<Match*>(match_expr);
   XLS_ASSERT_OK_AND_ASSIGN(std::vector<AstNode*> nodes,
                            CollectUnder(match, /*want_types=*/false));
-  ASSERT_EQ(nodes.size(), 16);
+  ASSERT_EQ(nodes.size(), 15);
 
   EXPECT_EQ(nodes[0]->ToString(), "t");
   EXPECT_EQ(nodes[0]->GetNodeTypeName(), "NameRef");
 
-  EXPECT_EQ(nodes[1]->ToString(), "0");
-  EXPECT_EQ(nodes[1]->GetNodeTypeName(), "Number");
+  EXPECT_EQ(nodes[1]->ToString(), "x");
+  EXPECT_EQ(nodes[1]->GetNodeTypeName(), "NameDef");
 
   EXPECT_EQ(nodes[2]->ToString(), "x");
-  EXPECT_EQ(nodes[2]->GetNodeTypeName(), "NameDef");
+  EXPECT_EQ(nodes[2]->GetNodeTypeName(), "NameDefTree");
 
-  EXPECT_EQ(nodes[3]->ToString(), "x");
-  EXPECT_EQ(nodes[3]->GetNodeTypeName(), "NameDefTree");
+  EXPECT_EQ(nodes[3]->ToString(), "u32");
+  EXPECT_EQ(nodes[3]->GetNodeTypeName(), "BuiltinTypeAnnotation");
 
-  EXPECT_EQ(nodes[4]->ToString(), "u32");
-  EXPECT_EQ(nodes[4]->GetNodeTypeName(), "BuiltinTypeAnnotation");
+  EXPECT_EQ(nodes[4]->ToString(), "u32:0");
+  EXPECT_EQ(nodes[4]->GetNodeTypeName(), "Number");
 
-  EXPECT_EQ(nodes[5]->ToString(), "u32:0");
-  EXPECT_EQ(nodes[5]->GetNodeTypeName(), "Number");
+  EXPECT_EQ(nodes[5]->ToString(), "x => u32:0");
+  EXPECT_EQ(nodes[5]->GetNodeTypeName(), "MatchArm");
 
-  EXPECT_EQ(nodes[6]->ToString(), "x => u32:0");
-  EXPECT_EQ(nodes[6]->GetNodeTypeName(), "MatchArm");
+  EXPECT_EQ(nodes[6]->ToString(), "y");
+  EXPECT_EQ(nodes[6]->GetNodeTypeName(), "NameDef");
 
   EXPECT_EQ(nodes[7]->ToString(), "y");
-  EXPECT_EQ(nodes[7]->GetNodeTypeName(), "NameDef");
+  EXPECT_EQ(nodes[7]->GetNodeTypeName(), "NameDefTree");
 
-  EXPECT_EQ(nodes[8]->ToString(), "y");
-  EXPECT_EQ(nodes[8]->GetNodeTypeName(), "NameDefTree");
+  EXPECT_EQ(nodes[8]->ToString(), "z");
+  EXPECT_EQ(nodes[8]->GetNodeTypeName(), "NameDef");
 
   EXPECT_EQ(nodes[9]->ToString(), "z");
-  EXPECT_EQ(nodes[9]->GetNodeTypeName(), "NameDef");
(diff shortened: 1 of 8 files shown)


magancarz:
I've added formatting `const match` and rebased changes on main.

magancarz:
I've added handling different types for arms, fixed spawning procs using `const match`, changed emitting bytecode to only use selected branch and added more DSLX/C++ test cases.

magancarz:
@richmckeever is there anything else required to add?

magancarz:
I've added proper handling of `const match` with proc-scoped channels by handling only the selected arm and included a fix from #3745 which allows proc-scoped channels tests with parametrized procs to pass.

xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -544,6 +544,27 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
     return DefaultHandler(node);
   }
 
+  absl::Status HandleConstMatch(const Match* node) {
+    VLOG(5) << "HandleConstMatch: " << node->ToString();
+    std::vector<TypeAnnotation*> members;
+
+    for (MatchArm* arm : node->arms()) {
+      XLS_RETURN_IF_ERROR(
```
richmckeever: I think it would be better to merge this logic into the for loop that is in HandleMatch. i.e. turn the body of that loop into "if (node->IsConst() { ...  } else { ... }" perhaps with some common logic at the beginning or end of the loop body. This would be similar to how HandleConditional works.

We generally don't use InferenceTable in such a way that we configure a node the normal way and then reconfigure it if we are in some special case. Even if it happens to work, that is a brittle approach because the normal way could get more complicated later and start doing stuff you don't want for the const case.

xls/dslx/frontend/ast.h
```diff
@@ -2707,16 +2744,21 @@ class Match : public Expr {
 
   const std::vector<MatchArm*>& arms() const { return arms_; }
   Expr* matched() const { return matched_; }
+  Number* arm_idx() const { return arm_idx_; }
```
richmckeever: I think it would be better to model the notion of the chosen arm with dedicated TypeInfo API functions, i.e. add the ability to say
type_info->SetConstActiveArmIndex(match_node, 3)

as opposed to modeling this as a sentinel AST node with a constexpr value. The reason not to use the AST node is that it's is always a bogus zero value, and using that directly would be an easy mistake.

xls/dslx/type_system_v2/type_annotation_resolver.cc
```diff
@@ -537,6 +541,45 @@ class StatefulResolver : public TypeAnnotationResolver {
...
+      ti = (*parametric_context)->type_info();
+    } else {
+      XLS_ASSIGN_OR_RETURN(
+          ti,
+          import_data_.GetRootTypeInfoForNode(match));
+    }
+
+    XLS_ASSIGN_OR_RETURN(
+        InterpValue interp_match,
+        evaluator_.ConstMatchWhichArm(parametric_context, ti, match));
```
richmckeever: I think if we move the arm selection to CollectConstants (see other comment), we will be able to just ask TypeInfo for the chosen arm index here, Evaluator will not need a function for it, and we will not need to set anything in TI further down.

xls/dslx/constexpr_evaluator.cc
```diff
@@ -126,6 +126,46 @@ std::string CallStackToString(std::string_view root_span_string,
   return NotConstantErrorStatus(expr->span(), expr, file_table);
 }
 
+/* static */ absl::StatusOr<InterpValue> ConstexprEvaluator::ConstMatchWhichArm(
```
richmckeever: It would be more in line with the general data flow in type inference if we could put this logic in CollectConstants() because that function basically runs as soon as it's feasible to decide the constant(s) associated with some node. It relieves ConstexprEvaluator of type inference-related knowledge Ideally ConstexprEvaluator just farms math out to the bytecode interpreter and has minimal inference logic.

Perhaps in CollectConstants, we say after dispatching the node to the visitor: if node->parent() is a Match and node == match->matched(), do this logic and then set the chosen arm index in TypeInfo. It could be a helper function inside that file.

magancarz:
Thank you for the review! I've applied your suggestions, rebased the branch onto latest main and fixed few failing tests after the rebase.

I've also noticed `GetConstExpr` breaks inside parametric functions when converting DSLX to IR with proc-scoped channels enabled and function as a top. I think it is a case similar to the issued described in #3745.

magancarz:
I rebased changes on the latest main.

xls/dslx/type_system/type_info.h
```diff
@@ -424,6 +424,10 @@ class TypeInfo {
   // environments.
   std::vector<Expr*> GetAllUnrolledLoops(const ConstFor* loop) const;
 
+  // Stores the result index of `const match` arm selection evaluation.
+  void NoteArmSelectionResult(const Match* expr, InterpValue value);
```
richmckeever: It might be better to make the value an int64_t (or some flavor of int) here. Is there any case where having it be an InterpValue helps?

xls/dslx/type_system_v2/constant_collector.cc
```diff
@@ -721,10 +721,68 @@ class ConstantCollectorImpl : public ConstantCollector {
                     file_table_, converter_, evaluator_,
                     parametric_struct_instantiator_, tracer_,
                     parametric_context, type, ti, std::move(trace));
-    return node->Accept(&visitor);
+    XLS_RETURN_IF_ERROR(node->Accept(&visitor));
+
+    if (auto match = dynamic_cast<const Match*>(node->parent());
```
richmckeever: We should check if node->parent() is nullptr first, and use a kind() check followed by down_cast() instead of dynamic_cast. Even though this is more cumbersome, we have been avoiding dynamic_cast in type inference due to poor performance.

xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -611,6 +612,28 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
               table_.SetTypeVariable(ToAstNode(pattern->leaf()), matched_var));
         }
       }
+
+      if (node->IsConst()) {
+        XLS_RETURN_IF_ERROR(
```
richmckeever, line 608: Let's put the setting of the type variable on arm->expr() in an if/else  with the one above (line 607) so that it only sets one variable or the other.

xls/dslx/type_system_v2/constant_collector.cc
magancarz: I've removed the usage of `dynamic_cast` and changed it to the described way.

xls/dslx/type_system/type_info.h
magancarz: `InterpValue` is unnecessary. I've changed it to store `uint32_t`.

xls/dslx/type_system_v2/populate_table_visitor.cc
magancarz: I've applied requested change.

magancarz:
Thank you for the re-review! I've applied your suggestions.

magancarz:
I've rebased changes onto the latest main.

@richmckeever is there anything else required to add?

PR #3627, opened 2025-12-30
[opt] Neg canonicalization, abs equals zero optimization.

cdleary:
* Canonicalize `add(not(x), 1) => neg(x)`
* Optimize `eq(sel(p, cases=[x, neg(x)]), 0) => eq(x, 0)` (for nez as well)

### DIFF
--- a/xls/passes/arith_simplification_pass.cc
+++ b/xls/passes/arith_simplification_pass.cc
@@ -976,6 +976,27 @@ absl::StatusOr<bool> MatchSignedDivide(Node* original_div_op,
 // different expression).
 absl::StatusOr<bool> MatchArithPatterns(int64_t opt_level, Node* n,
                                         const QueryEngine& query_engine) {
+  // Pattern: Add(Not(x), 1) => Neg(x)
+  //
+  // In two's complement arithmetic: ~x + 1 == -x (mod 2^N).
+  if (n->op() == Op::kAdd) {
+    Node* negated = nullptr;
+    if (n->operand(0)->op() == Op::kNot &&
+        IsLiteralUnsignedOne(n->operand(1))) {
+      negated = n->operand(0)->operand(0);
+    } else if (n->operand(1)->op() == Op::kNot &&
+               IsLiteralUnsignedOne(n->operand(0))) {
+      // Handle the commuted operand ordering as well.
+      negated = n->operand(1)->operand(0);
+    }
+    if (negated != nullptr) {
+      VLOG(2) << "FOUND: Add(Not(x), 1)";
+      XLS_RETURN_IF_ERROR(
+          n->ReplaceUsesWithNew<UnOp>(negated, Op::kNeg).status());
+      return true;
+    }
+  }
+
   // Pattern: UDiv/UMul/SMul by a positive power of two.
   if ((n->op() == Op::kSMul || n->op() == Op::kUMul || n->op() == Op::kUDiv) &&
       query_engine.IsFullyKnown(n->operand(1))) {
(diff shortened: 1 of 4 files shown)


ericastor:
BTW: we've started to think it would be helpful to just add an absolute-value unary operation to XLS IR, so we can canonicalize to it & model it more easily for various optimizations and analyses... without relying on detecting the exact pattern `sel(x < 0, cases=[x, neg(x)])`. What do you think about taking this step in general?

cdleary:
@ericastor Sounds totally reasonable and definitely like something that's worth us experimenting with! Also generally a pretty type-2 decision since we could always desugar it at some early point in the pipeline if we wanted to.

I think there's a natural tension between having fewer primitives (which you'd pick because then because the op representations are more universally present in the IR) vs having more clear composite operations because their aggregate properties are particularly interesting. E.g. for canonicalizing to neg I was wondering if then we'd miss any patterns that were smashing together add(x, $N) on account of negate "laundering" that inside.

But it never seems wrong to have a higher level primitive the user can opt to provide the system that "worst case it can do is lower".

xls/passes/arith_simplification_pass.cc
```diff
@@ -976,6 +976,36 @@ absl::StatusOr<bool> MatchSignedDivide(Node* original_div_op,
 // different expression).
 absl::StatusOr<bool> MatchArithPatterns(int64_t opt_level, Node* n,
                                         const QueryEngine& query_engine) {
+  // Pattern: Add(Not(x), 1) => Neg(x)
+  //
+  // In two's complement arithmetic: ~x + 1 == -x (mod 2^N).
+  if (n->op() == Op::kAdd) {
+    auto is_ones_complement_plus_one_literal = [&](Node* node) -> bool {
```
ericastor: Shouldn't this just be `is_literal_one`? There's nothing this checks about ones' complement logic.

xls/passes/arith_simplification_pass.cc
```diff
@@ -976,6 +976,36 @@ absl::StatusOr<bool> MatchSignedDivide(Node* original_div_op,
...
+      if (!node->GetType()->IsBits()) {
+        return false;
+      }
+      std::optional<Bits> bits = query_engine.KnownValueAsBits(node);
+      return bits.has_value() && bits->bit_count() == n->BitCountOrDie() &&
```
ericastor: I don't *think* we need to check if the bit count matches, since addition only accepts two equal-sized values; otherwise, the IR is invalid. Also, I don't think the logic here cares whether it matches?

xls/passes/arith_simplification_pass.cc
```diff
@@ -1771,6 +1801,74 @@ absl::StatusOr<bool> MatchArithPatterns(int64_t opt_level, Node* n,
...
+        return true;
+      }
+    }
+  }
+
+  // Pattern: Comparison of not (ones-complement negation) to literal
+  //   eq(~expr, K) => eq(expr, ~K)
+  //   ne(~expr, K) => ne(expr, ~K)
+  //
+  // This is particularly useful as a follow-on simplification after
```
ericastor: Don't we have another simplification that canonicalizes `cmp(neg(x), K)` to `rev_cmp(x, -K)`? If so, this *should* be redundant... can we try to see why that's not applying, now that we're aggressively canonicalizing `add(~x, 1)` to `neg(x)`?

xls/passes/arith_simplification_pass_test.cc
```diff
@@ -138,6 +138,49 @@ TEST_F(ArithSimplificationPassTest, CompareEqNegatedConstant) {
   EXPECT_THAT(f->return_value(), m::Eq(m::Param("x"), m::Literal(253)));
 }
 
+TEST_F(ArithSimplificationPassTest, CompareEqTwosComplementNegatedConstant) {
```
ericastor, line 141: We've been trying to move away from parsed-IR tests to IR-builder-based construction; it makes test maintenance easier in the long run. Would you mind converting these new tests to use a FunctionBuilder?

cdleary: Done.

xls/passes/arith_simplification_pass.cc
cdleary: Done.

xls/passes/arith_simplification_pass.cc
cdleary: Right, removed that check 👍

xls/passes/arith_simplification_pass.cc
cdleary: Note this block is for Op::kNot, not Op::kNeg (maybe I should spell out not() here instead of using the tilde). Even with `add(~x,1) -> neg(x)`, we still need `eq/ne(~expr, K) -> eq/ne(expr, ~K)` because `~x` is not the same as `-x` without the `+1`.

Example: `eq(not(x), 3)` should fold to `eq(x, 252)` (8-bit); the `cmp(neg(x), K)` rewrite can’t apply since there’s no `neg` node.

LMK if that makes sense.

ericastor: That makes sense... but the comment specifically says it's useful for after:
```
// MatchComparisonOfInjectiveOp rewrites:
//   eq(add(~x, 1), C) => eq(~x, C-1) => eq(x, ~(C-1))
```
which should never happen anymore, since we canonicalize `add(~x, 1)` to `neg(x)` first. Maybe I'm just objecting to the comment, since I don't think that'll come up?

cdleary: Ah I see, good point, removed that comment.

cdleary: (Notably that was something I was observing early when I had added one pattern but not the other yet and ran regression tests, so seems like it all lines up.)

ericastor: Thanks! If we want to make one last tweak before we land... this *is* a comparison of an injective operation against a constant, which suggests we could add it to MatchComparisonOfInjectiveOp rather than putting it here. But I don't think we strictly need to do that, it'd just be for consistency/cleanliness.

PR #3638, opened 2025-12-31
[opt] Optimize one-bit logical shifts.

cdleary:
(shra was already handled but shrl and shll were not)


ericastor:
Do we know where Shra is handled? It seems like all three should ideally be handled in the same place.

cdleary:
@ericastor I moved it so it's right above the cited opt now

PR #3639, opened 2026-01-01
[opt] Simplify `or_reduce(neg(x)) => or_reduce(x)`

cdleary:
This is a strict local improvement (`or_reduce(neg(x))` depends on `x` instead of `neg(x)`)

However, it also highlights an existing representational redundancy: for bits values, `or_reduce(x)` and `ne(x, 0)` both mean `x != 0` and we don’t currently canonicalize between them. We may want to decide on a canonical nez form, or make some query helpers that recognize both as equivalent representations, but I think best done in a follow-up after some discussion, rather than as part of this change.

### DIFF
--- a/xls/passes/basic_simplification_pass_test.cc
+++ b/xls/passes/basic_simplification_pass_test.cc
@@ -359,6 +359,37 @@ TEST_F(BasicSimplificationPassTest, SingleBitOrReduce) {
   EXPECT_THAT(f->return_value(), m::Param("x"));
 }
 
+TEST_F(BasicSimplificationPassTest, OrReduceOfNeg) {
+  auto p = CreatePackage();
+  FunctionBuilder fb(TestName(), p.get());
+  Type* t = p->GetBitsType(32);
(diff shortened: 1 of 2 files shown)


ericastor (approved):
Agreed that we should follow up by looking into canonicalizing the two forms of not-equal-to-zero checks.

xls/passes/basic_simplification_pass_test.cc
```diff
@@ -359,6 +359,37 @@ TEST_F(BasicSimplificationPassTest, SingleBitOrReduce) {
   EXPECT_THAT(f->return_value(), m::Param("x"));
 }
 
+TEST_F(BasicSimplificationPassTest, OrReduceOfNeg) {
```
angelomatni1, line 362: Small nit, otherwise the CL looks good to me: This unit test feels like it doesn't add much value, because `OrReduceOfNegWithOtherUses` is negligibly more complicated and is a superset of this.

PR #3640, opened 2026-01-02
[opt] Optimize commutative op working on (explicitly) commuted operands.

cdleary:
`op(sel(p, cases=[a, b]), sel(p, cases=[b, a]))` can be optimized to `op(a, b)` when `op` is commutative.

- Add MatchBinarySelectLike() (and BinarySelectArms) in node_util to normalize binary sel and binary priority_sel into {selector, on_false, on_true}.
- Use that helper in BasicSimplificationPass to fold commutative ops over swapped binary selects:
`X(sel(p,[a,b]), sel(p,[b,a])) -> X(a,b)` (applies to any 2-operand commutative, non-side-effecting op).

`select_simplification_pass` has several 2-way-select peepholes (case swapping / collapsing nested selects / detecting binary select patterns), but not this "commutative op of swapped cases collapses" rule.

### DIFF
--- a/xls/passes/arith_simplification_pass.cc
+++ b/xls/passes/arith_simplification_pass.cc
@@ -1823,7 +1823,7 @@ absl::StatusOr<bool> MatchArithPatterns(int64_t opt_level, Node* n,
       selector_node = n->operand(1);
       zero_node = n->operand(0);
     }
-    if (selector_node != nullptr && IsBinarySelect(selector_node)) {
+    if (selector_node != nullptr && IsBinarySelectTwoCases(selector_node)) {
       Select* sel = selector_node->As<Select>();
       Node* case0 = sel->get_case(0);
       Node* case1 = sel->get_case(1);


ericastor:
This seems good to me... but continues to suggest that we may need better pass organization, since I'm not sure this should qualify as "basic".

ericastor (approved):

xls/ir/node_util.h
```diff
@@ -417,6 +417,23 @@ bool IsBinarySelect(Node* node);
 // default value)
 bool IsBinaryPrioritySelect(Node* node);
 
+// A uniform view of a "binary select-like" node.
+//
+// For `sel(p, cases=[on_false, on_true])`, the selector is `p`.
+// For a binary `priority_sel(p, cases=[on_true], default=on_false)`, the
+// selector is `p` (which is required to be a single bit).
+struct BinarySelectArms {
```
angelomatni1: Maybe rename to BinarySelect since this includes both the arms and the selector?

xls/ir/node_util.cc
```diff
@@ -908,6 +908,25 @@ bool IsBinaryPrioritySelect(Node* node) {
...
 }
 
+absl::StatusOr<std::optional<BinarySelectArms>> MatchBinarySelectLike(
+    Node* node) {
+  if (IsBinarySelect(node)) {
+    Select* sel = node->As<Select>();
+    XLS_RET_CHECK_EQ(sel->selector()->BitCountOrDie(), 1);
+    return BinarySelectArms{.selector = sel->selector(),
+                            .on_false = sel->get_case(0),
+                            .on_true = sel->get_case(1)};
```
angelomatni1, line 984: Is it possible for the select to have a single case and the default value to cover the case for selector == 0b1?

xls/ir/node_util_test.cc
```diff
@@ -177,6 +177,54 @@ TEST_F(NodeUtilTest, GatherAllTheBits) {
   EXPECT_THAT(f->return_value(), m::Param("x"));
 }
 
+TEST_F(NodeUtilTest, MatchBinarySelectLikeSelect) {
+  auto p = CreatePackage();
+  FunctionBuilder fb(TestName(), p.get());
+  BValue selector = fb.Param("p", p->GetBitsType(1));
+  BValue a = fb.Param("a", p->GetBitsType(32));
+  BValue b = fb.Param("b", p->GetBitsType(32));
+  BValue sel = fb.Select(selector, {a, b});
```
angelomatni1, line 187: Unused variable `sel`; same for the below 2 unit tests

xls/passes/basic_simplification_pass_test.cc
```diff
@@ -501,6 +501,54 @@ TEST_F(BasicSimplificationPassTest, XorDuplicateOperandsCollapseToXor) {
...
+  BValue sel_ba = fb.Select(p_sel, {b, a});
+  fb.Eq(sel_ab, sel_ba);
+  XLS_ASSERT_OK_AND_ASSIGN(Function * f, fb.Build());
+
+  ScopedVerifyEquivalence sve(f);
+  ASSERT_THAT(Run(p.get()), IsOkAndHolds(true));
+  EXPECT_THAT(f->return_value(), m::Eq(m::Param("a"), m::Param("b")));
+}
+
 TEST_F(BasicSimplificationPassTest, AddWithZero) {
```
angelomatni1, line 696: Can we add a unit test on a function that is non-commutative?

xls/ir/node_util.h
cdleary: Good point, renamed to BinarySelectView, since it's extracting some lightweight stuff that describes the underlying node, though that made sense.

xls/ir/node_util_test.cc
cdleary: Switched to BuildWithReturnValue to be more explicit and avoid the unused BValue

xls/passes/basic_simplification_pass_test.cc
cdleary: Good idea, added for Sub

xls/ir/node_util.cc
cdleary: Good eye -- I think we're starting to feel the lack of a strong canonical form here.

I switched IsBinarySelect to be named IsBinarySelectTwoCases to better document that we're missing the one-case-with-default if we only handle that as a predicate. You can see some other places popped up as "maybe they were missing the one-case form there as well"

cdleary:
@ericastor Because PRs don't have a natural/easy way to stack I believe this PR conflicted with an earlier one I had put up that got merged! At first I thought it was something downstream but it's just that downstream reflected the earlier merge.

PTAL, I rebased and updated this PR.

ericastor (approved):

PR #3670, opened 2026-01-10
[opt:ra] Add range analysis support for Encode op.

cdleary:
Takes a whack at range analysis for the encode IR op.

Fast exact case (small input range): If the input’s IntervalSet contains only a small number of concrete values (<= kMaxIterationSize), we enumerate each possible input value, compute encode result exactly, and union the results. This produces an exact output set and is still cheap because the input set is small by construction.

Fallback (large input range): If the input range is large, we avoid enumerating it. Instead, we extract a ternary mask for the input bits (known-0 / known-1 / unknown) from the input interval set, and compute each output bit using the encode op definition ("output bit j is the OR of input bits whose index has bit j set"). If any input bit is known-1 the output bit is known-1; if none are known-1 but at least one is unknown the output bit is unknown, otherwise it’s known-0. We convert that output ternary mask into an IntervalSet (interval_ops::FromTernary), which is conservative but fast.

Fuzzed multiprocess for >15 mins on a many-core machine.

### DIFF
--- a/xls/passes/range_query_engine.cc
+++ b/xls/passes/range_query_engine.cc
@@ -117,7 +117,6 @@ class RangeQueryVisitor : public DfsVisitor {
 
   // The maximum number of points covered by an interval set that can be
   // iterated over in an analysis.
-  static constexpr int64_t kMaxIterationSize = 1024;
 
   // Wrapper around GetIntervalSetTree for consistency with the
   // SetIntervalSetTree wrapper.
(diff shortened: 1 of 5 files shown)


xls/passes/range_query_engine.cc
```diff
@@ -528,7 +528,76 @@ absl::Status RangeQueryVisitor::HandleDynamicCountedFor(
...
+  XLS_RET_CHECK(arg.IsNormalized());
+
+  const int64_t input_width = encode->operand(0)->BitCountOrDie();
+  const int64_t output_width = encode->BitCountOrDie();
+
+  // For small ranges, compute the exact output by enumerating all possible
+  // input values. This keeps Encode precise when the input range is small, and
+  // avoids blowing up for large ranges.
+  if (std::optional<int64_t> sz = arg.Size();
+      sz.has_value() && *sz <= kMaxIterationSize) {
```
allight: 1024 possible bit values seems like a lot to push into this path. Especially since IntervalSet generally avoids having anywhere near that number of splits (generally sticks to ~25 and crushes down to 5 for most operations). A much smaller limit may be appropriate.

xls/passes/range_query_engine.cc
```diff
@@ -528,7 +528,76 @@ absl::Status RangeQueryVisitor::HandleDynamicCountedFor(
 
 absl::Status RangeQueryVisitor::HandleEncode(Encode* encode) {
```
allight, line 528: Please put the actual logic (and a test) in interval_ops.cc like the rest of the arithmetic functions here. That will also let you use the much faster interval-fuzz test directly to check for errors (though the implementation looks good to me FWIW)

cdleary: Makes sense! Think I did it the right way now.

xls/passes/range_query_engine.cc
cdleary: Makes sense, thanks for the concrete guidance -- I reused the kConstant that was in interval_ops.cc for a similar situation which made it 16 now.

PR #3672, opened 2026-01-11
[opt] Add strength reduction for `sgtz(x) -> and(not(msb(x)), nez(x))` and `sgt(x, 2^k-1)`

cdleary:
Also ran multiprocess fuzz for 15 mins on a manycore box

### DIFF
--- a/xls/contrib/xlscc/integration_tests/artifact_analyzability_test.py
+++ b/xls/contrib/xlscc/integration_tests/artifact_analyzability_test.py
@@ -169,7 +169,7 @@ class ArtifactAnalyzabilityTest(absltest.TestCase):
         input_interface, for_variable="i", bit_count=32
     )
     self._assert_loop_variable_bit_width(
-        output_interface, for_variable="i", bit_count=2
+        output_interface, for_variable="i", bit_count=3
     )
 
 


xls/contrib/xlscc/integration_tests/artifact_analyzability_test.py
```diff
@@ -169,7 +169,7 @@ def test_backward_pipelined_loop_is_narrowed(self):
         input_interface, for_variable="i", bit_count=32
     )
     self._assert_loop_variable_bit_width(
-        output_interface, for_variable="i", bit_count=2
+        output_interface, for_variable="i", bit_count=3
```
allight, line 172: This shows this transform inhibits analyzability enough that I'm not sure we want to do it as early as arith_simp performs it.

I'd honestly say both of the transforms added here should be delayed until at least NarrowingEnabled and probably until SplitsEnabled (opt_level == 3).

cdleary: Going to close this PR based on our discussion in the other PR

I do think we need some better indicator that changes like this one are not workable in the OSS code (more than "a random looking thing happens in one fragile seeming xlscc test"), do you have thoughts on how we could accomplish that so ecosystem contributions can understand whether something will help vs hurt @allight  ?

PR #3685, opened 2026-01-14
Add support for array-typed parameters to VAST.

meheff:
These were representable but were unusable because they could not be converted to a IndexableExpression. Also add unpacked array parameters and array assignment to the C API.

### DIFF
--- a/xls/codegen/vast/vast_test.cc
+++ b/xls/codegen/vast/vast_test.cc
@@ -2134,8 +2134,6 @@ TEST_P(VastTest, ArrayParameters) {
   Module* m = f.AddModule("top", SourceInfo());
 
   const SourceInfo si;
-  // Literals used for the array assignment patterns below.
-  Expression* tick0 = f.Make<UnsizedZeroLiteral>(si);
 
   Def* p1_def =
       f.Make<Def>(si, "P1", DataKind::kInt,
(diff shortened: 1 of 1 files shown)


xls/codegen/vast/vast_test.cc
```diff
@@ -2126,6 +2126,79 @@ TEST_P(VastTest, ArrayAssignmentPattern) {
...
+      {f.Literal(0x45, 8, si), f.Literal(0x46, 8, si), f.Literal(0x47, 8, si)},
+      si);
+  Def* p5_def =
+      f.Make<Def>(si, "P5", DataKind::kLogic,
+                  f.UnpackedArrayType(/*element_bit_count=*/8, {2, 3}, si));
+  m->AddParameter(p5_def, f.ArrayAssignmentPattern({p5_row0, p5_row1}, si), si);
+
+  EXPECT_EQ(m->Emit(nullptr),
+            R"(module top;
+  parameter P0[2] = '{'0, '0};
```
mikex-oss: I don't believe it's valid SystemVerilog to have an unpacked param without an explicit data type. Similar below.

meheff: Thanks for the review. Sorry I missed it earlier.

There is not a specific example of this construct in the spec, but the grammar production rules allow it (6.20.1) and it doesn't appear to be explicitly disallowed. There is some discussion about aggregate constants at the end of 6.20.2 where such an exception seemingly would be made. I assume it behaves just as an array of the implicit data type (32-bit signed logic in SV).

mikex-oss: Both Verific and VCS report syntax errors in this case.  E.g.

```
Error-[SE] Syntax error
  Following verilog source has syntax error :
...
token is '='
    parameter P0[2] = '{'0, '0};
```

meheff: Sounds good. I removed them from the tests. I didn't add a check as checking is awkward as it would require deep inspection of the datatype and VAST generally does not do extensive checking of the AST for legality, just trivial checks.

ericastor:
On import, we're seeing unused variable warnings (which we treat as errors):

```
xls/codegen/vast/vast_test.cc:2138:15: error: unused variable 'tick0' [-Werror,-Wunused-variable]
 2138 |   Expression* tick0 = f.Make<UnsizedZeroLiteral>(si);
      |               ^~~~~
xls/codegen/vast/vast_test.cc:2157:15: error: unused variable 'p3_row' [-Werror,-Wunused-variable]
 2157 |   Expression* p3_row =
      |               ^~~~~~
```

Mind fixing this, or should we do it at point-of-import?

meheff:
This should be fixed now. Thanks, Eric!

PR #3689, opened 2026-01-15
[opt] Hoist comparisons across selects when it is cheap to do so

cdleary:
Add a SelectSimplificationPass rewrite to push CompareOp through sel / priority_sel / one_hot_sel when comparing against a literal and the select-like node has ≤1 non-literal arm (including default) plus a single-user guard.

e.g.

```
eq(priority_sel(s, cases=[a, b, c], default=d), t)
  -> priority_sel(s,
                  cases=[eq(a, t), eq(b, t), eq(c, t)],
                  default=eq(d, t))
```

The fact we only do it when (at most) one case is non-constant means it should be more area-preserving to push this through. (It does suggest maybe we want some knob like --speculation_level=N similar to how we have --opt_level=N in the future though.)

For one_hot_sel, apply only when the selector is provably exactly-one-hot (IsKnownOneHotSelector), including common “one-hot by construction” patterns.

Extend GenericSelect with a small helper to rebuild the same select-like op with new arms (MakeSelectLikeWithNewArms).

Add query_engine_utils (+ tests) and new select_simplification_pass tests covering prio/sel/onehot, non-commutative operand ordering, and all-literal arms.

### DIFF
--- a/xls/passes/select_simplification_pass.cc
+++ b/xls/passes/select_simplification_pass.cc
@@ -1202,9 +1202,115 @@ absl::StatusOr<bool> MaybeReorderSelect(Node* node,
   return true;
 }
 
+absl::StatusOr<bool> TryHoistCompareThroughSelectLike(
+    Node* node, const QueryEngine& query_engine) {
+  if (!node->Is<CompareOp>()) {
+    return false;
+  }
+  CompareOp* cmp = node->As<CompareOp>();
+
+  std::optional<GenericSelect> sel;
+  Node* other = nullptr;
+  bool select_like_is_lhs = false;
+  // Look for `cmp(sel_like, literal)` or `cmp(literal, sel_like)` where
+  // `sel_like` is any select-like node wrapped by `GenericSelect`. We keep
+  // track of which side the select-like node is on so we preserve operand
+  // ordering for non-commutative compares (e.g. ULT/UGT).
+  if (auto gs = GenericSelect::From(cmp->operand(0)); gs.ok()) {
+    sel = *gs;
+    other = cmp->operand(1);
+    select_like_is_lhs = true;
+  } else if (auto gs = GenericSelect::From(cmp->operand(1)); gs.ok()) {
+    sel = *gs;
+    other = cmp->operand(0);
+    select_like_is_lhs = false;
+  } else {
+    return false;
+  }
+  XLS_RET_CHECK(sel.has_value() && sel->valid());
+  if (sel->kind() == GenericSelect::Kind::kOneHotSel &&
+      !query_engine.ExactlyOneBitTrue(sel->selector())) {
+    // For `one_hot_sel`, hoisting the compare into the arms is only semantics-
+    // preserving when the selector is exactly-one-hot. Otherwise, `one_hot_sel`
+    // has "multiple active arms" behavior which is not equivalent to selecting
+    // exactly one arm and then comparing.
+    return false;
+  }
+
+  // Guardrails:
+  //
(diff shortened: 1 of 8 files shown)


xls/passes/select_simplification_pass.cc
```diff
@@ -1202,9 +1202,114 @@ absl::StatusOr<bool> MaybeReorderSelect(Node* node,
   return true;
 }
 
+absl::StatusOr<bool> TryPushCompareThroughSelectLike(
```
meheff: Naming nit "hoist" rather than "push"?

xls/passes/select_simplification_pass.cc
```diff
@@ -1202,9 +1202,114 @@ absl::StatusOr<bool> MaybeReorderSelect(Node* node,
...
+
+  // Guardrails:
+  //
+  // - Single explicit user: avoids duplicating the select tree under multiple
+  //   compares/users (this rewrite clones the compare into each arm).
+  // - Literal other operand: maximizes folding opportunities after pushing the
+  //   comparison through the select(-like) into the arms.
+  //   This keeps the rewrite roughly area-preserving in the intended cases:
+  //   all-but-(at most)-one arm becomes a constant after folding.
+  if (sel->AsNode()->users().size() != 1 || !other->Is<Literal>()) {
```
meheff: Use HasSingleUse as that properly considers implicit uses as well. https://github.com/google/xls/blob/732cb9f382ceb3d87cebd219d8bf74c92321dbcb/xls/ir/node_util.h#L126-L131

xls/passes/select_simplification_pass_test.cc
```diff
@@ -163,6 +163,115 @@ TEST_P(SelectSimplificationPassTest, BinaryTuplePrioritySelect) {
...
+               m::Select(m::Param("s"),
+                         {m::Eq(m::Literal(1), m::Literal(0)),
+                          m::Eq(m::Param("x"), m::Literal(0))},
+                         m::Eq(m::Literal(2), m::Literal(0))),
+               m::OneHotSelect(m::Concat(m::Not(m::Param("p")), m::Param("p")),
+                               {m::Eq(m::Literal(1), m::Literal(0)),
+                                m::Eq(m::Param("x"), m::Literal(0))})));
+}
+
+TEST_P(SelectSimplificationPassTest,
```
meheff, line 211: Add a test with >1 nonliteral cases where no transform occurs?

xls/passes/select_simplification_pass.cc
cdleary: Done.

xls/passes/select_simplification_pass.cc
cdleary: Done.

xls/passes/select_simplification_pass_test.cc
cdleary: Done.

xls/passes/select_simplification_pass.cc
```diff
@@ -1202,9 +1202,115 @@ absl::StatusOr<bool> MaybeReorderSelect(Node* node,
   return true;
 }
 
+absl::StatusOr<bool> TryHoistCompareThroughSelectLike(
```
angelomatni1, line 1205: Part of me wishes this is part of select_lifting or another pass instead of being added to select_simplification, especially because it isn't strictly simplification, but mostly because I worry about this hoist and select_lifting competing with each other. Is it possible that this transformation moves the `eq` into the select and then select_lifting pulls out the `eq`s right back out?

PR #3698, opened 2026-01-17
[opt] not/inc zero-equivalence

cdleary:
Rewrites via the observation that ~x and x+1 are "zero-equivalent".

```
  // Pattern:
  //   `(eq|ne)(sel(c, cases=[not(x), add(x, 1)]), 0)` =>
  //   `(eq|ne)(x, all_ones(width(x)))`
  //
  // Why this works (not/inc zero-equivalence):
  // - `(~x == 0)`  <=>  `(x == all_ones)`
  // - `(x+1 == 0)` <=>  `(x == all_ones)`   (note: given wraparound arithmetic)
  // Therefore for `y = sel(c, [~x, x+1])`, `(y != 0)` is exactly `(x !=
  // all_ones)`, independent of selector `c`.
```

### DIFF
--- a/xls/passes/basic_simplification_pass.cc
+++ b/xls/passes/basic_simplification_pass.cc
@@ -68,6 +68,66 @@ bool IsBinaryIrreflexiveRelation(Node* node) {
   }
 }
 
+struct EqNeSelectNotOrIncMatch {
+  Node* x;
+  Op cmp;
+};
+
+// Matches either:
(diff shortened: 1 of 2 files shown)


allight (approved):
This seems like a weird piece of code to end up with in normal compilation. I wonder if there's some more generic way to recognize cases like these.

LGTM though.

PR #3703, opened 2026-01-20
Add `const for` expression

magancarz:
This PR adds support for the `const for` expression which acts like an `unroll_for!` construct.


magancarz:
@richmckeever is there anything else required to add?

xls/dslx/frontend/parser.cc
```diff
@@ -3781,6 +3784,35 @@ absl::StatusOr<For*> Parser::ParseFor(Bindings& bindings) {
                             iterable, body, init);
 }
 
+absl::StatusOr<For*> Parser::ParseConstFor(Bindings& bindings) {
```
richmckeever: Could we instead combine the parsing for the different styles of for loops into one ParseFor function that creates the appropriate node type? The logic seems to be virtually the same.

xls/dslx/frontend/ast.h
```diff
@@ -3965,7 +3965,7 @@ class ForLoopBase : public Expr {
  public:
   ForLoopBase(Module* owner, Span span, NameDefTree* names,
               TypeAnnotation* type, Expr* iterable, StatementBlock* body,
-              Expr* init, bool in_parens = false);
+              Expr* init, bool is_const = false, bool in_parens = false);
```
richmckeever: It feels like if we add the const flag here, the value of UnrollFor as a separate node type (which forces you to care that it's different than regular For in visitors etc.) disappears.

I think we should either
1. Rename the UnrollFor node type to ConstFor and use that for both (with a flag that remembers which keyword they used).
2. or get rid of ForLoopBase and UnrollFor, and add the "which keyword they used" flag in For.

I think (1) is a little bit better because you still get the "force visitors to care" effect.

xls/dslx/frontend/parser.cc
magancarz: I've merged for-loop parsing functions into one.

xls/dslx/frontend/ast.h
magancarz: I've applied the `UnrollFor` -> `ConstFor` rename and added a flag for storing the used keyword.

magancarz:
Thank you for the review! I've applied your suggestions and also reduced the unnecessary long bytecode in the bytecode emitter `SimpleConstFor` test case.

magancarz:
@richmckeever I've noticed that the CI failed on internal checks. Can you help me find which test case fails?

PR #3704, opened 2026-01-20
Linenoise is now in BCR, use that repo directly.

hzeller:
Linenoise was added in
https://github.com/bazelbuild/bazel-central-registry/pull/7191


hzeller:
The lock file needs to be updated, but since I can't currently run the bazel in a clean environment, I leave this up to @mikex-oss

mikex-oss (approved):
I'll patch this internally with the lock file update.

PR #3745, opened 2026-01-27
Properly handle `const if` expression when using proc-scoped channels

magancarz:
This PR adds handling proper branch from `const if` when converting DSLX to IR with `lower_to_proc_scoped_channels` flag enabled. While it works on added `ConstConditionalProcScoped` test case, it breaks when trying to get a constexpr value from the `test` expression if `const if` is inside a parametric proc with the following error:

```
Error: NOT_FOUND: No constexpr value found for node `CONFIG` (NameRef) @ xls/examples/const_if.x:47:18-47:24
```

This issue can be reproduced by trying to build the [const_if_ir](https://github.com/google/xls/blob/main/xls/examples/BUILD#L1427) example with changes from this PR.

### DIFF
--- a/xls/dslx/ir_convert/get_conversion_records.cc
+++ b/xls/dslx/ir_convert/get_conversion_records.cc
@@ -75,6 +75,29 @@ class ConversionRecordVisitor : public AstNodeVisitorWithDefault {
         records_(records),
         processed_invocations_(processed_invocations) {}
 
+  absl::Status VisitProcFunctionsWithSeparateTypeInfos(
+      Module* owner, const Function* config_function,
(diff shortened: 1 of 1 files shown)


magancarz:
I've included a fix for using the correct type info when traversing parametrized proc's hierarchy in `get_conversion_records.cc` with proc-scoped channels enabled. It is based on deferring the traversal in `HandleProc` for parametric procs, so that the type info objects for `config` and `next` functions are later retrieved from spawn data and from `resolved_proc_alias` in case of top parametric procedures.
FYI @dplassgit

Issue #3746, opened 2026-01-27
Parsing breaks with non-trivial final expression in proc's config function


magancarz:
**Describe the bug**
Parsing breaks when the final expression in proc's config function is not a simple tuple, but its type is still a tuple.

**To Reproduce**
Issue can be reproduced by parsing this example:

```
const COND = true;
proc Proc {
  input: chan<()> in;
  output: chan<()> out;
  config() {
    const if COND {
      let (first_output, first_input) = chan<()>("first");
      (first_input, first_output)
    } else {
      let (second_output, second_input) = chan<()>("second");
      (second_input, second_output)
    }
  }
  init { () }
  next(state: ()) { () }
}
```

It fails with the following error:

```
INVALID_ARGUMENT: ParseError: foo.x:9:11-15:6 The final expression in a Proc config must be a tuple with one element for each Proc data member.
```

**Expected behavior**
Such expressions can be allowed to pass the parsing stage and generate an error during type checking when the type returned by the final expression is not a matching tuple.

PR #3756, opened 2026-01-28
Allow non-trivial final expressions in proc's config function

magancarz:
This PR fixes the bug mentioned in #3746 and allows more complex final expressions in proc's config function by skipping the tuple check in `parser.cc`.

Fixes: #3746

### DIFF
--- a/xls/dslx/type_system_v2/typecheck_module_v2_test.cc
+++ b/xls/dslx/type_system_v2/typecheck_module_v2_test.cc
@@ -7416,6 +7416,154 @@ proc Proc {
                 HasNodeWithType("b", "chan(sN[32], dir=in)[2]"))));
 }
 
+TEST(TypecheckV2Test, ProcConfigRequireTuple) {
+  EXPECT_THAT(
+      R"(
+proc Proc {
+  input: chan<()> in;
+  config(input: chan<()> in) {
+    (input)
+  }
+  init { () }
+  next(state: ()) { () }
+}
+
+)",
+      TypecheckFails(
+          HasTypeMismatch("(chan<()> in,)", "chan<()> in")));
+}
+
+TEST(TypecheckV2Test, ProcConfigTooFewChannels) {
+  EXPECT_THAT(
+      R"(
+proc Proc {
+  input: chan<()> in;
+  output: chan<()> out;
+  config(input: chan<()> in) {
+    (input,)
+  }
+  init { () }
+  next(state: ()) { () }
+}
+
+)",
+      TypecheckFails(
+          HasSubstr("Cannot match a 2-element tuple to 1 values.")));
+}
+
+TEST(TypecheckV2Test, ProcConfigTooManyChannels) {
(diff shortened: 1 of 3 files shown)


xls/dslx/type_system_v2/typecheck_module_v2_test.cc
```diff
@@ -7416,6 +7416,56 @@ proc Proc {
...
+  EXPECT_THAT(
+      R"(
+const A = u32:5;
+proc Proc {
+  input: chan<()> in;
+  output: chan<()> out;
+  config() {
+    const if A == u32:5 {
+      let (first_output, first_input) = chan<()>("first");
+      (first_input)
```
dplassgit: Probably should be `(first_input,)` to make it a tuple

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
```diff
@@ -7416,6 +7416,56 @@ proc Proc {
...
+
+TEST(TypecheckV2Test, ProcWithFailedBranchedFinalExpression) {
+  EXPECT_THAT(
+      R"(
+const A = u32:5;
+proc Proc {
+  input: chan<()> in;
+  output: chan<()> out;
+  config() {
+    const if A == u32:5 {
```
dplassgit, line 7551: Please add additional tests with a non-const if. Since the parser doesn't require the tuple any more let's make sure the type checker will.

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
```diff
@@ -7416,6 +7416,56 @@ proc Proc {
...
+      let (second_output, second_input) = chan<()>("second");
+      (second_input, second_output)
+    }
+  }
+  init { () }
+  next(state: ()) { () }
+}
+
+)",
+      TypecheckFails(HasTypeMismatch("(chan<()> in, chan<()> out)", "chan<()> in")));
```
dplassgit: Please add additional tests for channel type mismatch within the `const if` tuple, e.g., `chan<u32>` vs `chan<()>`

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
magancarz: I've added the following tests for the type checker to require a proper tuple:
- `ProcConfigRequireTuple`
- `ProcConfigTooFewChannels`
- `ProcConfigTooManyChannels`
- `ProcConfigMismatchingChannelTypes`

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
magancarz: I've added this in the `ProcConfigBranchedChannelTypeMismatch` test case.

magancarz:
Thanks for the review @dplassgit! I've applied requested changes.

PR #3816, opened 2026-02-10
Add enumerate() to DSLX stdlib

joannabrozek:
This PR adds the 'enumerate()' function with the 'generics' attribute to DSLX stdlib, replacing the existing built-in implementation.

This change was created because we wanted to enable conversion 'enumerate()' to IR.

### DIFF
--- a/xls/dslx/stdlib/std.x
+++ b/xls/dslx/stdlib/std.x
@@ -17,38 +17,38 @@
 #![feature(generics)]
 
 pub fn enumerate<T: type, N: u32>(x: T[N]) -> (u32, T)[N] {
-    for (i, result): (u32, (u32, T)[N]) in u32:0..N {
+    for (i, result) in 0..N {
         update(result, i, (i, x[i]))
-    }(((u32, T)[N]:[(u32:0, zero!<T>()), ...]))
+    }([(u32:0, zero!<T>()), ...])
 }
 
 #[test]
 fn emumerate_test() {
-    let array = u8[4]:[1, 2, 4, 8];
+    let array = [1, 2, 4, 8];
     let enumerated = enumerate(array);
-    assert_eq(enumerated[u32:0], (u32:0, u8:1));
-    assert_eq(enumerated[u32:1], (u32:1, u8:2));
-    assert_eq(enumerated[u32:2], (u32:2, u8:4));
-    assert_eq(enumerated[u32:3], (u32:3, u8:8));
+    assert_eq(enumerated[0], (0, 1));
+    assert_eq(enumerated[1], (1, 2));
+    assert_eq(enumerated[2], (2, 4));
+    assert_eq(enumerated[3], (3, 8));
 }
 
 #[test]
 fn enumerate_type_test() {
     type RamData = uN[8];
-    const DATA = RamData[4]:[RamData:1, RamData:2, RamData:4, RamData:8];
+    const DATA = [RamData:1, 2, 4, 8];
     let enumerated = enumerate(DATA);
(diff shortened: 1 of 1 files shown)


joannabrozek:
This PR is marked as draft since after adding function with the 'generics' attribute to stdlib, the  `xls_dslx_fmt_test` test started failing, so we're looking into it. We suspect that this test doesn't support GenericTypeAnnotation.

joannabrozek:
> This PR is marked as draft since after adding function with the 'generics' attribute to stdlib, the `xls_dslx_fmt_test` test started failing, so we're looking into it. We suspect that this test doesn't support GenericTypeAnnotation.

This issue has now been resolved, so this PR is ready for review.

tmichalak:
@richmckeever PTAL

xls/dslx/stdlib/std.x
```diff
@@ -14,6 +14,43 @@
 
 // DSLX standard library routines.
 
+#![feature(generics)]
+
+pub fn enumerate<T: type, N: u32>(x: T[N]) -> (u32, T)[N] {
+    for (i, result): (u32, (u32, T)[N]) in u32:0..N {
```
richmckeever: Type inference allows more simplification here than in the past; we should be able to write

```
    for (i, result) in 0..N {
        update(result, i, (i, x[i]))
    }([(u32:0, zero!<T>()), ...])
```

xls/dslx/stdlib/std.x
```diff
@@ -14,6 +14,43 @@
...
+        update(result, i, (i, x[i]))
+    }(((u32, T)[N]:[(u32:0, zero!<T>()), ...]))
+}
+
+#[test]
+fn emumerate_test() {
+    let array = u8[4]:[1, 2, 4, 8];
```
richmckeever: Similarly in the tests, we should leverage type inference where possible, e.g. indexes should not need type annotations, and assert_eq with an enumerated[something] on the left should not need type annotations on the right side.

For constant array declarations, the preference would be [u8:1, 2, 4, 8], [RamData:1, 2, 4, 8] etc. so you don't need to write either a repeated type or the number of elements.

joannabrozek:
I rebased on the current main branch  and applied suggested changes.
If there's anything else I need to do, let me know.

rw1nkler:
Do we need anything else here? @richmckeever

joannabrozek:
@richmckeever, I've noticed that the CI failed on internal checks. Can you help me find which test case fails?
Regarding the ZSTD pipeline, I think it should start passing after merging this PR: https://github.com/google/xls/pull/3931

joannabrozek:
I've rebased changes onto the latest main, so fix for ZSTD pipeline from 
https://github.com/google/xls/pull/3931 is now included.

kbieganski:
@richmckeever Can we get a hint as to what internal checks are failing on this?

richmckeever:
I think what we are seeing internally is just more cases where we would have to replace enumerate() with std::enumerate(). It may be easiest if you could separate the parts that are removing the old builtin into a separate PR. Then we can commit this PR, change all internal stuff to use the new function, and then you commit the removal of the old function.

joannabrozek:
@richmckeever, I split this PR into two:
* https://github.com/google/xls/pull/4021 
* https://github.com/google/xls/pull/4022

Please let me know if there is anything else I can do for you.

mikex-oss:
Closing in favor of https://github.com/google/xls/pull/4021 and https://github.com/google/xls/pull/4022.

PR #3848, opened 2026-02-14
Expose AOT compilation in C API.

meheff:
Add a function for compiling and generating aot metadata, and a separate function for running the code from the generated artifacts. Use ORC JIT for loading the object code.

### DIFF
--- a/xls/public/c_api.cc
+++ b/xls/public/c_api.cc
@@ -90,10 +90,10 @@ bool xls_convert_dslx_to_ir_with_warnings(
     size_t disable_warnings_count, bool warnings_as_errors,
     bool force_implicit_token_calling_convention, char*** warnings_out,
     size_t* warnings_out_count, char** error_out, char** ir_out) {
-  CHECK(dslx != nullptr);
-  CHECK(path != nullptr);
-  CHECK(dslx_stdlib_path != nullptr);
-  CHECK(error_out != nullptr);
+  CHECK_NE(dslx, nullptr);
+  CHECK_NE(path, nullptr);
+  CHECK_NE(dslx_stdlib_path, nullptr);
+  CHECK_NE(error_out, nullptr);
 
   std::vector<std::filesystem::path> additional_search_paths_cpp =
       xls::ToCppPaths(additional_search_paths, additional_search_paths_count);
@@ -153,11 +153,11 @@ bool xls_convert_dslx_path_to_ir_with_warnings(
     bool warnings_as_errors, bool force_implicit_token_calling_convention,
     char*** warnings_out, size_t* warnings_out_count, char** error_out,
     char** ir_out) {
-  CHECK(path != nullptr);
-  CHECK(dslx_stdlib_path != nullptr);
-  CHECK(error_out != nullptr);
+  CHECK_NE(path, nullptr);
+  CHECK_NE(dslx_stdlib_path, nullptr);
+  CHECK_NE(error_out, nullptr);
   if (warnings_out != nullptr) {
-    CHECK(warnings_out_count != nullptr);
+    CHECK_NE(warnings_out_count, nullptr);
   }
 
   std::vector<std::filesystem::path> additional_search_paths_cpp =
@@ -208,8 +208,8 @@ bool xls_convert_dslx_path_to_ir(const char* path, const char* dslx_stdlib_path,
 
 bool xls_optimize_ir(const char* ir, const char* top, char** error_out,
                      char** ir_out) {
-  CHECK(ir != nullptr);
-  CHECK(top != nullptr);
+  CHECK_NE(ir, nullptr);
+  CHECK_NE(top, nullptr);
 
   absl::StatusOr<std::string> result = xls::OptimizeIr(ir, top);
   return xls::ReturnStringHelper(result, error_out, ir_out);
@@ -217,8 +217,8 @@ bool xls_optimize_ir(const char* ir, const char* top, char** error_out,
 
 bool xls_mangle_dslx_name(const char* module_name, const char* function_name,
                           char** error_out, char** mangled_out) {
-  CHECK(module_name != nullptr);
-  CHECK(function_name != nullptr);
+  CHECK_NE(module_name, nullptr);
+  CHECK_NE(function_name, nullptr);
 
   absl::StatusOr<std::string> result =
       xls::MangleDslxName(module_name, function_name);
@@ -252,10 +252,10 @@ bool xls_mangle_dslx_name_full(
     xls_calling_convention convention, const char* const free_keys[],
     size_t free_keys_count, const struct xls_dslx_parametric_env* param_env,
     const char* scope, char** error_out, char** mangled_out) {
-  CHECK(module_name != nullptr);
-  CHECK(function_name != nullptr);
-  CHECK(error_out != nullptr);
-  CHECK(mangled_out != nullptr);
+  CHECK_NE(module_name, nullptr);
+  CHECK_NE(function_name, nullptr);
+  CHECK_NE(error_out, nullptr);
+  CHECK_NE(mangled_out, nullptr);
 
   xls::dslx::CallingConvention cc_cpp;
   if (!CallingConventionFromC(convention, &cc_cpp, error_out)) {
@@ -264,7 +264,7 @@ bool xls_mangle_dslx_name_full(
 
   absl::btree_set<std::string> free_key_set;
   for (size_t i = 0; i < free_keys_count; ++i) {
-    CHECK(free_keys[i] != nullptr);
+    CHECK_NE(free_keys[i], nullptr);
     free_key_set.insert(free_keys[i]);
   }
 
@@ -285,11 +285,11 @@ bool xls_schedule_and_codegen_package(
     xls_package* p, const char* scheduling_options_flags_proto,
     const char* codegen_flags_proto, bool with_delay_model, char** error_out,
     struct xls_schedule_and_codegen_result** result_out) {
-  CHECK(p != nullptr);
-  CHECK(scheduling_options_flags_proto != nullptr);
-  CHECK(codegen_flags_proto != nullptr);
-  CHECK(error_out != nullptr);
-  CHECK(result_out != nullptr);
+  CHECK_NE(p, nullptr);
+  CHECK_NE(scheduling_options_flags_proto, nullptr);
+  CHECK_NE(codegen_flags_proto, nullptr);
+  CHECK_NE(error_out, nullptr);
+  CHECK_NE(result_out, nullptr);
 
   xls::Package* cpp_package = reinterpret_cast<xls::Package*>(p);
 
@@ -327,7 +327,7 @@ bool xls_schedule_and_codegen_package(
 
 char* xls_schedule_and_codegen_result_get_verilog_text(
     const struct xls_schedule_and_codegen_result* result) {
-  CHECK(result != nullptr);
+  CHECK_NE(result, nullptr);
   auto* cpp_result =
       reinterpret_cast<const xls::ScheduleAndCodegenResult*>(result);
   return xls::ToOwnedCString(cpp_result->codegen_result.verilog_text);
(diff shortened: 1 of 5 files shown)


xls/public/c_api.cc
```diff
@@ -1333,20 +1402,266 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+    return false;
+  }
+
+  absl::StatusOr<std::unique_ptr<xls::OrcJit>> maybe_orc_jit =
+      xls::OrcJit::Create();
+  if (!maybe_orc_jit.ok()) {
+    *error_out = xls::ToOwnedCString(maybe_orc_jit.status().ToString());
+    return false;
+  }
+  std::unique_ptr<xls::OrcJit> orc_jit = std::move(*maybe_orc_jit);
```
allight: Since you aren't actually using any xls::OrcJit functionality can you just call the llvm functions directly here. As is this is doing a bunch of stuff you don't care about since you're using it as essentially a combined 'ld' and 'dlopen'.

xls/public/c_api.cc
```diff
@@ -1333,20 +1402,266 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+    return false;
+  }
+
+  const xls::FunctionEntrypoint& entrypoint = object_code->entrypoints.front();
+  auto context = std::make_unique<llvm::LLVMContext>();
+  xls::LlvmTypeConverter type_converter(context.get(), object_code->data_layout);
+  xls::Package* package = object_code->package ? object_code->package.get()
+                                               : xls_function->package();
+  absl::StatusOr<xls::AotEntrypointProto> aot_entrypoint =
+      xls::GenerateAotEntrypointProto(package, entrypoint, kHasMsan,
```
allight: I'm not sure this is actually enough to make msan work since I think you'd want to link it to the main processes TLS/msan stuff but I don't think the way you use orc-jit as a linker would allow for that.

allight (commented):
I don't have any serious issues with this but looking at the xlsynth-crate branch I wonder if it might be simpler to just have fork-exec the aot_compiler_main and use it that way. This is a lot of code to cram into xls-c-api.

I also wonder whether it might be easier to give the aot stuff the ability to run lld (through the c++ library) and then a lot of this gets a bunch simpler since you can have your stuff just dlopen and dlsym the library.

xls/public/c_api.cc
```diff
@@ -1333,20 +1402,266 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+  xls::JitFunctionType fn = absl::bit_cast<xls::JitFunctionType>(function_ptr);
+  int64_t continuation = fn(inputs, outputs, temp_buffer, &context->events,
+                            &context->instance_context, context->jit_runtime.get(),
+                            continuation_point);
+  *trace_messages_count_out = context->events.GetTraceMessages().size();
+  *assert_messages_count_out = context->events.GetAssertMessages().size();
+  return continuation;
+}
+
+bool xls_aot_run_function(
```
allight: Thinking about this a little more it would be much cleaner IMO if this function took a function-pointer (and all the entrypoints info etc) and the rust side was responsible for linking in the object code (either using orcjit or lld or something).

Heck it would be easy enough to have it return a struct to hold the entire jit so you can pass multiple calls to it without reloading it every time.

xls/public/c_api.cc
meheff: Removed. false is passed unconditionally here.

xls/public/c_api.cc
meheff: Removed the function altogether.

xls/public/c_api.cc
meheff: Removed LoadObjectCode method which was added to OrcJit and just called the llvm functions directly.

meheff:
Thanks for the review, Alex. I simplified things quite a bit which I think resolves your comments:

* Removed xls_aot_run_function from the C API. This took object code and ran it with irvalues. This was more for testing than for production purposes and isn't really necessary.
* Removed the added LoadObjectCode method from OrcJit. The test now just call llvm directly to load the object code.

Regarding fork-exec'ing aot_compiler_main, this would require carrying the aot_compiler_main binary along with the library. The proto generation code is already shared with aot_compiler_main, and the object code generation is just a single function call. The rest of the code in xls_aot_compile_function is just the verbose fluff needed to do things C style (manual allocation/free).

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+  CHECK(context != nullptr);
+  context->events.Clear();
+}
+
+void xls_aot_exec_context_free(struct xls_aot_exec_context* context) {
+  delete context;
+}
+
+int64_t xls_aot_entrypoint_trampoline(
+    uintptr_t function_ptr, const uint8_t* const* inputs,
```
allight: Can you have this be `void* function_ptr` since it does need to be an actual memory location.

allight (commented):
One minor change and LGTM.

I would like to remind you too that though we won't go around changing it for no reason the abi of our generated entrypoints shouldn't really be considered stable. We may want to change it dramatically for any number of reasons. Practically the only thing likely to be long term (relatively) stable are the external interfaces our wrapper rules create. For something more long term stable I'd really suggest having an api that you give the entrypoint pointer and metadata and it returns a handle that you can request to execute using some defined 'xls::Value' mirror in the c api. For now though this is fine.

xls/public/c_api.cc
meheff: fixed.

meheff:
> One minor change and LGTM.
> 
> I would like to remind you too that though we won't go around changing it for no reason the abi of our generated entrypoints shouldn't really be considered stable. We may want to change it dramatically for any number of reasons. Practically the only thing likely to be long term (relatively) stable are the external interfaces our wrapper rules create. For something more long term stable I'd really suggest having an api that you give the entrypoint pointer and metadata and it returns a handle that you can request to execute using some defined 'xls::Value' mirror in the c api. For now though this is fine.

Thanks for the heads up and understood about no guarantee about stability. We're willing to tradeoff stability for performance here (passing in IrValues has some significant performance penalty). Since we don't live at head we can fix things as necessary when they break. We have a wrapper layer in rust which insulates end users from the xls abi details and it should be flexible enough to deal with xls-side changes.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
   delete reinterpret_cast<xls::FunctionJit*>(jit);
 }
 
+bool xls_aot_compile_function(struct xls_function* function, char** error_out,
+                              uint8_t** object_code_out,
+                              size_t* object_code_count_out,
+                              uint8_t** entrypoints_proto_out,
+                              size_t* entrypoints_proto_count_out) {
+  CHECK(function != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+
+void xls_aot_entrypoints_proto_free(uint8_t* entrypoints_proto) {
+  xls_bytes_free(entrypoints_proto);
+}
+
+bool xls_aot_exec_context_create(const uint8_t* entrypoints_proto,
+                                 size_t entrypoints_proto_count,
+                                 char** error_out,
+                                 struct xls_aot_exec_context** out) {
+  CHECK(entrypoints_proto != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+    return false;
+  }
+
+  *out = context;
+  *error_out = nullptr;
+  return true;
+}
+
+void xls_aot_exec_context_clear_events(struct xls_aot_exec_context* context) {
+  CHECK(context != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+void xls_aot_exec_context_free(struct xls_aot_exec_context* context) {
+  delete context;
+}
+
+int64_t xls_aot_entrypoint_trampoline(
+    void* function_ptr, const uint8_t* const* inputs, uint8_t* const* outputs,
+    void* temp_buffer, struct xls_aot_exec_context* context,
+    int64_t continuation_point, size_t* trace_messages_count_out,
+    size_t* assert_messages_count_out) {
+  CHECK(function_ptr != 0);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+                            context->jit_runtime.get(), continuation_point);
+  *trace_messages_count_out = context->events.GetTraceMessages().size();
+  *assert_messages_count_out = context->events.GetAssertMessages().size();
+  return continuation;
+}
+
+bool xls_aot_exec_context_get_trace_message(
+    const struct xls_aot_exec_context* context, size_t index, char** error_out,
+    struct xls_trace_message* trace_message_out) {
+  CHECK(context != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1333,6 +1347,237 @@ void xls_function_jit_free(struct xls_function_jit* jit) {
...
+          ? trace_messages.Get(index).statement().verbosity()
+          : 0;
+  *error_out = nullptr;
+  return true;
+}
+
+bool xls_aot_exec_context_get_assert_message(
+    const struct xls_aot_exec_context* context, size_t index, char** error_out,
+    char** assert_message_out) {
+  CHECK(context != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

xls/public/c_api.cc
```diff
@@ -1346,6 +1591,10 @@ bool xls_function_jit_run(struct xls_function_jit* jit, size_t argc,
                           struct xls_value** result_out) {
   CHECK(jit != nullptr);
   CHECK(error_out != nullptr);
+  CHECK(trace_messages_out != nullptr);
```
proppy: > Consider using CHECK_NE(a, b) instead of CHECK(a != b)

Internal lint warning here and below.

proppy:
FYI, got some internal lint warnings, but it seems that the currently checked code also violate those checks so I'm not sure if that's really for yours to fix!

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

xls/public/c_api.cc
meheff: Fixed.

meheff:
> FYI, got some internal lint warnings, but it seems that the currently checked code also violate those checks so I'm not sure if that's really for yours to fix!

I went ahead and changed all instances in xls/public. This should avoid this issue in the future for both human and AI coders who tend to follow what is already in the file.

proppy (approved):

proppy:
@meheff sorry for the extra chore but can you also rebase, this seems to conflict at `xls/jit/BUILD#212,212`

meheff:
> @meheff sorry for the extra chore but can you also rebase, this seems to conflict at `xls/jit/BUILD#212,212`

done.

PR #3860, opened 2026-02-18
[DSLX:FE] Warn on params & struct members that are SV keywords, as they fail in verilog land

cdleary:
Adds a DSLX warning for normal function parameters that use Verilog/SystemVerilog keywords (new `keyword_parameter_name` warning kind). Renames affected function parameters across `.x` files (stdlib/examples/modules/tests/third_party) to avoid new warnings and keep formatting consistent.

Previously if you tried to codegen a DSLX function that had "input" as a parameter it would fail in codegen with no early warning at DSLX development time. Recall that we don't auto-sanitize these as module ports are part of the resulting module interface.


allight:
@richmckeever PTAL. I'm fine with just pulling this in though I wonder if it would be better to just have ir_convert do a better job at mangling names.

cdleary:
@allight I noted in the PR that I believe we don't do that rn as a policy because we want what appears in the DSLX to be the module signature that the user observes. We could /also/ do that, but even in that case I'd still want this warning available.

cdleary:
Also note the zstd breakage appears unrelated to this change.

allight (approved):

proppy (approved):

proppy:
interestingly I don't see the zstd failure in https://github.com/google/xls/actions/workflows/modules-zstd.yml?query=is%3Afailure, @rw1nkler do you have a clue?

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
```diff
@@ -18,18 +18,19 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
 #include "xls/common/status/matchers.h"
 #include "xls/dslx/create_import_data.h"
 #include "xls/dslx/import_data.h"
 #include "xls/dslx/type_system/typecheck_test_utils.h"
 #include "xls/dslx/type_system_v2/matchers.h"
 #include "xls/dslx/type_system_v2/type_system_test_utils.h"
 #include "xls/dslx/virtualizable_file_system.h"
+#include "xls/dslx/warning_kind.h"
```
proppy, line 34: can we also the `cc_library` in the `deps` of the `cc_test`?

(this currently fails build cleaner)

proppy:
@cdleary  can we also rebase?

rw1nkler:
> interestingly I don't see the zstd failure in https://github.com/google/xls/actions/workflows/modules-zstd.yml?query=is%3Afailure, @rw1nkler do you have a clue?

This is probably caused by `continue-on-error`: https://github.com/google/xls/blob/main/.github/workflows/modules-zstd.yml#L23, that we set to run all the tests and have more information about the failures. We will check if this is the case, and open a PR which removes this option, to make the error easier to spot in the future.

> Also note the zstd breakage appears unrelated to this change.

We are working on fixing zstd. It seems that the earliest failures in the CI could be observed when proc-scoped channels were introduced: https://github.com/google/xls/pull/3717. Hopefully, we will have a fix for it soon.

xls/dslx/type_system_v2/typecheck_module_v2_test.cc
cdleary: Done.

cdleary:
@proppy should be rebased now, LMK if you have feedback

proppy (approved):

proppy:
@cdleary, sorry can you rebase again?

proppy:
marking this as a draft for now (so that it's out of the review queue), @cdleary  can you re-mark is as active once you rebased?

PR #3882, opened 2026-02-24
Add IR conversion feedback to the DSLX interpreter

magancarz:
This PR adds an IR conversion test to the DSLX interpreter enabled by `lower_to_ir` flag which is set by default. It performs the IR conversion check on every non-parametric proc or function and returns error if any fails. Tests can also be included in the check by adding `convert_tests` flag.

### DIFF
--- a/xls/build_rules/xls_dslx_rules.bzl
+++ b/xls/build_rules/xls_dslx_rules.bzl
@@ -112,6 +112,8 @@ def _get_dslx_test_cmdline(ctx, src, all_srcs, append_cmd_line_args = True):
         "format_preference",
         "configured_values",
         "lower_to_proc_scoped_channels",
+        "lower_to_ir",
+        "convert_tests",
     )
 
     dslx_test_args = dict(_dslx_test_args)
--- a/xls/dslx/BUILD
+++ b/xls/dslx/BUILD
@@ -618,6 +618,7 @@ cc_binary(
     visibility = ["//xls:xls_users"],
     deps = [
         ":command_line_utils",
+        ":create_import_data",
         ":default_dslx_stdlib_path",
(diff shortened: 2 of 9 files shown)


rw1nkler:
FYI @richmckeever

proppy:
should we also add this to typecheck_main?

proppy:
This also need to be rebased.

magancarz:
I've rebased changes on the latest main.

> should we also add this to typecheck_main?

Sounds reasonable. @richmckeever what do you think?

mikex-oss:
There seems to be no way to configure IR convert options through `xls_dslx_test`. 

Rather than overloading xls_dslx_test in this way, what do we think about a separate `xls_dslx_ir_convert_test` or similar that just calls `xls_dslx_ir` + `build_test`? This is similar to how we have a separate `xls_dslx_fmt_test`, rather than doing it as `check_formatting = True` inside `xls_dslx_test`.

magancarz:
@mikex-oss The intent behind IR conversion check when running DSLX tests was to prevent situations where the user would spend time on developing DSLX code that wouldn't be possible to convert to IR, which is an issue that would surface only when the user would run a xls_dslx_ir target. Moving this task to another target would make these changes unnecessary. @richmckeever What do you think?

richmckeever (approved):

meheff:
It looks like this results in quadratic behavior in the size of the DSLX input file. In particular this code:

```
for (std::string& elem : module_elements) {
    ConvertFilesToPackage(..., elem, ...);
  }
```

AIUI, this will do a full parse/conversion of the files for ~every module member. We are seeing timeouts because of this. 

Can this feature default to false?

PR #3889, opened 2026-02-25
Prevent parameter names of top functions from being changed during ir conversion.

meheff:
The parameter names of top functions become part of the exposed interface in the generated code. These names need to be preserved. This change raises an error if the legalized name during IR conversion differs from the original name.

A potentially more flexible change would be allow keywords as identifiers, but that is pretty invasive and I don't know if there are hidden problems here. This change at lease prevents silently renaming the visible interface of functions/modules.

### DIFF
--- a/xls/dslx/ir_convert/function_converter.cc
+++ b/xls/dslx/ir_convert/function_converter.cc
@@ -819,18 +819,27 @@ absl::Status FunctionConverter::HandleParam(const Param* node) {
   VLOG(5) << "FunctionConverter::HandleParam: " << node->ToString();
   XLS_ASSIGN_OR_RETURN(xls::Type * type,
                        ResolveTypeToIr(node->type_annotation()));
-  Def(node->name_def(), [&](const SourceInfo& loc) {
+  BValue param = Def(node->name_def(), [&](const SourceInfo& loc) {
     return function_builder_->Param(node->identifier(), type);
   });
+  if (is_top_ && param.GetName() != node->identifier()) {
+    return IrConversionErrorStatus(
+        node->span(),
+        absl::StrFormat(
+            "Top function parameter name `%s` would be changed to `%s` in IR. "
+            "Choose a different name.",
+            node->identifier(), param.GetName()),
+        file_table());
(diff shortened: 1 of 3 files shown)


xls/dslx/ir_convert/function_converter.cc
```diff
@@ -819,18 +819,27 @@ absl::Status FunctionConverter::HandleParam(const Param* node) {
...
                        ResolveTypeToIr(node->type_annotation()));
-  Def(node->name_def(), [&](const SourceInfo& loc) {
+  BValue param = Def(node->name_def(), [&](const SourceInfo& loc) {
     return function_builder_->Param(node->identifier(), type);
   });
+  if (is_top_ && param.GetName() != node->identifier()) {
+    return IrConversionErrorStatus(
+        node->span(),
+        absl::StrFormat(
+            "Top function parameter name `%s` would be changed to `%s` in IR. "
```
proppy, line 829: would it makes sense to fail earlier instead of legalizing and then catching that something has changed?

meheff: I think this is the most precise test and least error prone way of doing this. If this is done earlier, it needs to replicate the logic of the uniquification/legalization and there is risk of divergence between the two implementations. Also, as a minor point, this only contrains names if the the DSLX is lowered to IR and not in all cases. This is similar to the verilog keyword constraints where you can use verilog keywords however you want in DSLX, but when and only when you lower to verilog does the error pop up for clashing durable identifiers.

proppy: > this only contrains names if the the DSLX is lowered to IR

One issue I was worried about is that this might contribute to some of the current https://en.wikipedia.org/wiki/Principle_of_least_astonishment tax we have: when something pass DSLX typecheck/test but doesn't convert (and the developer only figure out about it "later" in the dev cycle).

/cc @richmckeever and @grebe as we discussed this again recently.

Related #3882

proppy (approved):

proppy (approved):

PR #3896, opened 2026-02-27
Throw `invalid_argument` error in case of dynamic indexing of channels array

magancarz:
This PR adds throwing an error in case of dynamic indexing of channels array. It is handled by checking constness of value used for indexing an array of channels in `semantics_analysis.cc`.

- [x]  Fix `test_constexpr_rollover_warning` test in `error_modules_test.py`.

### DIFF
--- a/xls/dslx/type_system/typecheck_module_test.cc
+++ b/xls/dslx/type_system/typecheck_module_test.cc
@@ -4854,5 +4854,104 @@ TEST_F(TypecheckV2Test, AssertFmtArgCountMismatch) {
                          "but has 1 argument(s)")));
 }
 
+TEST_F(TypecheckV2Test, DynamicChannelArrayIndexing) {
+  constexpr std::string_view kProgram = R"(
+proc Test {
+  sel_r: chan<u32> in;
+  array_r: chan<u32>[10] in;
+  data_s: chan<u32> out;
+
+  init { }
+
+  config(
+     sel_r: chan<u32> in,
+     array_r: chan<u32>[10] in,
+     data_s: chan<u32> out
+  ) { (sel_r, array_r, data_s) }
+
+  next(state: ()) {
+    let (tok, sel) = recv(join(), sel_r);
+    let (tok, result) = recv(tok, array_r[sel]);
+    let tok = send(tok, data_s, result);
+  }
+}
+)";
+  EXPECT_THAT(
+      Typecheck(kProgram),
+      StatusIs(absl::StatusCode::kInvalidArgument,
+               HasSubstr("Non-constexpr value used in channel array indexing")));
+}
+
+TEST_F(TypecheckV2Test, ProcMemberInDynamicLoop) {
+  constexpr std::string_view kProgram = R"(
+proc Proc {
+  inputs: chan<s32>[2] in;
+  outputs: chan<s32>[2] out;
+  config() {
+    let (a, b) = chan<s32>[2]("c");
+    (b, a)
+  }
+
+  init {  }
+
+  next(state: ()) {
+    let tok = join();
+    for (i, _) in u32:0..u32:2 {
+        recv(tok, inputs[i]);
+        send(tok, outputs[i], s32:1);
+    } (());
+  }
+}
+)";
+  EXPECT_THAT(
+      Typecheck(kProgram),
+      StatusIs(absl::StatusCode::kInvalidArgument,
+               HasSubstr("Non-constexpr value used in channel array indexing")));
+}
+
+TEST_F(TypecheckV2Test, ProcMemberInUnrollFor) {
+  XLS_EXPECT_OK(Typecheck(R"(
+proc Proc {
+  inputs: chan<s32>[2] in;
+  outputs: chan<s32>[2] out;
+  config() {
+    let (a, b) = chan<s32>[2]("c");
+    (b, a)
+  }
+
+  init {  }
+
+  next(state: ()) {
+    let tok = join();
+    unroll_for! (i, _) in u32:0..u32:2 {
+        recv(tok, inputs[i]);
+        send(tok, outputs[i], s32:1);
+    } (());
+  }
+})"));
+}
+
+TEST_F(TypecheckV2Test, ProcMemberInConstFor) {
+  XLS_EXPECT_OK(Typecheck(R"(
+proc Proc {
+  inputs: chan<s32>[2] in;
+  outputs: chan<s32>[2] out;
+  config() {
+    let (a, b) = chan<s32>[2]("c");
+    (b, a)
+  }
+
+  init {  }
+
+  next(state: ()) {
+    let tok = join();
+    const for (i, _) in u32:0..u32:2 {
+        recv(tok, inputs[i]);
+        send(tok, outputs[i], s32:1);
+    } (());
+  }
+})"));
+}
+
 }  // namespace
 }  // namespace xls::dslx
--- a/xls/dslx/type_system_v2/constant_collector.cc
+++ b/xls/dslx/type_system_v2/constant_collector.cc
@@ -345,6 +345,13 @@ class Visitor : public AstNodeVisitorWithDefault {
   }
 
   absl::Status HandleIndex(const Index* index) override {
+    if (type_.IsChannel() && std::holds_alternative<Expr*>(index->rhs()) &&
+        !Evaluate(std::get<Expr*>(index->rhs())).ok()) {
+      return xls::dslx::TypeInferenceErrorStatus(
+          index->span(), &type_,
+          "Non-constexpr value used in channel array indexing", file_table_);
+    }
+
     // A `Slice` actually has its bounds stored in `TypeInfo` out-of-band from
     // the real type info, mirroring the `StartAndWidthExprs` that we store in
     // the `InferenceTable`.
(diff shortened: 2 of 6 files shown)


magancarz:
`test_constexpr_rollover_warning` test was failing, because `constant_collector` evaluated and noted constexpr value for `Binop` without rollover warning before `evaluator.cc` had a chance to do it. I've had to follow the workaround used [here](https://github.com/google/xls/blob/main/xls/dslx/type_system_v2/evaluator.cc#L144). After fixing that test case, I've removed the draft state and the pull request is now ready for review.

magancarz:
I've rebased changes on latest main, so Continuous Integration job will now succeed.

xls/dslx/type_system/typecheck_module_test.cc
```diff
@@ -5220,5 +5220,62 @@ TEST_F(TypecheckV2Test, AssertFmtArgCountMismatch) {
...
+  config() {
+    let (a, b) = chan<s32>[2]("c");
+    (b, a)
+  }
+
+  init {  }
+
+  next(state: ()) {
+    let tok = join();
+    for (i, _) in u32:0..u32:2 {
```
dplassgit, line 4899: I'd like to see another test that has an unroll_for! which succeeds in contrast to this one which (correctly) does not.

magancarz: Added a test that has a `const for` as well.

magancarz:
Thanks for the review! I've applied your suggestion.

dplassgit (approved):

xls/dslx/frontend/semantics_analysis.cc
```diff
@@ -761,6 +562,20 @@ void SemanticsAnalysis::SetNameDefType(const NameDef* def, const Type* type) {
...
 }
 
+absl::Status SemanticsAnalysis::CheckChannelArrayIndex(
+    const Index* array_index, const TypeInfo* type_info) {
+  VLOG(5) << absl::StrFormat("CheckChannelArrayIndex: %s",
+                             array_index->ToString());
+  const AstNode* rhs = ToAstNode(array_index->rhs());
+  std::vector<AstNode*> rhs_children = rhs->GetChildren(false);
+  if (!type_info->IsKnownConstExpr(rhs)) {
+    return absl::InvalidArgumentError(absl::StrFormat(
```
proppy: could we also add a line number?

magancarz: Sure, now the error is in form of:

```
foo/example.x:36:42-36:47
0034:   next(state: ()) {
0035:     let (tok, sel) = recv(join(), sel_r);
0036:     let (tok, result) = recv(tok, array_r[sel]);
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^---^ ParseError: Non-constexpr value used in channel array indexing
0037:     let tok = send(tok, data_s, result);
0038:   }
```

magancarz:
I've added more descriptive error log. See [review comment](https://github.com/google/xls/pull/3896#discussion_r2888288851).

richmckeever (approved):

magancarz:
Merging #3931 should help with the failing `Test ZSTD module` job.

richmckeever (approved):

dplassgit:
This is failing internal tests; the DSLX index is correctly a const (it's the index in an unroll for) but the new check is claiming it's not a const.

magancarz:
> This is failing internal tests; the DSLX index is correctly a const (it's the index in an unroll for) but the new check is claiming it's not a const.

@dplassgit the check is based on confirming that the index value is a known const, so I suspect that constant collector couldn't collect the value before the check (it was a similar case with `Binop`) or the used type info object is incorrect. Can you help me identify which AST node was used as the index?

dplassgit:
It looks something like this (partially redacted)

```
 let v =
            unroll_for! (st_slot_idx, store_data_value): (
                u32, common::dv[NS][NV]
            ) in u32:0..NV {
                let (_, sdv) = recv_if(
                    tok, channel_array[st_slot_idx],  // <<< failing on st_slot_idx
                    ready,
                    zero!<...>());
            }(zero!<...>());
```

wsipak:
Hey @dplassgit ,
the code sample you've provided doesn't seem sufficient to reproduce the error you're seeing in the test.
I've created a proc with a similar `unroll_for` and channel array access, but I haven't found a way to trigger the error here.
Is there a chance you can provide more information or a minimal reproducible example with the error?

magancarz:
@dplassgit you can find our attempt to reproduce your failing test case in [this commit](https://github.com/antmicro/xls/commit/7171c77d0363a74f162e5e5fec82099caa58a57e) in our fork.

magancarz:
I've rebased changes onto the latest main, so `Test ZSTD module` should pass now and changed instance of `for` to `const for` in `zstd_dec_test.x` which also triggered the dynamic indexing error.

mikex-oss:
Re-approving to run internal checks.

mikex-oss (approved):

kbieganski:
@dplassgit Were you able to narrow down the problem?

dplassgit:
Since @mikex-oss approved, I'm satisfied it's OK.

mikex-oss:
@dplassgit to clarify see https://github.com/google/xls/pull/3896#issuecomment-4156944427.

I wanted to check if the newer commits fixed the issues. Unfortunately, the problem still persists on the internal code. I left a comment on the import CL.

proppy (approved):

dplassgit:
The updated PR still breaks internal tests. @mikex-oss @richmckeever do you have any ideas?

xls/dslx/frontend/semantics_analysis.cc
```diff
@@ -769,6 +770,22 @@ void SemanticsAnalysis::SetNameDefType(const NameDef* def, const Type* type) {
   }
 }
 
+absl::Status SemanticsAnalysis::CheckChannelArrayIndex(
+    const Index* array_index, const TypeInfo* type_info,
+    const FileTable& file_table) {
+  VLOG(5) << absl::StrFormat("CheckChannelArrayIndex: %s",
```
richmckeever: The internal failure is a separate bug fixed by https://github.com/google/xls/pull/4303. That should get this PR to pass.

But after looking at this more closely to investigate, I think we should move the check to the HandleIndex function in constant_collector.cc rather than involving SemanticsAnalysis (no need to add an extra hook since that existing one can capture it).

xls/dslx/frontend/semantics_analysis.cc
```diff
@@ -769,6 +770,22 @@ void SemanticsAnalysis::SetNameDefType(const NameDef* def, const Type* type) {
...
+                             array_index->ToString());
+  const AstNode* rhs = ToAstNode(array_index->rhs());
```
richmckeever: Once moved to ConstantCollector, don't forget to limit the check to when the RHS variant holds an Expr* (i.e. not a slice) and the type IsChannel (currently the caller does it).

xls/dslx/frontend/semantics_analysis.cc
```diff
@@ -769,6 +770,22 @@ void SemanticsAnalysis::SetNameDefType(const NameDef* def, const Type* type) {
...
+  std::vector<AstNode*> rhs_children = rhs->GetChildren(false);
+  if (!type_info->IsKnownConstExpr(rhs)) {
```
richmckeever: Once moved to ConstantCollector, use Evaluate(rhs).ok() as opposed to IsKnownConstExpr(rhs).

xls/dslx/frontend/semantics_analysis.cc
```diff
@@ -769,6 +770,22 @@ void SemanticsAnalysis::SetNameDefType(const NameDef* def, const Type* type) {
...
+    return xls::dslx::ParseErrorStatus(
```
richmckeever: Once moved, use TypeInferenceErrorStatus.

richmckeever (changes requested):

proppy:
@magancarz can you rebase?

magancarz:
@richmckeever Thanks for more suggestions. I've applied them as requested and rebased changes onto the latest `main`.

xls/dslx/type_system_v2/constant_collector.cc
```diff
@@ -91,6 +91,21 @@ class Visitor : public AstNodeVisitorWithDefault {
         trace_(std::move(trace)),
         next_channel_id_(next_channel_id) {}
 
+  absl::Status HandleBinop(const Binop* binop) override {
+    // See TODO comment in ConstexprEvaluator::InterpretExpr(const Expr*).
+    bool warn_rollover =
+        binop->parent() && binop->parent()->kind() ==
+                           AstNodeKind::kParametricBinding;
+    absl::StatusOr<InterpValue> value = ConstexprEvaluator::EvaluateToValue(
```
richmckeever: Eagerly evaluating every binop here seems like it would degrade performance, primarily in cases we don't care about for channel indexing. What happens if you take out this HandleBinop function? Does the Evaluate() in HandleIndex() somehow fail?

richmckeever (changes requested):

xls/dslx/type_system_v2/constant_collector.cc
rw1nkler: I removed the `HandleBinop()` and the existing tests still pass

richmckeever (approved):

magancarz:
@richmckeever I've noticed that the CI failed on Google's internal checks. Can you help me find which test case failed?

proppy (approved):

PR #3899, opened 2026-02-28
[codegen] fix bug in mod emission when divisor was zero, not XLS IR semantics

cdleary:
previous expression we would emit as verilog would be Xs when the divisor was zero, which is not equivalent to the original XLS IR semantics

also adds tests for signed ops with negative divisors

### DIFF
--- a/xls/codegen/testdata/combinational_generator_test_SMod.svtxt
+++ b/xls/codegen/testdata/combinational_generator_test_SMod.svtxt
@@ -1,0 +1,14 @@
+module SMod(
+  input wire [31:0] x,
+  input wire [31:0] y,
+  output wire [31:0] out
+);
+  function automatic [31:0] smod_32b (input reg [31:0] lhs, input reg [31:0] rhs);
+    begin
+      smod_32b = rhs == 32'h0000_0000 ? 32'h0000_0000 : $unsigned($signed(lhs) % $signed(rhs));
+    end
+  endfunction
+  wire [31:0] smod_6;
+  assign smod_6 = smod_32b(x, y);
+  assign out = smod_6;
+endmodule
--- a/xls/codegen/testdata/combinational_generator_test_SModSmallWidthNegativeDivisor.svtxt
+++ b/xls/codegen/testdata/combinational_generator_test_SModSmallWidthNegativeDivisor.svtxt
@@ -1,0 +1,14 @@
+module SModSmallWidthNegativeDivisor(
+  input wire [3:0] x,
+  input wire [3:0] y,
+  output wire [3:0] out
+);
+  function automatic [3:0] smod_4b (input reg [3:0] lhs, input reg [3:0] rhs);
+    begin
+      smod_4b = rhs == 4'h0 ? 4'h0 : $unsigned($signed(lhs) % $signed(rhs));
+    end
+  endfunction
+  wire [3:0] smod_6;
+  assign smod_6 = smod_4b(x, y);
+  assign out = smod_6;
+endmodule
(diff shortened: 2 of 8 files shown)


proppy:
> fix bug in div/mod emission

nit: looks like this is just fixing `SMod/UMod` is this intentional?

xls/codegen/testdata/combinational_generator_test_SMod.vtxt
proppy, line 1: don't we also need the `.svtxt` version?

xls/codegen/testdata/combinational_generator_test_UMod.vtxt
proppy, line 1: don't we also need the `.svtxt` version?

xls/codegen_v_1_5/testdata/codegen_combinational_function_test_SMod.vtxt
proppy, line 1: don't we also need the `.svtxt` version?

xls/codegen_v_1_5/testdata/codegen_combinational_function_test_SModSmallWidthNegativeDivisor.vtxt
proppy, line 1: don't we also need the `.svtxt` version?

proppy (changes requested):

xls/codegen/testdata/combinational_generator_test_SMod.vtxt
cdleary: Added the matching SystemVerilog golden for completeness; there isnt any SV-specific behavior here, but it makes sense to keep the paired goldens.

xls/codegen_v_1_5/testdata/codegen_combinational_function_test_SMod.vtxt
cdleary: Added, per above.

xls/codegen_v_1_5/testdata/codegen_combinational_function_test_SModSmallWidthNegativeDivisor.vtxt
cdleary: Added, per above.

xls/codegen/testdata/combinational_generator_test_UMod.vtxt
cdleary: Added, per above.

cdleary:
UDiv/SDiv already had guarded lowering, and this change adds the matching guarded behavior for SMod/UMod. I also added the missing UMod coverage so the full {u,s}{div,mod} test matrix is explicit in both combinational suites.

ericastor:
That makes a lot of sense - though then I think the PR name has ended up a little misleading. I've taken the liberty of fixing that.

Thanks for getting the svtxt goldens in place; it was breaking our tests otherwise.

ericastor:
Interesting - we're still getting some test failures internally, mostly involving crashes during various simulation runs! We're going to need to look into these.

cdleary:
Okay thanks for the update -- let me know if I can help in some way.

I double checked based on your note and think this should be "what we want" via verilog semantics -- the ternary condition being non-x means we select just the arm of interest. /If/ the condition were x then it would "x merge" the left hand side and right hand sides, but the idea is we should avoid x values flowing in XLS-generated verilog, so the condition should not be x in any normal circumstance.

ericastor:
Sorry it took a while to get back to this. It seems the new Verilog is crashing DSim with a `SIGFPE`, even though the result of the mod operation isn't used? Which seems strange when the old Verilog wasn't causing problems.

EDIT: This only happens for SMod, not UMod, and only on the (MinSigned() % -1) test... so the issue may well be the test finding a DSim bug, rather than the changed Verilog.

ericastor:
We've tracked the issue down to what appears to be a bug in the Verilog simulator in question. I'll merge this with a small patch to let us skip that specific test for buggy simulators.

Issue #3637, opened 2025-12-31
`xls_benchmark_ir` targets failing due to internal checks


magancarz:
**Describe the bug**
Some `xls_benchmark_ir` targets fail due to an internal error e.g.:

```
F1209 20:24:43.801930   52301 lazy_node_data.h:289 Check failed: f_->Contains(node) Must be attached to foo in order to get info from baz. Currently attached to bar
```

Stacktrace:
```
*** Check failure stack trace: ***
    @     0x55e43bfbab84  absl::lts_20260107::log_internal::LogMessage::SendToLog()
    @     0x55e43bfbab32  absl::lts_20260107::log_internal::LogMessage::Flush()
    @     0x55e43801f126  xls::LazyNodeData<>::GetInfo()
    @     0x55e43801e96d  xls::BddQueryEngine::GetBddNode()
    @     0x55e43bac8909  xls::BddQueryEngine::KnownValue()
    @     0x55e43800b24a  xls::(anonymous namespace)::PrintScheduleInfo()
    @     0x55e438004bb2  xls::(anonymous namespace)::RealMain()
    @     0x55e438002a75  main
    @     0x7fa9e4c29d90  (unknown)
```

A workaround was introduced with a [commit](https://github.com/antmicro/xls/commit/412ca7f4f50ea3ae6a8bfe9e4e2d6e32325a5f4b) in #3636 which checks if a node is tracked by the current function before getting the node info.

**To Reproduce**
This bug was reproduced during the latest [Test ZSTD module](https://github.com/google/xls/actions/runs/20077207368/job/57594655379) job while running `xls_benchmark_ir` targets.

**Expected behavior**
Failing targets should pass and generate the IR benchmark data.

allight:
Sorry we missed this for so long. It must have just got lost in the shuffle during the holidays.

PR #3931, opened 2026-03-10
Fix regressions in ZSTD module

magancarz:
This PR:
- reverts enabling proc-scoped channels in ZSTD module, which caused regressions in builds and decoding test cases ([77dcc8](https://github.com/google/xls/commit/77dcc81894b9d281249a64f2e448293916207aa8))
- fixes place and route target failures by specifying larger die dimensions
- adds a fix for `opt_ir_benchmark` targets by checking if a node is tracked before getting info in `lazy_query_engine.h`. (Fixes https://github.com/google/xls/issues/3637)

### DIFF
--- a/.github/workflows/modules-zstd.yml
+++ b/.github/workflows/modules-zstd.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Build and run synthesis benchmarks of the ZSTD module (opt)
         if: ${{ !cancelled() }}
         run: |
-          for target in $(bazel query 'filter("_benchmark_synth$", kind(rule, //xls/modules/zstd/...))');
+          for target in $(bazel query 'filter("_benchmark_synth$", kind(rule, //xls/modules/zstd/...)) except attr(tags, "long", //xls/modules/zstd/...)');
           do
             echo "running $target";
             bazel run -c opt $target -- --logtostderr;
--- a/xls/modules/zstd/BUILD
+++ b/xls/modules/zstd/BUILD
@@ -2737,7 +2737,7 @@ synthesize_rtl(
 benchmark_synth(
     name = "literals_decoder_benchmark_synth",
     synth_target = ":literals_decoder_synth_asap7",
-    tags = ["manual"],
+    tags = ["manual", "long"],
 )
 
 place_and_route(
@@ -3399,7 +3399,7 @@ synthesize_rtl(
 benchmark_synth(
     name = "huffman_literals_dec_benchmark_synth",
     synth_target = ":huffman_literals_dec_synth_asap7",
-    tags = ["manual"],
+    tags = ["manual", "long"],
 )
 
 place_and_route(


xls/modules/zstd/memory/BUILD
```diff
@@ -251,6 +251,8 @@ place_and_route(
     name = "remove_empty_bytes_place_and_route",
     clock_period = CLOCK_PERIOD_PS,
     core_padding_microns = 2,
+    die_height_microns = 150,
```
proppy, line 254: looks unrelated?

xls/passes/lazy_query_engine.h
```diff
@@ -188,6 +188,9 @@ class LazyQueryEngine : public internal::CacheableQueryEngine {
   const LazyNodeInfo<Info>& info() const { return qe_info(); }
 
   std::optional<SharedLeafTypeTree<Info>> GetInfo(Node* node) const {
+    if (!IsTracked(node)) {
```
proppy: looks unrelated?

magancarz: This change fixes the following IR benchmark error:
```
F0209 17:32:00.065199   37509 lazy_node_data.h:289] Check failed: f_->Contains(node) Must be attached to __fse_dec__FseDecoder_0__64_8_32_4_64_7_next in order to get info from __state_0: bits[4] = state_read(state_element=__state_0, id=4740). Currently attached to __fse_dec__FseDecoderInst_0_next
```

For further reference, see #3637 and `Build and run ZSTD IR benchmark rules` in [this CI run](https://github.com/google/xls/actions/runs/21834284359/job/63000780314), as logs from the job mentioned in the issue are expired.

xls/modules/zstd/memory/BUILD
magancarz: `remove_empty_bytes_place_and_route` target also fails with the `Number of IO pins exceeds maximum number of available positions.` error without specifying a larger die size.

proppy:
Looks like the zstd CI was cancelled after 5 hours, is that expected?

proppy:
> List of commits which introduced regressions to ZSTD module:
> - [0b3b7d](https://github.com/google/xls/commit/0b3b7dd38a5a8fd46861517740b2f30a489a7711)

it doesn't looks [0b3b7d](https://github.com/google/xls/commit/0b3b7dd38a5a8fd46861517740b2f30a489a7711) is being reverted to me, am I missing something?

magancarz:
> Looks like the zstd CI was cancelled after 5 hours, is that expected?

No, the timeout error was caused by long-running P&R targets and I think those should be tagged as `long` and skipped in CI to prevent these kind of problems, that is:
- sequence_executor_place_and_route
- command_constructor_place_and_route
- fse_dec_place_and_route
- ram_mux_place_and_route
- huffman_decoder_place_and_route
- mem_writer_place_and_route

I'll update the pull request with relevant changes.

> it doesn't looks [0b3b7d](https://github.com/google/xls/commit/0b3b7dd38a5a8fd46861517740b2f30a489a7711) is being reverted to me, am I missing something?

Sorry, that was unclear. This commit created the `Test ZSTD module` job run which showed the error caused by the `opt_ir_benchmark` targets. Previous `Test ZSTD module` run on `main`, as seen on the `Actions` page, was dispatched by [cc8501](https://github.com/google/xls/commit/cc8501f7d1d4b3cd2ce6854bed0e81a459fab685) and ended without errors. The error was introduced somewhere in between, probably by [b20719](https://github.com/google/xls/commit/b20719d124ac865dd64d592e4a87a8fdb6ec9d78) and [8bacd2](https://github.com/google/xls/pull/3931/changes/8bacd2baefc3e2f65d2a670ab0c118e9e339fba0) fixes it.

magancarz:
I've added a `long` tag to long-running benchmark synth / P&R targets and excluded them in the `Test ZSTD module` job.

.github/workflows/modules-zstd.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Build and run synthesis benchmarks of the ZSTD module (opt)
         if: ${{ !cancelled() }}
         run: |
-          for target in $(bazel query 'filter("_benchmark_synth$", kind(rule, //xls/modules/zstd/...))');
+          for target in $(bazel query 'filter("_benchmark_synth$", kind(rule, //xls/modules/zstd/...)) except attr(tags, "long", //xls/modules/zstd/...)');
```
proppy, line 66: since the "long" targets are also tagged as manual, maybe we could simply keep a single wildcard bazel run (remove the for) and let the default bazel behavior exclude them?

magancarz: As far as I know, `bazel run` allows to run a single target at once, so in the end it would still require `xargs` or `for` type invocation for all benchmark synth targets and `bazel query` includes targets with the `manual` tag by default. I believe we could make use of the default `manual` targets exclusion if we would run all test/build targets in the ZSTD module simultaneously with the wildcard pattern, just like in the [Continuous Integration job](https://github.com/google/xls/blob/main/.github/workflows/continuous-integration.yml#L48).

proppy: ah that makes sense, I forgot about this thanks!

proppy (approved):

xls/passes/lazy_query_engine.h
proppy: @ericastor @allight this is fixing https://github.com/google/xls/issues/3637 and looks consistent w/ what being done here:
https://github.com/google/xls/blob/d68fa8d8b24441049966e6fc0e55da77c4158b26/xls/passes/bdd_query_engine.cc#L300-L302

would you prefer this in a separate CL?

allight: This change is not safe and we shouldn't merge it. The actual fix is in: https://github.com/google/xls/pull/3975

Without this files' edits the rest of this CL looks fine to me.

rw1nkler: Thank you @allight for taking care of this!

magancarz: Thanks! I've dropped the commit adding these changes.

magancarz:
I've dropped the commit checking if a node is tracked before getting the info since the actual fix is in #3975.

magancarz:
I've rebased changes onto the latest main, so fix from #3975 is included.

PR #3982, opened 2026-03-22
docs: improve XLS IR execution flow explanation (modified docs_src/interpreters.md)

Waleed99i:
This PR improves the documentation of XLS IR by adding a clear explanation of the IR execution flow.

It introduces:
- A high-level overview of how IR is executed
- Differences between interpreter, JIT, and code generation paths
- A complete flow from DSLX to Verilog

This addition helps new contributors better understand how XLS transitions from software-style execution to hardware generation.

This is especially useful for contributors working on compiler optimizations and hardware-aware transformations.

### DIFF
--- a/docs_src/interpreters.md
+++ b/docs_src/interpreters.md
@@ -25,18 +25,19 @@ XLS supports multiple execution and validation mechanisms across different abstr
 
 - **IR-level Simulation**
   - Supports LLVM-based JIT compilation as well as Ahead-of-Time (AOT) compilation
-  - Includes an interpreter for functional validation of IR
   - Enables cycle-accurate simulation using XLS Block IR after code generation
   - JIT/AOT options compile to native machine code for high-performance simulation.
 
 - **Code Generation (Hardware Path)**
-  - Converts XLS IR into Verilog/SystemVerilog
-  - Enables synthesis of hardware designs
-  - Supports both combinational and pipelined circuit generation
 
-- **Netlist Simulation**
-  - Evaluates synthesized netlists using cell libraries
-  - Used for post-synthesis validation
+  - Converts XLS IR into Verilog/SystemVerilog
+  -  Enables generation of synthesizable hardware designs
+  -  Supports both combinational and pipelined circuit generation
+
+- **Simulation and Validation**
+
+  -  Generated Verilog can be simulated using external simulators (e.g., Verilator)
+  -  Supports validation at both RTL and post-synthesis (netlist) levels
 
 
 ### Why this matters ?


Waleed99i:
@google-cla please check , I have signed in for CLA 

image

ericastor:
This high-level summary seems like it belongs somewhere other than at the tail end of the detailed docs for the XLS IR JIT/AOT compiler. Have you considered finding a better home for it?

Waleed99i:
> This high-level summary seems like it belongs somewhere other than at the tail end of the detailed docs for the XLS IR JIT/AOT compiler. Have you considered finding a better home for it?

@ericastor  Thanks for the suggestion! I agree that this high-level overview doesn’t fit well at the end of the detailed JIT documentation.
Initially, I placed it there just to avoid disrupting the existing flow of the document (that was my only intention of placing it at the end ) , but I realize it would be better suited as a separate document.
I’m considering moving it to a new file (e.g., a general execution flow overview named something like ir_execution_flow.md ) or potentially integrating it into an existing high-level document if that fits better (i.e i see there's a markdown file named ir_overview.md).
I’d appreciate your feedback on the preferred location, and I’ll update the PR accordingly.

ericastor:
If you look at our documentation page (https://google.github.io/xls/), these files are what compile to that website!

Maybe the best place would be as an upgrade on the [Interpreters page](https://google.github.io/xls/interpreters/), under the Tools section; this comes from https://github.com/google/xls/blob/main/docs_src/interpreters.md. You could even consider renaming the page to "Simulation", rather than "Interpreters"!

Waleed99i:
> If you look at our documentation page (https://google.github.io/xls/), these files are what compile to that website!
> 
> Maybe the best place would be as an upgrade on the [Interpreters page](https://google.github.io/xls/interpreters/), under the Tools section; this comes from https://github.com/google/xls/blob/main/docs_src/interpreters.md. You could even consider renaming the page to "Simulation", rather than "Interpreters"!

@ericastor Thanks for the suggestion! I’ve updated the PR accordingly.

- I’ve restored `ir_jit.md` to its original state to avoid placing high-level content there.
- I’ve moved the execution flow explanation to `interpreters.md`, where it fits better in the overall context.
- I’ve also updated the page title to "Simulation and Execution" and adjusted the explanation to align with that structure.

I have not renamed the file itself yet, but I’m happy to do so if that’s preferred.

Kindly take a look and let me know your feedback. Thanks again for the guidance!

docs_src/interpreters.md
```diff
@@ -1,10 +1,55 @@
...
+  - Executes high-level DSLX code directly
+  - Useful for early-stage validation and unit testing
+
+- **IR Interpreter**
+  - Executes XLS IR without compilation
+  - Useful for debugging and correctness checking
+
+- **IR JIT (LLVM-based)**
+  - Compiles IR into native machine code
+  - Provides faster execution for performance evaluation
```
ericastor: It's probably worth mentioning here that the "JIT" also includes AOT support, and that the IR JIT & interpreter both also support cycle-accurate simulation using XLS "block IR" after codegen.

docs_src/interpreters.md
```diff
@@ -1,10 +1,55 @@
...
+  - Enables synthesis of hardware designs
+  - Supports both combinational and pipelined circuit generation
+
+- **Netlist Simulation**
+  - Evaluates synthesized netlists using cell libraries
+  - Used for post-synthesis validation
+
+### Complete Execution Flow
+
+DSLX → IR → (Interpret / JIT) → Codegen → Verilog → Netlist → Simulation
```
ericastor: This line is a little misleading; it looks like you can only execute between IR and Codegen, and like it's a required step in XLS compilation. I'm not sure this is adding much as is; maybe if it showed the compilation flow up top, with arrows pointing *down* from each possible execution point?

Waleed99i: It is showing Compilation flow basically , I just added this at the end ; you may say its a short summary ;and yes its from top to bottom . Well I can do 2 things in this regards , (1) remove it  , (2) modify it to show with top to bottom arrow . Kindly let me know what should be done .

docs_src/interpreters.md
Waleed99i: - **IR Execution (JIT / AOT / Interpreter)**
  - Supports LLVM-based JIT compilation as well as Ahead-of-Time (AOT) execution
  - Includes an interpreter for functional validation of IR
  - Enables cycle-accurate simulation using XLS Block IR after code generation
 
Is that alright ?
Please let me know if you need further changing so that I'll commit !

docs_src/interpreters.md
ericastor: Given that it literally is incorrect, in the sense that (Interpret / JIT) is not a step in the compilation flow nor an intermediate step in an execution flow - yes, this needs to either be removed or corrected.

docs_src/interpreters.md
ericastor: There's one major point of confusion in your proposed summary, and one minor one.

(major) AOT describes ahead-of-time *compilation*, not execution.

(minor) It should probably be called out that the JIT/AOT options compile to native machine code for high-performance simulation.

Waleed99i: Thanks for the clarification!
You're right  AOT refers to ahead-of-time compilation, not execution. That was my mistake in wording.
Please let me know if this looks better.

- **IR Execution (Interpreter / JIT / AOT)**
  - LLVM-based JIT compiles IR to native machine code for high-performance simulation
  - Supports LLVM-based JIT compilation as well as Ahead-of-Time (AOT) compilation
  - Includes an interpreter for functional validation of IR
  - Enables cycle-accurate simulation using XLS Block IR after code generation

docs_src/interpreters.md
Waleed99i: yes you are right , I was a bit confused thats why put parenthesis there and asked for your review , so there's a diagram which I found in documentation (https://google.github.io/xls/images/xls_stack_diagram.png) 
image
I think it would be better if make a slightly simpler version here , like the one I have made in draw.io 
image

Appreciate your point on it .

ericastor: Your simplified diagram is still misleading, as it suggests that simulation is the only endpoint. The main diagram is confusing, I admit, because it's trying to capture every possible endpoint in a single picture. We might want to have a central "everything we support" diagram, along with separate diagrams for each user journey. Those latter diagrams would still be more complex than what you've sketched - so let's leave them aside for now.

For this PR, let's just remove the line in question.

docs_src/interpreters.md
ericastor: To be honest, this is reading more and more like LLM output that doesn't quite understand the distinctions being made. I'd suggest separate information for the interpreter (for direct functional validation of IR) and for native simulation (JIT/AOT).

docs_src/interpreters.md
Waleed99i: I understand , okay , I will remove it in next commit !

docs_src/interpreters.md
Waleed99i: Based upon what you were pointed out last time , I think the best possible would be 
- Supports LLVM-based JIT compilation as well as Ahead-of-Time (AOT) compilation
- Includes an interpreter for functional validation of IR
- Enables cycle-accurate simulation using XLS Block IR after code generation
- JIT/AOT options compile to native machine code for high-performance simulation.
I think it would be better if we explain these 4 points in single sub-heading which (you) may tell what it should be . 
I am saying this actually because if you see my last commit we already have this
- **IR Interpreter**
  - Executes XLS IR without compilation
  - Useful for debugging and correctness checking
  so in my opinion only one subheading would look more professional , Or we could modify both of them. Please let me know what exactly you want so that I would commit . Thanks for your consideration by the way !

ericastor: That seems reasonable & clear - and just to be clear, *clarity* is the highest goal here.

If you want to fold these together, I'd suggest something like "IR-level Simulation" as the header.

Waleed99i: Alright ! 
I have committed again as per your final instructions , please check it out .Looking forward to let it be merged now

docs_src/interpreters.md
```diff
@@ -1,10 +1,54 @@
...
+  - Executes high-level DSLX code directly
+  - Useful for early-stage validation and unit testing
+
+- **IR Interpreter**
+  - Executes XLS IR without compilation
+  - Useful for debugging and correctness checking
+
+- **IR-level Simulation**
+  - Supports LLVM-based JIT compilation as well as Ahead-of-Time (AOT) compilation
+  - Includes an interpreter for functional validation of IR
```
ericastor: I think this line is redundant now and could be removed? (This one isn't critical, I can remove it on import.)

docs_src/interpreters.md
```diff
@@ -1,10 +1,54 @@
...
+  - Enables cycle-accurate simulation using XLS Block IR after code generation
+  - JIT/AOT options compile to native machine code for high-performance simulation.
+
+- **Code Generation (Hardware Path)**
+  - Converts XLS IR into Verilog/SystemVerilog
+  - Enables synthesis of hardware designs
+  - Supports both combinational and pipelined circuit generation
+
+- **Netlist Simulation**
```
ericastor: We should probably clarify that XLS does not *provide* a Verilog simulator (and also that this is more general than netlists, it's generally full Verilog support).

docs_src/interpreters.md
Waleed99i: You are talking about 1st two lines or just the 2nd i think ? Should I remove it?and , yep it could be removed though

docs_src/interpreters.md
Waleed99i: is that sounds good ?

**Code Generation (Hardware Path)**
- Converts XLS IR into Verilog/SystemVerilog
- Enables generation of synthesizable hardware designs
- Supports both combinational and pipelined circuit generation

**Simulation and Validation**
- Generated Verilog can be simulated using external simulators (e.g., Verilator)
- Supports validation at both RTL and post-synthesis (netlist) levels

so i did addition of word "synthesizable" to clarify what you were saying . Also slightly modifed Netlist simulation one . Moreover please let me know should i remove Verilator from there or is it fine ? 

waiting for your response to commit

ericastor: Yep, that seems good to me!

docs_src/interpreters.md
ericastor: Yes, please remove the second line that talks about the interpreter.

docs_src/interpreters.md
Waleed99i: have done with commit , please check !

ericastor:
The CI is complaining about a missing newline at the end of the file - but I can fix that on import, so no worries. I'll check that this works internally, and get a second internal OK, then I think we'll land this! Thanks for the contribution.

ericastor (approved):

PR #4021, opened 2026-03-31
1/2: Add enumerate() to DSLX stdlib

joannabrozek:
One of two PRs responsible for migrating from the built-in enumerate, which cannot be lowered to IR. 
This PR includes: adding the enumerate function to the DSLX standard library and replacing the calls.


joannabrozek:
@richmckeever, I've noticed that the CI failed on internal checks. Can you help me find which test case fails?

dplassgit:
We've addressed the failures from our side and pushed and merged the fix. Thanks for the original PR!

PR #4107, opened 2026-04-14
Avoid cloning invokes in strength reduction.

meheff:
Strength reduction will sink operations into a select if the transformation can result in constant folding or dramatic simplification. This was previously guarded by OpIsSideEffecting. This change also prevents the transformation if the operation is an invoke as the invoked function may be side-effecting.

This change fixes inf-looping of the optimizer with FFI functions. Invokes were being transformed, but the old dead invokes were not being DCE'd because dce doesn't remove invokes because they may result in side-effects. This change causes the optimizer to treat invokes as potentially side-effecting more consistently.


ericastor (approved):
Thanks!

PR #4452, opened 2026-06-23
SPIN verification (PoC)

rw1nkler:
This PR adds experimental SPIN model checker integration for XLS DSLX designs.

The core of the implementation consists of:

* `promela_generator`, which converts XLS IR into a Promela model,
* `spin_runner`, which serves as the entry point for running SPIN verification from the `interpreter_main`
* `trace_compare`, which compares execution traces produced by SPIN and the DSLX interpreter.

The integration supports two modes:

* Guided simulation - where SPIN executes a single execution path and the resulting trace is compared against the DSLX interpreter trace.
* Exhaustive simulation, where SPIN explores all possible interleavings of proc operations to detect issues such as deadlocks, livelocks, and ordering-dependent behavior.

This PR also includes Bazel rules for integrating and experimenting with SPIN verification directly in build targets.
More information about the general implementation can be found in the [README](https://github.com/antmicro/xls/blob/f45cc31a1ef93da8cbaec88771c65d544e0357ce/xls/spin/README.md).

There are also example designs in the `examples/` directory, which include demonstration of SPIN integration and show how to enable and run both guided and exhaustive verification modes.


magancarz:
I've resolved merge conflicts with latest main.

proppy:
(turning this into a draft PR since it's a PoC)

PR #4457, opened 2026-06-23
dslx: parse semantic sum declarations and constructors

dank-openai:
## Summary

- Add DSLX frontend support for semantic-sum `enum` declarations with unit, tuple, and struct payload variants.
- Add semantic-sum constructor expressions and match payload patterns so the AST preserves the source payload shape.
- Teach formatting, cloning, LSP symbols, type-info plumbing, conversion scaffolding, and diagnostics about the new AST nodes while leaving full typechecking and lowering semantics for follow-up work.
- Preserve existing numeric-enum parsing behavior, including explicitly typed empty enums.

## Example

```dslx
enum Option<T: type> {
  None,
  Some(T),
}
```

This PR makes declarations and constructors like `Option::Some(x)` parse and format as semantic-sum syntax. Later changes can build typechecking and conversion semantics on top of these AST nodes.

## Validation

Validation ran on Ubuntu 24.04 at commit `1f8ad4299fa61d5d35d2bbd4fe4780f79314f2d8`:

- `bazel test //xls/dslx/fmt:ast_fmt_test //xls/dslx/frontend:parser_test //xls/dslx/frontend:ast_cloner_test //xls/dslx/frontend:scanner_test //xls/dslx:dslx_fmt_test //xls/dslx:constexpr_evaluator_test //xls/dslx/bytecode:bytecode_emitter_test //xls/dslx/exhaustiveness:exhaustiveness_match_test //xls/dslx/lsp:document_symbols_test //xls/dslx/translators:dslx_to_verilog_test //xls/fuzzer:value_generator_test //xls/dslx/ir_convert:ir_converter_test //xls/dslx/ir_convert:ir_converter_legacy_test //xls/dslx/tests/errors:error_modules_test //xls/dslx/run_routines:run_routines_test //xls/dslx/type_system:type_info_to_proto_test //xls/dslx/type_system_v2:typecheck_module_v2_enum_test --test_output=errors --keep_going --test_timeout=900 --cache_test_results=no`

## Misc.

This is PR 1/7 for phase 1 implementation of sum types.


xls/dslx/frontend/ast.h
```diff
@@ -3225,6 +3294,147 @@ class EnumDef : public AstNode {
...
+    kUnit,
+    kTuple,
+    kStruct,
+  };
+
+  SumVariant(Module* owner, Span span, NameDef* name_def,
+             PayloadShape payload_shape,
+             std::vector<TypeAnnotation*> tuple_members,
+             std::vector<StructMemberNode*> struct_members,
+             Expr* discriminant = nullptr,
```
richmckeever: The usual convention in here is to use std::optional<T*> to indicate a nullable T* and force users of it to check if it has a value.

xls/dslx/frontend/parser.cc
```diff
@@ -2878,6 +2990,13 @@ absl::StatusOr<Expr*> Parser::ParseTermRhs(Expr* lhs, Bindings& outer_bindings,
         goto done;
       }
 
+      if (IsLocalSumConstructorRef(lhs, outer_bindings)) {
```
richmckeever: Can we make the condition for this more generic or defer the check to sometime later than parsing? Resolving references generally is not something the parser is well equipped for, and the direction is to be less equipped for it.

xls/dslx/fmt/ast_fmt_test.cc
```diff
@@ -1646,6 +1646,146 @@ SECOND = 1,
...
+    EmptyTuple(),
+    EmptyStruct { },
+    Some(u32),
+    Point { x: u32 },
+}
+
+fn f(x: Option) -> Option {
+    match x {
+        Option::EmptyTuple() => Option::EmptyTuple(),
+        Option::EmptyStruct { } => Option::EmptyStruct {},
```
richmckeever: It seems like it would be less surprising to have no space between the empty braces on both sides here (and in the declaration of the enum).

xls/dslx/frontend/ast.h
```diff
@@ -2682,6 +2686,71 @@ class MatchArm : public AstNode {
...
+//
+//   Option::Some(value)
+//   Message::Point { x: px, y: py }
+//
+// The constructor itself is always spelled as a `ColonRef`.
+//
+// `payload_shape()` is the source of truth for tuple-vs-struct spelling. Empty
+// child vectors are valid for both `Case()` and `Case { }`, so callers must
+// not recover the shape from vector emptiness alone.
+class SumVariantPayloadPattern : public AstNode {
```
richmckeever, line 2731: A few observations about this:
* This class itself is an ad hoc sum type in C++(one for tuple + one for struct). It's an element in what is already a more standard sum type, NameDefTree::Leaf.
* It would be nice to foresee how we'd use struct patterns with regular structs. Having a struct pattern thing that is not married to sum types would solve that.
* This is actually a subtree, and breaks the concept of what is a "leaf" in NameDefTree. I've always thought NameDefTree was unintuitive anyway.

Maybe we should do away with the NameDefTree class and just define PatternTree = std::variant<TuplePattern*, StructPattern*, NameDef*, NameRef*, etc. from what is now Leaf>. We replace NameDefTree uses with PatternTree. We split your proposed SumVariantPayloadPattern into TuplePattern and StructPattern. TuplePattern's constructor ref is nullable. StructPattern's can later be a ColonRef | TRTA, and we can use it for regular structs then.

richmckeever (changes requested):

xls/dslx/frontend/ast.h
dank-openai: Done - `SumVariant` now stores and returns `std::optional<Expr*>`, and its callers explicitly handle absent discriminants.

xls/dslx/frontend/parser.cc
dank-openai: Done - I removed parser-side constructor-reference resolution. Constructor-level explicit parametrics are rejected during semantic typechecking after resolution, with regression coverage for local and imported constructors.

xls/dslx/fmt/ast_fmt_test.cc
dank-openai: Done - empty struct payloads now consistently render as `{}` in declarations, constructor patterns, and constructor expressions.

xls/dslx/frontend/ast.h
dank-openai: Done - the underlying pattern-tree refactor landed in https://github.com/google/xls/pull/4614.

richmckeever: I think the idea was to now have this PR add a StructPattern in the PatternTree variant alongside TuplePattern, so that the choice between tuple and struct does not live in StructVariantPayloadPattern per se. Would that make sense?

dank-openai: Done - `PatternTree` now contains both `TuplePattern` and `StructPattern`. `SumVariantPayloadPattern` only carries the constructor and its concrete payload pattern, so `Choice::Point { x: value }` becomes `Choice::Point` plus `StructPattern { x: value }`. Recursive constructor patterns are no longer treated as leaves.

xls/dslx/type_system_v2/validate_concrete_type.cc
```diff
@@ -57,6 +57,22 @@
 namespace xls::dslx {
 namespace {
 
+bool ContainsConstructorPattern(const PatternTree& pattern) {
+  bool contains_constructor = false;
```
richmckeever: Nit: just return true upon discovery of any constructor pattern.

xls/dslx/type_system_v2/validate_concrete_type.cc
```diff
@@ -378,12 +394,22 @@ class TypeValidator : public AstNodeVisitorWithDefault {
...
     for (MatchArm* arm : node->arms()) {
       for (const PatternTree& pattern : arm->patterns()) {
-        bool exhaustive_before = exhaustiveness_checker.IsExhaustive();
-        exhaustiveness_checker.AddPattern(pattern);
-        if (exhaustive_before) {
-          warning_collector_.Add(
-              GetPatternSpan(pattern), WarningKind::kAlreadyExhaustiveMatch,
-              "Match is already exhaustive before this pattern");
+        const bool has_constructor_pattern =
+            ContainsConstructorPattern(pattern);
```
richmckeever: Nit: inline the call into the if-condition and unnest the else since there is a return.

richmckeever (approved):

xls/dslx/type_system_v2/validate_concrete_type.cc
dank-openai: done

xls/dslx/type_system_v2/validate_concrete_type.cc
dank-openai: done

richmckeever (approved):

PR #4474, opened 2026-06-26
Make sure .bazelrc configures the whole c++20 options

hzeller:
The c++20 options were removed in the .bazelrc in favor of configuring it in the llvm toolchain configured in the MODULE.bazel.

https://github.com/google/xls/commit/17525ebe11980848e0e74f875d9e7b508e17224a

...however, this only works if the toolchain can be used. Since it is not hermetic, it needs to be disabled on many platforms. So the original way of c++20 config was partially re-instantiated in

https://github.com/google/xls/commit/41f6104f2b0fc91c3a6ae22d299859cff51da3cc

Turns out, also the `BAZEL_CXXOPTS=-std=c++20` needs to be set again. This is happening in this PR.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -35,6 +35,7 @@ build:macos --macos_minimum_os=10.13
 # Minimum c++ standard used. We set it in the llvm toolchain, but
 # we also need to specify it here when it is not possible to use that toolchain.
 build --cxxopt "-std=c++20"            --host_cxxopt "-std=c++20"
+build --client_env=BAZEL_CXXOPTS=-std=c++20
 
 # Depending on the installation, clang or clang-tidy need to be told that
 # a c++ file is, indeed, containing c++ (bazel just invokes clang, not clang++)


hzeller:
the CI issue looks unrelated.

dplassgit:
What does the failure look like so I know if this PR fixes it?

hzeller:
Well, for users of the toolchain in the MODULE.bazel, things work, but for everyone else who always has to manually comment it out because of its non-hermeticity and run with local compiler sees things like these:

```
/nix/store/qxaq7jz61a6zkr2mq49i0zvqip2m2jj8-gcc-15.2.0/include/c++/15.2.0/type_traits:533:12: note: candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 1 was provided
In file included from bazel-out/k8-opt/bin/xls/dslx/cpp_transpiler/_gen_test_types_lib_srcs/test_types_lib4.cc:8:
In file included from ./bazel-out/k8-opt/bin/xls/dslx/cpp_transpiler/test_types_lib.h:12:
In file included from bazel-out/k8-opt/bin/xls/dslx/cpp_transpiler/test_types_dep_lib.h:13:
In file included from ./xls/public/value.h:29:
In file included from ./xls/ir/value_view.h:38:
In file included from ./xls/ir/package.h:31:
./xls/ir/channel.h:75:52: warning: defaulted comparison operators are a C++20 extension [-Wc++20-extensions]
   75 |   bool operator==(const FifoConfig& other) const = default;
      |                                                    ^
./xls/ir/channel.h:76:16: warning: '<=>' is a single token in C++20; add a space to avoid a change in behavior [-Wc++20-compat]
   76 |   bool operator<=>(const FifoConfig& other) const = default;
      |                ^ 
      |                   
./xls/ir/channel.h:76:8: error: 'operator<=' cannot be the name of a variable or data member
   76 |   bool operator<=>(const FifoConfig& other) const = default;
      |        ^
./xls/ir/channel.h:76:18: error: expected ';' at end of declaration list
   76 |   bool operator<=>(const FifoConfig& other) const = default;
      |                  ^
      |                  ;
./xls/ir/channel.h:58:51: error: constexpr constructor's 5th parameter type 'std::optional<std::string>' (aka 'optional<basic_string<char>>') is not a literal type
   58 |                        std::optional<std::string> fifo_wrapper = std::nullopt)
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
/nix/store/qxaq7jz61a6zkr2mq49i0zvqip2m2jj8-gcc-15.2.0/include/c++/15.2.0/optional:779:7: note: 'optional<std::basic_string<char>>' is not literal because it has base class '_Optional_base<basic_string<char>>' of non-literal type
  779 |     : private _Optional_base<_Tp>,
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

... errors that indicate that the compiler didn't think it compiled in c++20.

dplassgit:
"Well, for users of the toolchain in the MODULE.bazel, things work, but for everyone else who always has to manually comment it out because of its non-hermeticity and run with local compiler sees things like these:"

Should we even allow this? I know that's a different question, though.

dplassgit (approved):

hzeller:
unfortunately it is necessary. The toolchain we sideload is pretty annoying and only works on some systems. Accidentally.

I always have to comment out the toolchain as it is not working on an actual hermetic system. XLS is essentially impossible to compile outside the very narrow Ubuntu band.

PR #4521, opened 2026-07-07
Fix unnecessary parse and typecheck overhead when using IR conversion check

magancarz:
This pull request fixes unnecessary overhead introduced by duplicate parsing and typechecking when IR conversion check is enabled. As tested with `bazel run -c opt //xls/modules/zstd:zstd_dec_dslx_test` and commented tests for simplicity of measuring IR conversion check overhead - on `main` branch it takes ~3.5 min to finish executing the target, while with this PR changes it finishes in ~38 seconds. Without IR conversion check, the execution time is ~27 seconds. As a result it narrows down IR conversion check overhead from ~3 min to ~11 seconds in the tested case.

### DIFF
--- a/xls/dslx/interpreter_main.cc
+++ b/xls/dslx/interpreter_main.cc
@@ -186,10 +186,10 @@ absl::StatusOr<TestResult> RealMain(
                                    : TypeInferenceVersion::kVersion1)
           : std::nullopt;
 
-  RealFilesystem vfs;
+  auto vfs = std::make_unique<RealFilesystem>();
 
   XLS_ASSIGN_OR_RETURN(std::string program,
-                       vfs.GetFileContents(entry_module_path));
+                       vfs->GetFileContents(entry_module_path));
   XLS_ASSIGN_OR_RETURN(std::string module_name, PathToName(entry_module_path));
 
   std::unique_ptr<AbstractRunComparator> run_comparator;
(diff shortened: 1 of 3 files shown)


magancarz:
@richmckeever I've answered your question about used compilation mode by updating the description with used command.

xls/dslx/interpreter_main.cc
```diff
@@ -292,58 +299,46 @@ absl::StatusOr<TestResult> RealMain(
...
+    ParametricEnv stub_parametric_env;
     std::vector<std::string> failed_ir_conversion_entries;
     for (std::string& elem : module_elements) {
       // Convert to IR each element separately.
-      absl::StatusOr<PackageConversionData> ir_conv_result =
-          ConvertFilesToPackage(module_path, dslx_stdlib_path.string(),
-                                dslx_paths, ir_convert_options, elem,
-                                module_name, &printed_error);
+      absl::StatusOr<std::string> ir_conv_result =
+          ConvertOneFunction(tm.module, elem, &import_data,
```
dplassgit, line 340: Is this change the whole reason it's faster? I'm having trouble wrapping my head around the immense increase in speed with this change.

magancarz: It speeds up the IR conversion check, because [ConvertOneFunction](https://github.com/google/xls/blob/e6e6363879a45e75ef80db6adf476395196abdcb/xls/dslx/ir_convert/ir_converter.cc#L582) gives a possibility to do module typechecking just once by performing it earlier and passing it with `tm.module`, while [ConvertFilesToPackage](https://github.com/google/xls/blob/e6e6363879a45e75ef80db6adf476395196abdcb/xls/dslx/ir_convert/ir_converter.cc#L670) call did module typechecking separately for each IR conversion check entry in [ir_converter.cc#L641](https://github.com/google/xls/blob/e6e6363879a45e75ef80db6adf476395196abdcb/xls/dslx/ir_convert/ir_converter.cc#L641).

dplassgit (approved):
@richmckeever do you have any thoughts? It looks good to me

PR #4568, opened 2026-07-14
Switch to hermetic toolchain

### LINKED ISSUE #4566: Switch to a hermetic toolchain
We're using the `@llvm_toolchain//:all`, which has been a trouble ever since, because it is the opposite of hermetic (it cuts toolchain to host dependencies at the worst possible spot). It only narrowly worked for Ubuntu and maybe a few select other systems.

It downloads some binaries from the llvm project for a narrow set of operating systems. It moves the contract from 'bazel invokes binary on local system' to 'bazel invokes binary that expects certain shared objects on the system', making it worse than hermetic. We have been bitten many times with this (libxml, libtinfo...).

Also, with all that, it then relied on the local c++ standard libraries and includes of the system - often times if gcc is installed, these headers not always work with llvm with random issues such as different iterator contracts etc.

_I never was able to compile XLS at home due to that, and thus only had very limited way to contribute outside machines at work._

There is an actually hermetic toolchain now: https://github.com/hermeticbuild/hermetic-llvm
It also distributes binaries, but they are fully statically linked using musl, also it comes with its on libc++.

OpenROAD previously also used the `@llvm_toolchain//` we currently use in XLS and it always was as painful as XLS is right now. Now they [recently switched to the hermetic](https://github.com/The-OpenROAD-Project/OpenROAD/pull/10812), and it is a breeze of air. It allowed me, for the first time, to compile that project out of the box on my personal machines without having to disable the toolchain. It also works on my RBE set-up at home.

So I strongly suggest we switch to it as well.

I am happy to do the work, just need the nod of approval :)

hzeller:
* Switch to using https://github.com/hermeticbuild/hermetic-llvm
 * Update rules_cc to latest (we don't need the pinned old version anymore)
 * Pin some tools to a particular version known to work with the toolchain.

Fixes: #4566

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,17 +4,9 @@ module(
 )
 
 # Compiler toolchain
-bazel_dep(name = "toolchains_llvm", version = "1.4.0")
+bazel_dep(name = "llvm", version = "0.8.11", dev_dependency = True)
 
-# Configure and register the toolchain.
-llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
-llvm.toolchain(
-    cxx_standard = {"": "c++20"},
-    llvm_version = "19.1.7",
-)
-use_repo(llvm, "llvm_toolchain")
-
-register_toolchains("@llvm_toolchain//:all")
+register_toolchains("@llvm//toolchain:all")
 
 # Python toolchain
 bazel_dep(name = "rules_python", version = "1.9.0")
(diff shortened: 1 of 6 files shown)


hzeller:
(Draft: WIP)

hzeller:
that issue is in openroad tcl and patched there, so let's wait first until #4567 is merged.

hzeller:
ok, looks like the local tcl patch in the OpenROAD change does not make it here. I am going to fix that upstream in BCR first.

hzeller:
BCR tcl patch in https://github.com/bazelbuild/bazel-central-registry/pull/9678  (awaiting merge)

hzeller:
Looks good, continuation on #4597

hzeller:
Closing this experiment, the final change has been merged.

PR #4584, opened 2026-07-14
Fix flags set in transpiler compilation.

hzeller:
Files compiled with the transpiler rule were
not compiled with c++ flags (e.g. all the flags such as --cxxopt in .bazelrc), which results in them not picking up e.g. -std=c++20 flags (which will be important in the hermetic compiler; in the old compiler this was forced by setting it in a different place).

So e.g. the following never received all cxx flags:

```
bazel build -c opt -s //xls/dslx/cpp_transpiler:test_types_lib
```

Solution: the previously named `_gen_test_types_lib_srcs` subdirectory is named now `_gen_test_types_lib_srcs.cc` which helps bazels flag matcher to choose the correct compiler and flags.

### DIFF
--- a/xls/build_rules/xls_dslx_rules.bzl
+++ b/xls/build_rules/xls_dslx_rules.bzl
@@ -520,7 +520,7 @@ def _xls_dslx_generate_cpp_type_files_impl(ctx):
     dslx_path = ":${PWD}:" + ctx.genfiles_dir.path + ":" + ctx.bin_dir.path + dslx_srcs_wsroot_path + wsroot_dslx_path
 
     # Make arguments for the cpp_transpiler tool.
-    cc_source_dir = ctx.actions.declare_directory(ctx.label.name + "_srcs")
+    cc_source_dir = ctx.actions.declare_directory(ctx.label.name + "_srcs.cc")
     h_file = ctx.outputs.header_file
 
     include_header_files = []
--- a/xls/build_rules/xls_dslx_rules.bzl
+++ b/xls/build_rules/xls_dslx_rules.bzl
@@ -520,6 +520,8 @@ def _xls_dslx_generate_cpp_type_files_impl(ctx):
     dslx_path = ":${PWD}:" + ctx.genfiles_dir.path + ":" + ctx.bin_dir.path + dslx_srcs_wsroot_path + wsroot_dslx_path
 
     # Make arguments for the cpp_transpiler tool.
+    # Add an extension to the directory that helps bazel classify the contents
+    # as C++, otherwise it will not append configured --cxxopt for compilation.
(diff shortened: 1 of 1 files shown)


hzeller:
(thanks antigravity for the debug help)

hzeller:
Working towards #4566

mikex-oss:
> Solution: the previously named _gen_test_types_lib_srcs subdirectory is named now _gen_test_types_lib_srcs.cc which helps bazels flag matcher to choose the correct compiler and flags.

Is there something documenting why a directory named `.cc` is helpful for Bazel? I'm not sure I understand how this works.

hzeller:
So Antigravity claims that it puts together from the fact that the toolchain running on a tree artifact does not have much to work with but the name of the directory to then choose what applies So it does not seem to be directly documented, but a combination of these circumstances.

(quote antigravity)

This behavior stems from how Bazel's C++ rules (`rules_cc`) map inputs to specific compilation actions and how it handles **Tree Artifacts** (opaque directories used as a single input/output).

While there isn't a single "warning: name your directories with `.cc`" page in the documentation, it is the result of several documented design choices in Bazel:

1. **Tree Artifacts are Opaque:** In Bazel, a Tree Artifact (declared via `ctx.actions.declare_directory`) is treated as a single entity. Bazel does not inherently know the contents of this directory at analysis time. Therefore, it relies entirely on the **name of the directory itself** to figure out how to process it when passed to rules like `cc_library`. 
*(See: [Bazel Documentation on Tree Artifacts](https://bazel.build/extending/rules#declaring_outputs))*

2. **Action Configurations use File Extensions:** Bazel's C++ toolchain (`cc_toolchain`) uses a strict mapping of file extensions to determine whether to launch a C compile action or a C++ compile action (`c++-compile`). C++ extensions are explicitly documented as `.cc`, `.cpp`, `.cxx`, `.c++`, and `.C`. 
*(See: [Bazel Documentation on C++ Toolchain Configuration](https://bazel.build/tutorials/cc-toolchain-config))*
[...]

By adding `.cc` to the directory name (`_gen_test_types_lib_srcs.cc`), we tricked the toolchain into classifying the entire Tree Artifact as a C++ input, which caused it to correctly wire up the `c++-compile` action and inject your `-std=c++20` flags!

(endquote antigravity)

mikex-oss:
Sounds good. Can you add a comment explaining that this is load bearing?

hzeller:
> Sounds good. Can you add a comment explaining that this is load bearing?

Done.

mikex-oss (approved):

hzeller:
Updated PR with related bazel issue. Probabaly needs re-approval @mikex-oss

mikex-oss (approved):

PR #4602, opened 2026-07-17
Fix most warnings in compile (build gardening :))

hzeller:
There are many hundred warnings in the build, most coming from the same places, but since they are in headers they show up many times.

Fixes

   * Add an override/final to a overriding virtual method where that was missing.
   * Fix field order in the constructor initialization to match member sequence.
   * Don't silently ignore absl::Status, instead wrap them in CHECK_OK() to make sure the nodiscard value makes itself known on issues.


hzeller:
Passing on to next build-gardener @erinzmoore

erinzmoore (approved):

Issue #3983, opened 2026-03-23
[Proposal] Add Matrix Multiplication (MatMul) as a Tensor-Level Abstraction in XLS IR


Waleed99i:
## Motivation

XLS currently operates primarily at a scalar level, where computations are expressed as fine-grained operations in IR. While this design is powerful for general computation, it is not optimal for emerging machine learning workloads, which rely heavily on tensor-level operations such as matrix multiplication.

Introducing a matrix multiplication (MatMul) abstraction at the IR level would allow XLS to better support hardware-efficient generation of ML accelerators.

---

## Problem

Currently, matrix multiplication must be manually expressed as a combination of scalar operations, which:

- Increases IR complexity
- Limits optimization opportunities
- Prevents the compiler from making tensor-level scheduling and architectural decisions
- Leads to suboptimal hardware in terms of area, timing, and power (PPA)

---

## Proposed Solution

I propose introducing a `MatMul` (or equivalent) tensor-level operation in XLS IR, along with multiple lowering strategies.

### 1. IR-Level Abstraction

- Define a `MatMul` operation in IR
- Represent matrix multiplication as a first-class operation instead of scalar expansion

---

### 2. Lowering Strategies

Implement multiple lowering paths:

#### (a) Combinational Expansion
- Fully unrolled scalar multiplications
- High area, low latency

#### (b) Pipelined MAC Tree
- Reduces critical path
- Improves timing performance

#### (c) Systolic Array Mapping (if possible)
- Hardware-efficient for ML workloads
- Enables scalable parallelism

---

### 3. Integration with Scheduling

- Allow the XLS scheduler to exploit MatMul-level knowledge
- Improve pipeline balancing and resource sharing

---

### 4. OpenROAD Integration

After RTL generation, use OpenROAD flow to:

- Evaluate Area
- Evaluate Timing
- Evaluate Power
- Evaluate Routability

This enables a feedback loop:

Compiler → RTL → Physical Design → PPA Metrics → Compiler Optimization

---

## Expected Outcomes

- Introduction of tensor-level abstraction in XLS
- Multiple hardware-efficient lowering strategies
- Improved PPA compared to scalar-expanded implementations
- A reproducible flow combining compiler + physical design tools

---

## Why this matters

This work bridges the gap between:

- Compiler design
- Hardware generation
- Physical design evaluation

It aligns with modern ML accelerator design trends and enables XLS to generate optimized hardware for real-world workloads.

---

## Status

[...]

PR #4634, opened 2026-07-23
Extending standard library to include ML-relevant operations

achowd32:
Hi, I'm a first time open source contributor looking to get involved and add something to XLS. After reading the [XLS ideas page](https://google.github.io/xls/ideas_and_projects/) and going through issues such as [this one](https://github.com/google/xls/issues/3983), I decided to start by making an extension to the standard library to include more ML-relevant operations (which are becoming increasingly important in the hardware domain). This work has been overseen by @msaligane, who is a visiting faculty at Google.

I wanted to keep the pull requests fairly atomic, so I started with just two operations: ReLU (which is a pure fn) and a weight-stationary systolic array (which is a proc). I have more operations I am working on (e.g. approximated Softmax, output-stationary systolic array, etc) which I would be happy to contribute if reviewers are satisfied with this work, alongside appropriate documentation. I could pivot to a related idea in the standard library if the XLS team would prefer (e.g. a quantized floating point format).

I would like to note that the code for the systolic array here is different to the one provided in examples [here](https://github.com/google/xls/blob/main/xls/examples/matmul_4x4/matmul_4x4.x). I found that my code got improved compilation times for the same architecture (details below). I am happy to gather PPA stats if it is deemed important.
| | IR Generation Time | IR Optimization Time | Code Generation Time | Verilog LoC |
| --- | --- | --- | --- | --- |
| My code | 3.659 s | 137.033 s | 30.124 s | 207498 |
| Example code | 6.945 s | 150.122 s | 40.654 s | 285690 |

The above times were averaged on 5 runs on a 16x16 systolic array.

I would sincerely appreciate all feedback, as I am trying to learn and would love to make this work.

P. S. I seem to have found a bug in the DSLX formatter wherein comments cannot be placed between multi-line arrays. There doesn't seem to be an open issue for this, so I may file one later.

### DIFF
--- a/xls/modules/ml/BUILD
+++ b/xls/modules/ml/BUILD
@@ -1,0 +1,68 @@
+# Copyright 2026 The XLS Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Machine-learning primitives for DSLX (float32).
+
+load(
+    "//xls/build_rules:xls_build_defs.bzl",
+    "xls_dslx_fmt_test",
+    "xls_dslx_library",
+    "xls_dslx_test",
+)
+
+package(
+    default_applicable_licenses = ["//:license"],
+    default_visibility = ["//xls:xls_internal"],
+    features = [
+        "layering_check",
+        "parse_headers",
+    ],
+    licenses = ["notice"],
+)
+
+xls_dslx_library(
+    name = "activation_dslx",
+    srcs = ["activation.x"],
+)
+
+xls_dslx_fmt_test(
+    name = "activation_dslx_fmt_test",
+    src = "activation.x",
+    opportunistic_postcondition = True,
+)
+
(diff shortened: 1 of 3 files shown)


allight:
@proppy @richmckeever Maybe one of you can take a look at this.

Given the many latency/throughput/implementation tradeoffs inherent in any matrix multiplier design I'm not sure we want to promote this to stdlib status but this seems like an excellent addition to the xls examples directory as the sort of flexibility DSLX allows a designer. Or maybe adding an xls/contrib/ directory for an ML dslx library might be worthwhile.

achowd32:
@allight I would be happy to add and contribute to an xls/contrib/ directory if you deem it ok. I could perhaps also focus more on operations with fewer inherent tradeoffs, like activation functions (e.g. approximated sigmoid) or approximated softmax.

If you'd rather not do the above, is there something similar I could contribute to stdlib/ or contrib/? Potentially a quantized FP library?

(We can add the systolic array implementation in the PR to examples/ if you don't think it fits elsewhere).

dplassgit (changes requested):
We've discussed this some more internally and feel it would be best to move this code to a new directory in xls/modules. We also acknowledge that we don't have published criteria regarding what belongs in stdlib, so we've opened an internal bug to track that. Thanks for the contribution and your patience!

achowd32:
I split the file into activation.x and ws_matmul.x and moved both into `xls/modules/ml`, along with a small BUILD file. Some formatting changes were made since the formatter now allows comments in arrays. Otherwise everything is the same.

Note for transparency: I used AI to help make these revisions, since they primarily involved restructuring and reformatting. I confirmed that no semantic changes were made, and all tests pass and the formatter does not error.

I have some more files that I will drop into this module directory soon.

dplassgit (approved):

Issue #1239, opened 2023-12-26
codegen should generate systemverilog struct


proppy:
Currently we flatten all struct into a single bus, ex:
```
struct Operand {
  a: u8,
  b: u8,
  c: u8,
}

fn muladd(op: Operand) -> u8 {}
```
will generate:
```
module user_module(
  input wire clk,
  input wire [23:0] op,
  output wire [7:0] out
);
```

It would be nice if instead we generated the corresponding SystemVerilog struct construct as defined in section 7.2.1 of the [1800-2017 standard](https://ieeexplore.ieee.org/document/8299595):

> 7.2.1 Packed structures
> A packed structure is a mechanism for subdividing a vector into subfields, which can be conveniently
> accessed as members. Consequently, a packed structure consists of bit fields, which are packed together in
> memory without gaps. An unpacked structure has an implementation-dependent packing, normally
> matching the C compiler. A packed structure differs from an unpacked structure in that, when a packed
> structure appears as a primary, it shall be treated as a single vector.

proppy:
Note this might require additional tagging in the IR as it looks like struct are currently flattend as tuple:
```
top fn __user_module__muladd(op: (bits[8], bits[8], bits[8])) -> bits[8] {
  op_a: bits[8] = tuple_index(op, index=0, id=2, pos=[(0,9,6)])
  op_b: bits[8] = tuple_index(op, index=1, id=3, pos=[(0,9,13)])
  op_c: bits[8] = tuple_index(op, index=2, id=5, pos=[(0,9,20)])
}
```

cdleary:
See also https://github.com/google/xls/issues/195 and https://github.com/google/xls/issues/394 (may be a dupe of the latter)

rw1nkler:
It would be useful to have the ability to dump struct members as separate ports when using Verilog as an output format

mikex-oss:
> It would be nice if instead we generated the corresponding SystemVerilog struct construct as defined in section 7.2.1 of the [1800-2017 standard](https://ieeexplore.ieee.org/document/8299595) and preserve the fields names for improved readibility:
> 
> ...
>
> ```
> typedef struct packed {
>   wire [7:0] a;
>   wire [7:0] b;
>   wire [7:0] c;
> } Operand;
> 
> module user_module(
>   input Operand [23:0] op;
>   output wire [7:0] out;
> );
> ```

[...]

PR #4668, opened 2026-07-28
Improve conversion of DSLX structs to SystemVerilog

magancarz:
This pull request proposes an automatic way of converting DSLX structs to SystemVerilog by adding a struct type in IR. Using structs in SystemVerilog can be used to improve debugging experience of the codegen result and discussion about this topic was introduced in #1239.

If a channel uses a struct type in DSLX, then it has its counterpart type assigned when converting to IR, so that codegen pass can determine whether to use corresponding SystemVerilog struct type for channel IO. To preserve compatibility with current IR implementation which uses tuples as flattened struct types, each struct type can be also treated as a tuple type whenever there is a `receive` or a `send` operation on a channel using structs or a `tuple_index` on the struct type. Struct type is then used in `channel_to_port_io_lowering_pass` to set proper `system_verilog_type` to input/output ports, which then is used in `module_builder` to replace a flattened bits type with relevant type ref. Example IR with struct types:

```
package structs

struct Request {
  id: bits[32]
  a: bits[32]
  b: bits[32]
}

top proc Sender<_req_r: Request in, _resp_s: bits[32] out>() {
  chan_interface _req_r(direction=receive, kind=streaming, strictness=proven_mutually_exclusive, flow_control=ready_valid, flop_kind=none)
  chan_interface _resp_s(direction=send, kind=streaming, strictness=proven_mutually_exclusive, flow_control=ready_valid, flop_kind=none)
  after_all.5: token = after_all(id=5)
  receive.21: (token, Request) = receive(after_all.5, channel=_req_r, id=21)
  req: Request = tuple_index(receive.21, index=1, id=9, pos=[(0,31,18)])
  req_a: bits[32] = tuple_index(req, index=1, id=10, pos=[(0,32,29)])
  req_b: bits[32] = tuple_index(req, index=2, id=11, pos=[(0,32,37)])
  tok: token = tuple_index(receive.21, index=0, id=8, pos=[(0,31,13)])
  add.12: bits[32] = add(req_a, req_b, id=12, pos=[(0,32,29)])
  send.22: token = send(tok, add.12, channel=_resp_s, id=22)
}
```

DSLX structs could already be converted and used in SystemVerilog with the following workflow:
- Use `dslx_to_verilog_main` to generate SystemVerilog from DSLX types.
- Add `sv_type` attribute to DSLX types user want to use in SystemVerilog.
- Pass IR interface proto with channel data about which `sv_type`s are used to `codegen_main`.
- Ensure `emit_sv_types` flag is set to `True` in `codegen_main`.

Automatic generation of structs to SystemVerilog can be tested by building `structs_verilog` target added with draft changes:

```
bazel build //xls/examples:structs_verilog
```

It will generate a `bazel-bin/xls/examples/structs.sv` file with struct definitions and `Sender` proc's codegen result with struct types used in channel ports.

I've also added a commit [1178df45](https://github.com/antmicro/xls/commit/1178df4533a9e799191ada081751a02ca82741fe) which fixes handling of `sv_type` attribute with input/output ports in codegen as copying the system verilog type value was omitted in the `Block::Clone()` function.

TODO:
- [ ]  Allow generation of SystemVerilog structs to a separate file like `dslx_to_verilog_main` does.
- [ ]  Make automatic conversion optional through a flag.
- [ ]  Add `unpacked` type as currently structs are [`packed` by default](https://github.com/google/xls/blob/d0a7f95d1ba99d32c9f2f1c7baa0f7ab69c5887f/xls/codegen/vast/vast.cc#L1846).
- [ ]  Rethink common handling of tuple and struct types in IR.


magancarz:
I've opened a separate pull request with fix for omitting SystemVerilog type when cloning an IR block in #4696.

ericastor:
Just to provide an update: we're trying to come up with a process internally for handling larger IR proposals. Expect an update on that process in the next week or two.

The main preview I can give is that, in most cases, we'll generally be preferring design discussion to come before sending code changes for review. We want to make sure we're aligned on *what* a proposal contains before folks dive into implementation, so we're not wasting anyone's time.

Issue #4638, opened 2026-07-24
Formatter errors when comments are placed alongside multi-line arrays


achowd32:
**Describe the bug**
As the title says, this is a fairly simple bug where the formatter will error on a DSLX file which has comments placed alongside multi-line arrays. I encountered this bug when writing [this](https://github.com/google/xls/pull/4634) PR and thought it might be good to file an issue for it.

**To Reproduce**
1. Write a DSLX file with comments placed on lines with multi-line arrays. A simple example might look as follows:
```
// example.x
fn make_arr(a: u32, b: u32, c: u32, d: u32) -> u32[2][2] {
   [[a, b], // comment here
    [c, d]] // comment here
}
```
2. Locate the path to your DSLX formatter binary. Let's say mine is simply `./dslx_fmt`. Run the formatter on the example file above, or another like it: `./dslx_fmt example.x`.
3. You should get the following error:
```
Error: INTERNAL: Formatting was skipped because a comment at example.x:2:14-3:1 would be deleted by the *this: // comment here
This is probably due to a bug (which may not have been reported yet). To complete formatting, try moving the comment to a different line.
```
**Expected behavior**
I would expect the formatter not to error on this. I have no strong preference on how the above example would be formatted; I would probably expect it to stay as is, though longer comments may be moved to their own lines.

**Environment (this can be helpful for troubleshooting):**
I encountered the above bug on Ubuntu 24.04.3.

**Additional context**
I am an open-source newcomer looking to get involved with the XLS project, and would love to try and tackle this issue if maintainers agree it is indeed undesired behavior. I have spent some time looking at `xls/dslx/fmt/ast_fmt.cc` and have some understanding of the bug. I would probably attempt to change `FormatArray` to `FormatArrayWithoutComments`, and add a new function `FormatArray` which is able to handle comments between multi-line arrays. This mirrors the structure used by `FormatTuple` and `FormatTupleWithoutComments`.

Let me know what you think, I am open to any and all feedback.

mikex-oss:
To be clear, the error itself is WAI (previously, it would just delete the comments as part of formatting). It lets you know that this is not supported yet.

But if you can get comments within arrays to work, then I think that would be great.

PR #4675, opened 2026-07-28
dslx/fmt: formatter now allows comments inside multi-line array literals

achowd32:
This is an attempt to close the issue that I filed [here](https://github.com/google/xls/issues/4638). The formatter no longer errors when given multi-line array literals containing comments. A couple of notes:

1. Unlike tuples containing comments, arrays that contain comments still maintain "clumping" behavior from before - they fit as many elements on a line as they can, until a comment is encountered. So
```
[ 
    1,
    2, 
    // comment
    3, 
] 
```
formats as
```
[ 
    1, 2, 
    // comment
    3, 
] 
```
2. Another difference between FormatArray and FormatTuple is that FormatArray can distinguish between comments that are at the end of a line (e.g. `2, 3, // comment`) versus those that are on their own line. The helper function `FirstCommentTrails` is useful not only because it is called multiple times in FormatArray, but might be useful to resolving the TODO on line 1530 within FormatTuple, which I would be interested in resolving.
3. The only issue that I've identified with this commit as it stands is that ellipsis can sometimes be formatted a little weirdly. Namely, when there is an ellipsis immediately following a comment, the ellipsis is moved before the comment. So 
```
[ 
    1,
    2, 
    // comment
    ...
] 
```
formats as 
```
[ 
    1, 2, ...
    // comment
] 
```
I am happy to look into this in the future as well, but thought it was worth filing the PR as is since this is already a significant improvement.

I'm a newcomer to XLS and open source generally, so please let me know if there is anything I can improve on! I would love to make more contributions.

### DIFF
--- a/xls/dslx/fmt/ast_fmt.cc
+++ b/xls/dslx/fmt/ast_fmt.cc
@@ -127,6 +127,13 @@ std::optional<DocRef> Formatter::FormatCommentsBetween(
   return ConcatN(arena_, pieces);
 }
 
+bool Formatter::FirstCommentTrails(const Pos& from, const Pos& to) {
+  std::vector<const CommentData*> comments =
+      comments_.GetComments(Span(from, to));
+  return !comments.empty() &&
+         comments.front()->span.start().lineno() == from.lineno();
+}
+
 // If there is a '.' modifier in the attribute given by s (e.g. if it's of the
 // form "ProcName.config") strips off the trailing modifier and returns the
 // stem.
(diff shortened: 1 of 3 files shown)


xls/dslx/fmt/ast_fmt.cc
```diff
@@ -597,6 +604,91 @@ DocRef Formatter::FormatArray(const Array& n) {
...
+      return;
+    }
+    body_pieces.push_back(FormatJoin(
+        current_run, Joiner::kCommaBreak1AsGroupTrailingCommaAlways));
+    current_run.clear();
+  };
+
+  // Split into multiple runs of continuous elements between comments.
+  // Each run is formatted using FormatJoin to clump elements.
+  for (const Expr* item : items) {
```
dplassgit: Optional: `const auto*` is the pattern used frequently in dslx code.

xls/dslx/fmt/ast_fmt.cc
```diff
@@ -597,6 +604,91 @@ DocRef Formatter::FormatArray(const Array& n) {
...
+
+    current_run.push_back(Format(item));
+    last_array_element_span_limit = span.limit();
+  }
+  flush_run();
+
+  // Handle ellipsis
+  if (n.has_ellipsis()) {
+    body_pieces.push_back(
+      ConcatNGroup(arena_, {arena_.break1(), arena_.MakeText("...")}));
```
dplassgit, line 668: Should this update lasts_array_element_span_limit?

dplassgit (changes requested):
I'm concerned this PR will affect the formatting of all dslx code -- including internal code -- so I'm going to do some investigation before approving.

xls/dslx/fmt/ast_fmt.cc
achowd32: changed!

xls/dslx/fmt/ast_fmt.cc
achowd32: If there were a trailing comment (a comment coming after all elements in the array) that came _before_ the ellipsis, then updating `lasts_array_element_span_limit` would delete that comment.

Side note:

There doesn't seem to be a clean way to determine whether the trailing comment comes before or after the ellipsis, because the Array class does not store information about the location of the ellipsis - only whether it exists. That's why I made a note of point 3 when I made the PR. I think, unless we want to put some effort into touching the frontend, we can't always preserve the ordering of the ellipsis and trailing comment.

I've left a TODO above the if statement to reflect this.

dplassgit:
I've cloned this change and will test on internal DSLX code to make sure we're not breaking anything. I'm pretty sure it will be OK, just don't want to cause too much churn.

dplassgit (approved):

PR #4678, opened 2026-07-29
Fix dependencies for :resource_sharing_pass_test_base

hzeller:
A recent refactoring left a lot of now unused dependencies in place.

Fixed with `bant` using our build-cleaner script:

```
source <(xls/dev_tools/run-build-cleaner.sh ...)
```


hzeller:
Probably Mircea is the best to review this one.

mtrofin (approved):

hzeller:
You probably also need to set the 'reviewing internally' label for it to be copybara'ed

PR #4696, opened 2026-07-31
Fix omitting SystemVerilog type when copying input/output ports in `Block::Clone()`

magancarz:
This pull request fixes a missing copy of input/output port's system verilog type when cloning an IR block. I've added two regression tests ensuring SystemVerilog type variable value will be properly handled when running block conversion passes.


ericastor (approved):
Solid solution, thanks!

PR #4713, opened 2026-08-04
Fixed bug where formatter adds comma to empty tuple with comment

### LINKED ISSUE #4676: Formatter appends comma to empty tuple that contains a comment
**Describe the bug**
This is an edge case where the formatter messes up an empty tuple that contains a comment.

**To Reproduce**
Consider the following code:
```
fn tup() -> () {
   (
       // this is an empty tuple
   )
}
```
Running this through the formatter generates
```
fn tup() -> () {
   (
       , // this is an empty tuple
   )
}
```
Notice the addition of the stray comma, which the interpreter is not able to parse.
**Expected behavior**
The comma should not be added.

**Additional context**
I think this is a fairly simple bug, I would be interested in resolving this is maintainers agree it is an issue.

achowd32:
This is a very small PR that closes #4676. Empty tuples containing comments no longer have commas added inside them. This is solved by adding a conditional inside the terminal comment check, which checks if the tuple is empty or not.


jbtayloriii (approved):
Thanks for the change, LGTM

PR #4731, opened 2026-08-06
dslx: reject duplicate match alternatives

dank-openai:
## Summary

- Reject duplicate DSLX match patterns inside grouped `|` alternatives.
- Point to both the original match pattern and its unreachable duplicate.
- Preserve valid grouped alternatives and existing duplicate-arm checks.

## Problem

DSLX currently accepts this unreachable second `E::A` pattern:

```dslx
match value {
  E::A => u32:0,
  E::B | E::A => u32:1,
  E::C => u32:2,
}
```

The compiler compares entire match arms, so it misses a repeated pattern nested
inside another arm's alternatives. Track the individual alternatives as well,
and report both occurrences when one repeats.

Programs containing exact duplicate alternatives now intentionally fail
compilation. Valid grouped alternatives continue to compile.

The separate pre-existing enum-alias spelling case remains documented by a
disabled regression and is not fixed by this change.

## Tests

The equivalent merged producer change passed:

- `//xls/dslx/type_system_v2:typecheck_module_v2_control_flow_test`
- `//xls/dslx/exhaustiveness:exhaustiveness_match_test`
- `//xls/dslx/type_system:typecheck_module_test`

The added diagnostic fixture is owned by
`//xls/dslx/tests/errors:error_modules_test`.

Current upstream `main` cannot start an ordinary Bazel build because its
`rules_hdl` patch targets a missing file; the same failure is already present
in [upstream main CI](https://github.com/google/xls/actions/runs/31119392836).
No unrelated dependency or build-configuration changes are included here.

### DIFF
--- a/xls/dslx/type_system_v2/populate_table_visitor.cc
+++ b/xls/dslx/type_system_v2/populate_table_visitor.cc
@@ -54,6 +54,7 @@
 #include "xls/dslx/frontend/pos.h"
 #include "xls/dslx/import_data.h"
 #include "xls/dslx/import_routines.h"
+#include "xls/dslx/status_payload_utils.h"
 #include "xls/dslx/type_system/deduce_utils.h"
 #include "xls/dslx/type_system_v2/import_utils.h"
 #include "xls/dslx/type_system_v2/inference_table.h"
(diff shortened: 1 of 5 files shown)


xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -594,19 +595,28 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
                                       file_table_);
     }
 
+    auto duplicate_pattern_error = [this](const Span& original_span,
+                                          const Span& duplicate_span,
+                                          std::string_view pattern) {
+      absl::Status status = absl::InvalidArgumentError(absl::StrFormat(
```
dplassgit: Can TypeInferenceErrorStatus be used here?

dplassgit:
"Current upstream main cannot start an ordinary Bazel build because its
rules_hdl patch targets a missing file; the same failure is already present
in [upstream main CI](https://github.com/google/xls/actions/runs/31119392836)."

That seems wrong. I filed https://github.com/google/xls/issues/4749

dplassgit (changes requested):
Good solution to this issue

ericastor:
> "Current upstream main cannot start an ordinary Bazel build because its rules_hdl patch targets a missing file; the same failure is already present in [upstream main CI](https://github.com/google/xls/actions/runs/31119392836)."
> 
> That seems wrong. I filed #4749

This is fixed at tip of tree; maybe this PR just needs to rebase?

xls/dslx/type_system_v2/populate_table_visitor.cc
dank-openai: Updated to use `TypeInferenceErrorStatus` while retaining both original and duplicate source-span payloads. The whole-arm regression now verifies both locations as well.

dplassgit (approved):

PR #4732, opened 2026-08-06
dslx: reject duplicate enum match alternatives through type aliases

dank-openai:
## Summary

- Reject duplicate enum match alternatives even when type aliases give the same
  member different source spellings.
- Report both the original and duplicate source locations.
- Cover direct aliases, alias chains, grouped alternatives, imported aliases,
  and valid distinct members.

Depends on #4731. This branch adds one focused, four-file commit on top of that
open prerequisite; GitHub's `main`-based diff will also include #4731 until it
merges. The recently merged pattern-tree refactor in #4614 currently conflicts
with #4731 in `populate_table_visitor.cc`; that prerequisite must be reconciled
with current `main` before this dependent change can land.

## Problem

An enum type alias does not create new enum members, so the second arm below can
never match:

```dslx
type Alias = E;

match value {
  E::A => u32:0,
  Alias::A => u32:1,
  E::B => u32:2,
}
```

The existing duplicate check compares source spellings, so `E::A` and
`Alias::A` incorrectly appear distinct. The compiler now rejects the program
with the existing two-location diagnostic:

```text
fake.x:10:5-10:13
0009:     E::A => u32:0,
0010:     Alias::A => u32:1,
~~~~~~~~~~^------^ TypeInferenceError: Exact-duplicate pattern match detected `Alias::A`; only the first could possibly match; previously @ fake.x:9:5-9:9
fake.x:9:5-9:9
0009:     E::A => u32:0,
~~~~~~~~~~^--^
```

## Implementation

Resolve enum alternatives through the existing type-alias and import resolver,
then compare their underlying enum-member declarations. Keep the existing
whole-arm and source-spelling checks unchanged.

Distinct declared members remain distinct even when they share an underlying
numeric value. Existing grouped alternatives, unknown-member diagnostics, and
other match-pattern semantics are preserved.

## Tests

- The direct-alias regression checks both exact source spans.
- Chained aliases, grouped alternatives, imported aliases, and valid distinct
  members each have focused regression coverage.
- The complete upstream
  `//xls/dslx/type_system_v2:typecheck_module_v2_control_flow_test` target
  passes, including all five focused alias cases, with validation-only
  dependency and macOS deployment-target overrides.
- The equivalent merged producer change passed the complete DSLX control-flow,
  exhaustiveness, and existing typechecker suites.
- Current upstream `main` cannot start an ordinary Bazel build because its
  `rules_hdl` patch targets a missing file; the same failure is already present
  in [upstream main CI](https://github.com/google/xls/actions/runs/31119392836).
  No unrelated dependency or build-configuration changes are included.

### DIFF
--- a/xls/dslx/tests/errors/duplicate_enum_match_arm.x
+++ b/xls/dslx/tests/errors/duplicate_enum_match_arm.x
@@ -1,0 +1,23 @@
+// Copyright 2026 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+enum Example: u2 { A = 0, B = 1, C = 2 }
+
+fn main(value: Example) -> u32 {
+  match value {
+    Example::A => u32:0,
+    Example::B | Example::A => u32:1,
+    Example::C => u32:2,
+  }
+}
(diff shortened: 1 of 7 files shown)


dplassgit:
Blocked on #4731

proppy:
@dplassgit unblocked? now that https://github.com/google/xls/pull/4731 is merged?

dplassgit (changes requested):
Is this rebased against TOT?

PR #4747, opened 2026-08-11
Preserve source locations in optimization passes

rw1nkler:
This PR adds a `MergeLocs `helper that merges and deduplicates source locations across a node's operands.
It's used wherever passes synthesize replacement nodes, since those call sites often dropped the
original location, making it harder to trace optimized IR and generated Verilog back to the source DSLX.
`VastNode::PreEmit` is also updated to emit one comment per merged location, with line-count tracking
adjusted accordingly. Tests cover `MergeLocs` and each annotation strategy end to end.

### DIFF
--- a/xls/ir/node_util.cc
+++ b/xls/ir/node_util.cc
@@ -87,6 +87,18 @@ std::vector<Node*> RemoveRedundantNodes(
 
 }  // namespace
 
+SourceInfo MergeLocs(absl::Span<Node* const> operands) {
+  SourceInfo loc;
+  for (Node* operand : operands) {
+    for (const SourceLocation& location : operand->loc().locations) {
+      if (!absl::c_linear_search(loc.locations, location)) {
+        loc.locations.push_back(location);
+      }
+    }
+  }
+  return loc;
+}
+
 bool IsLiteralWithRunOfSetBits(Node* node, int64_t* leading_zero_count,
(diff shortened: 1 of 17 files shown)


xls/ir/node_util.cc
```diff
@@ -87,6 +87,22 @@ std::vector<Node*> RemoveRedundantNodes(
 
 }  // namespace
 
+SourceInfo MergeLocs(absl::Span<Node* const> operands) {
+  SourceInfo loc;
+  for (Node* operand : operands) {
+    for (const SourceLocation& location : operand->loc().locations) {
+      bool already_present =
+          absl::c_any_of(loc.locations, [&](const SourceLocation& existing) {
+            return existing <=> location == std::strong_ordering::equal;
```
ericastor, line 96: ... is there a reason this isn't just `absl::c_linear_search(loc.locations, existing)`?

ericastor (commented):
This is a great step forward, thank you!

proppy:
can we rebase?

xls/ir/node_util.cc
rw1nkler: I added `==` operator and removed this part

ericastor:
This seems to have a failing test: `StrategyCommentEmitsLocationComments`?

ericastor (approved):

PR #4815, opened 2026-08-20
Added more modules to xls/modules/ml

achowd32:
Just building off of some of my previous work adding ML modules to the `xls/modules/` directory. I've added some more approximated activation functions, a mac adder tree, and an output stationary systolic array.


dplassgit:
Please fix failing tests

dplassgit (changes requested):
Please fix failing tests.

PR #4824, opened 2026-08-21
[jit] Add standalone C++ headers for AOT function entrypoints

### LINKED ISSUE #4823: [enhancement] Generate C++ headers for AOT entrypoints

Using an AOT-compiled XLS function from a standalone C++ program currently requires reconstructing its packed ABI from AotPackageEntrypointsProto. The consumer needs to know the packed input/output sizes, temporary-buffer requirements, XLS type structure, and the generated packed-function symbol.

This makes the object file somewhat awkward to consume outside of code that already depends on XLS/protobuf, and makes it easy for a C++ wrapper to get out of sync with the AOT metadata.

Generate the AOT metadata proto/textproto and either parse it in the consuming program or use it to maintain a separate C++ wrapper.

For lightweight consumers this means either introducing dependencies on XLS/protobuf solely for ABI metadata, or manually duplicating generated details such as buffer sizes, type layouts, temporary-buffer requirements, and the packed entrypoint symbol.

aot_compiler_main could optionally emit a self-contained C++ header describing the packed ABI of a function entrypoint.

The header could depend only on the C++ standard library and expose the input/result type structure, packed buffer sizes, temporary-buffer size/alignment, and a stable C++ identifier bound to the generated packed entrypoint.

For example, an --output_cpp_header option could generate this alongside the AOT object and existing proto/textproto metadata.

tilir:
This change adds optional generation of a self-contained C++20 header for AOT-compiled function entrypoints.

The generated header describes the packed ABI using standard C++ types and constants derived from the AOT entrypoint metadata. It also provides a C++ declaration for the generated packed function, allowing the AOT object to be consumed without parsing AotPackageEntrypointsProto at runtime or manually reproducing ABI details in the caller.

The header generation is exposed through a new aot_compiler_main option and is intended to complement the existing object and proto/textproto outputs.

Fixes #4823


tilir:
Will try to pass this through OSS: https://github.com/google/xls/pull/4824

allight:
See comment in #4823. This is not a feature we need. For functions the wrapper already provides access to the bit-packed entry points in a way that abstracts the internal details of how exactly our compiler works with essentially no overhead.

Furthermore, we absolutely should not be doing text templating like this in C++. We generally use python an Jinja to handle c++ code generation for things like this.

tilir:
Closing this for now while I take a closer look at the existing JIT wrapper mechanism discussed in #4823.

Thanks for pointing me in the right direction. If I find that my use case still requires something not covered by the current mechanism, I'll come back with a more focused proposal.

PR #4831, opened 2026-08-21
Fix IR conversion of static impl functions on parametric structs

### LINKED ISSUE #4692: [DSLX] IR conversion fails when calling static impl method on parametric struct inside a proc
**Describe the bug**
In DSLX, calling a static method defined inside an `impl` block of a parametric struct from inside a parametric proc causes IR conversion (ir_converter_main / GetConversionRecords) to fail with:

Error: INTERNAL: IrConversionError: ... Failed to convert IR because type was missing for AST node: bits[N]

**To Reproduce**
```
struct Foo<N: u32> {
    val: bits[N],
}

impl Foo<N> {
    fn new(x: bits[N]) -> Self {
        Foo<N> { val: x }
    }
}

proc Bar<N: u32> {
    init {
        Foo<N>::new(bits[N]:0)
    }
    config() {
        ()
    }
    next(state: Foo<N>) {
        state
    }
}

proc main {
    init { () }
    config() {
        spawn Bar<u32:32>();
    }
    next(state: ()) {
        ()
    }
}
```

Command:
bazel run //xls/dslx/ir_convert:ir_converter_main -- /path/to/repro.x

Output:
Error: INTERNAL: IrConversionError: /path/to/repro.x:6:15-6:22 Failed to convert IR because type was missing for AST node: bits[N] (kind: type annotation)
=== Source Location Trace: ===
xls/dslx/ir_convert/function_converter.cc:113
xls/dslx/ir_convert/function_converter.cc:5482
xls/dslx/ir_convert/function_converter.cc:868
xls/dslx/ir_convert/function_converter.cc:3529
xls/dslx/ir_convert/ir_converter.cc:365
xls/dslx/ir_convert/ir_converter.cc:413
xls/dslx/ir_convert/ir_converter.cc:663
xls/dslx/ir_convert/ir_converter.cc:734
xls/dslx/ir_convert/ir_converter_main.cc:159

**Expected behavior**
ir_converter_main recognizes static functions in `impl Foo<N>` as scoped to a parametric struct, skips un-instantiated root conversion, and monomorphizes `Foo<N>::new` with the parametric environment from `Bar<u32:32> (N=32)`.

**Additional context**
Possibly related: #2082 and #2043

JayPatel095:
Fixes #4692

`Function::IsMethodOnParametricStruct` requires `IsMethod()`, which only holds when the first parameter is `self`. A static function in an impl of a parametric struct has no `self`, so it was treated as fully existing despite its parameters referring to the struct's parametrics.

The failure points of this issue:

- `get_conversion_records.cc:180`: the guard doesn't skip and gets a conversion record with an empty `ParametricEnv`, resolving its `bits[N]` parameter then fails with `type was missing for AST node: bits[N]` (original issue)
- `get_conversion_records.cc:216`: takes the concrete branch of the env invariant, so the env supplied by the invocation fails on `RET_CHECK(env.empty())`
- `function_converter.cc:4746`: omits the instantiation suffix when mangling the callee, so the invocation looks up a non-existent name, failing with `Could not find name for invocation`
- `function_converter.cc:3634`: omits the same suffix when emitting the implicit-token wrapper, so both instantiations produce one symbol, failing with `not unique within package`

Fix is to replace `Function::IsMethodOnParametricStruct` with `Function::IsFunctionOnParametricStruct`, which doesn't have `IsMethod()` and is already the predicate `type_system_v2` uses for this distinction. 

Repro'd original bug (+ others), verified the more idiomatic Foo<N>::new(0) fails identically to the issue's Foo<N>::new(bits[N]:0). Verified that `xls/dslx/tests/parametric_impl.x`, `xls/dslx/stdlib/apfloat.x`, and the `ParametricProcDef` program produce byte-identical IR before and after.

### DIFF
--- a/xls/dslx/ir_convert/testdata/ir_converter_test_ConvertStaticMethodOnParametricStruct.ir
+++ b/xls/dslx/ir_convert/testdata/ir_converter_test_ConvertStaticMethodOnParametricStruct.ir
@@ -1,0 +1,23 @@
+package test_module
+
+file_number 0 "test_module.x"
+
+fn __test_module__S_A___make__16(a: bits[16] id=1) -> (bits[16]) {
+  A: bits[32] = literal(value=16, id=2)
+  ret tuple.3: (bits[16]) = tuple(a, id=3)
+}
+
+fn __test_module__S_A___make__32(a: bits[32] id=4) -> (bits[32]) {
+  A: bits[32] = literal(value=32, id=5)
+  ret tuple.6: (bits[32]) = tuple(a, id=6)
+}
+
+fn __test_module__main() -> (bits[16], bits[32]) {
+  literal.7: bits[16] = literal(value=100, id=7)
+  literal.10: bits[32] = literal(value=200, id=10)
+  invoke.8: (bits[16]) = invoke(literal.7, to_apply=__test_module__S_A___make__16, id=8)
+  invoke.11: (bits[32]) = invoke(literal.10, to_apply=__test_module__S_A___make__32, id=11)
+  X: bits[16] = tuple_index(invoke.8, index=0, id=9)
+  Y: bits[32] = tuple_index(invoke.11, index=0, id=12)
+  ret tuple.13: (bits[16], bits[32]) = tuple(X, Y, id=13)
+}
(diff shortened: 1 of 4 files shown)


erinzmoore (approved):

xls/dslx/ir_convert/testdata/ir_converter_test_ConvertStaticMethodOnParametricStruct.ir
erinzmoore, line 1: It looks like this no longer passes at HEAD. Can you rebase and fix up the test?

erinzmoore (changes requested):

JayPatel095:
@erinzmoore Rebased, regenerated the golden. Also confirmed that the fix is still needed.

allight (approved):

PR #4845, opened 2026-08-25
 Support use syntax in TIv2

rw1nkler:
This PR adds support for `use` support in TIv2, and makes it operational with bytecode interpreter that relies on TIv1 data structures.

What's included:
- Resolution of use-bound constants, functions, structs, and type aliases, used directly or reached via `ColonRef` on a whole-module use subject, including nested module paths
- Spawning use-imported procs and proc aliases
- Constexpr propagation for use-imported constants, so they work as parametric defaults
- A fix for an IR-conversion crash on whole-module use of a module named std

Implementation details:
- Shared `ImportData` helpers (`ResolveUseImportedTarget`, `GetUseImportedModuleInfo`) let the typechecker and bytecode emitter resolve a use-bound name consistently
- `ColonRef` resolution now distinguishes a whole-module use subject from a use-bound value
- The parser binds a use-bound proc's bare name plus  `.config`/`.next`/`.init` keys to the same entry. Since all four resolve to the same proc, the identifier's suffix is the only signal for which member was meant. This suffix lookup runs independently at typecheck time and at bytecode-emission time.

This PR also re-enables the use-syntax end-to-end DSLX tests that were previously disabled pending TIv2 support.


xls/dslx/frontend/ast_utils.cc
```diff
@@ -792,4 +794,21 @@ std::optional<const ParametricBinding*> GetProcParametricBinding(
   return binding;
 }
 
+std::optional<Function*> GetProcMemberBySuffix(Proc* proc,
+                                               std::string_view identifier) {
+  if (absl::EndsWith(identifier,
+                     absl::StrCat(kProcMemberSeparator, kProcConfigName))) {
```
richmckeever, line 800: I think we should just not support legacy procs at this point; we are now close to being able to eliminate them and just have impl-based procs. I think if you revert any logic making it work for legacy procs, the surface area will be considerably smaller.

richmckeever (changes requested):

Issue #1434, opened 2024-05-27
Tracking issue: OS X build/test status


cdleary:
Just an issue to track and take notes on the current (merged to google/xls) status of OS X support. Taking notes in an issue in the google/xls repo as it's (currently) canonical mainline. This is all work-aroundable, but my intent is to get mainline into a state where it "usually" works on OS X (via fixes and PRs if it happens to break).

- The hermetic LLVM toolchain does not work on OS X
- But if you comment out the registration of LLVM toolchain from the `WORKSPACE` file, almost everything works with the clang-15 installation that's present on latest OS X
- iverilog build doesn't work, which I think is understandable because we did a lot of hacktastic stuff to make it work on Linux
- current grpc dep (at HEAD) has issues
- AoT compilations fail here: https://github.com/google/xls/blob/ec4df616bc8b3acd25db670259316e675b53edc9/xls/jit/aot_compiler.cc#L99

```
$ clang --version
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Note that if you don't comment out the hermetic toolchain registration it gives you an error indicating "you invoked llvm-ar with invalid command line arguments". Aside: I often use `//xls/ir:bits_test` as a simple canary target for a binary to build/test.

```
$ bazel build -c opt //xls/ir:bits_test
INFO: Analyzed target //xls/ir:bits_test (1 packages loaded, 2104 targets configured).
INFO: Found 1 target...
ERROR: /private/var/tmp/_bazel_cdleary/60c3e5ce2889de0dfd8db639786fbce7/external/com_google_absl/absl/base/BUILD.bazel:60:11: Linking external/com_google_absl/absl/base/liblog_severity.a [for tool] failed: (Exit 1): llvm-ar failed: error executing command (from target @com_google_absl//absl/base:log_severity) external/llvm_toolchain/bin/llvm-ar @bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/com_google_absl/absl/base/liblog_severity.a-2.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/llvm_toolchain/bin/llvm-ar: error: an archive name must be specified
OVERVIEW: LLVM Archiver

[...]

PR #4862, opened 2026-08-27
Bump hermetic LLVM toolchain to 0.8.18 for a fetchable macOS SDK

zarubaf:
The macOS SDK archive pinned by `llvm` 0.8.11 is no longer served — Apple answers that URL with HTTP 403. Bazel falls back to the `web.archive.org` mirror listed next to it, which serves at roughly 3 KB/s, so the fetch does not fail fast but times out after several minutes:

```
Error in download: java.io.IOException: Error downloading
[https://swcdn.apple.com/.../CLTools_macOSNMOS_SDK.pkg,
 https://web.archive.org/.../CLTools_macOSNMOS_SDK.pkg]: java.net.SocketTimeoutException: Read timed out

ERROR: no such package '@@llvm++osx+macos_sdk//sysroot'
ERROR: Analysis of target '//xls/synthesis:_synthesis_service_cc_grpc_grpc_codegen' failed
```

Nothing builds on macOS past the analysis phase while this is the case.

0.8.14 through 0.8.17 still reference the same retired URL. 0.8.18 is the first release pointing at an SDK archive Apple currently serves (`MacOSX26.5.sdk`); I measured that one at ~7 MB/s.

### On the lockfile

The lockfile change here is limited to the BCR registry hashes implied by the bump (9 insertions, 6 deletions).

Worth flagging separately: running `bazel mod deps --lockfile_mode=update` on `main` today *also* rewrites the `llvm_raw_ext` entry, because the lock records llvm-project `fdb39b1112a6db0343de169b79820768bd16c7d9` while `dependency_support/llvm/extension.bzl` now specifies `ab547095ead5464dc024d66264d9b8a987f429f3`. That looks like it came in with `1e745d84d` ("Integrate LLVM at llvm/llvm-project@ab547095e") and is what CI is currently failing on:

```
MODULE.bazel.lock is no longer up-to-date because: The implementation of the extension
'@@//dependency_support/llvm:extension.bzl%llvm_raw_ext' or one of its transitive .bzl files has changed.
```

Refreshing that entry is not sufficient on its own — `dependency_support/llvm/platforms_config.patch` does not apply against the newer LLVM revision (`CONTENT_DOES_NOT_MATCH_TARGET` in `config.bzl`), so the patch needs updating too. That is unrelated to this change, so I have deliberately left the `llvm_raw_ext` entry alone rather than folding it in here.

### Verification

~Verified on macOS arm64: the SDK fetch fails before and succeeds after.~

~Because `main` is currently red for the unrelated reason above, the end-to-end check was done on `cf36eb5a3` (the last green ancestor). With this change plus the accompanying macOS fixes, the four `nightly-macos.yml` targets build successfully (5351 actions).~

Re-based.

Context: #1434


erinzmoore:
Thanks for finding this issue and fixing. Do you mind rebasing now that CI is green to show everything passing on this PR?

zarubaf:
Rebased, and I think CI is green now. Thanks for the review.

erinzmoore (approved):

PR #4864, opened 2026-08-27
Make the embedded subprocess_helper work on non-Linux hosts

zarubaf:
`xls/common/subprocess.cc` does not compile on macOS:

```
xls/common/subprocess.cc:18:10: fatal error: 'linux/memfd.h' file not found
```

`2c355ce53` ("Embed subprocess_helper in subprocess.cc") replaced the runfiles lookup with an embedded copy executed from a memfd. The mechanism is Linux specific — neither `memfd_create()` nor `/proc/self/fd` exists on macOS — and the include is unconditional.

The blast radius is larger than one file. `//xls/common:subprocess` is reachable from all four binaries that `nightly-macos.yml` builds (`interpreter_main`, `ir_converter_main`, `opt_main`, `codegen_main`), so the macOS nightly build cannot succeed while this is present.

### Change

Helper creation moves behind a single `CreateSubprocessHelper()` returning a path to spawn:

- **Linux** keeps the existing memfd behavior. The only difference is that the fd is now released rather than held in a `StatusOr` — it has to stay open for the process lifetime for `/proc/self/fd/N` to resolve, which the previous function-local static did implicitly.
- **Elsewhere** the embedded helper is written to a private `mkstemp()` file under `$TMPDIR`, marked executable, closed, and unlinked via `atexit`. Closing before exec matters: macOS returns `ETXTBSY` for a file still open for writing.

### The tradeoff, and a question

The temporary file gives up the "no dependency on build artifacts continuing to exist" property that motivated the memfd approach. I could not find a way to preserve it on macOS: there is no `fexecve()` and no `/proc`, so exec needs a real path.

If you would rather not have a temp file on macOS at all, the alternative is falling back to the previous `GetXlsRunfilePath()` lookup on non-Linux and keeping memfd on Linux. That keeps nothing on disk that was not already on disk, at the cost of reintroducing the runfiles dependency for that platform. Happy to switch if that is preferred — I picked the temp file because it keeps behavior uniform across platforms.

### Verification

macOS arm64, `-c opt`, on top of the minimum-OS and hermetic LLVM SDK changes needed to build there at all:

- `//xls/common:subprocess` fails to compile before, builds after
- the four `nightly-macos.yml` targets fail before, build after
- `//xls/common:subprocess_test` passes — though only with one additional local workaround, noted below

Since `main` is currently red on an unrelated lockfile issue, this was checked on `cf36eb5a3`, the last green ancestor.

The `subprocess_test` caveat: linking any Mach-O shared library currently fails because rules_cc appends a GNU-style `-Wl,-soname=` that ld64 rejects (`ld64.lld: error: unknown argument '-soname=...'`). I worked around it locally with `--nointerface_shared_objects` to run the test. That is a separate problem, not addressed here.

Context: #1434

### DIFF
--- a/xls/common/subprocess.cc
+++ b/xls/common/subprocess.cc
@@ -15,20 +15,27 @@
 #include "xls/common/subprocess.h"
 
 #include <fcntl.h>
-#include <linux/memfd.h>
 #include <signal.h>  // NOLINT
 #include <spawn.h>
 #include <stdlib.h>  // NOLINT for WIFEXITED, WEXITSTATUS; not in <cstdlib>
-#include <sys/mman.h>
 #include <sys/poll.h>
+#include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
+#if defined(__linux__)
+#include <linux/memfd.h>
+#include <sys/mman.h>
+#endif
+
 #include <atomic>
 #include <cerrno>
+#include <cstdint>
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
+#include <memory>
 #include <ctime>
 #include <filesystem>
 #include <optional>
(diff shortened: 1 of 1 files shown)


xls/common/subprocess.cc
```diff
@@ -149,30 +155,90 @@ class CleanableFd {
...
+      unlink(helper_path.c_str());
+      return absl::InternalError(absl::StrCat(
+          "Failed to make subprocess helper executable: ", Strerror(errno)));
+    }
+    // Closing here matters: macOS refuses to exec a file that is still open for
+    // writing (ETXTBSY).
+  }
+
+  static std::string* const cleanup_path = new std::string(helper_path);
+  std::atexit([] { unlink(cleanup_path->c_str()); });
```
allight: I'd much prefer to have static destructors handle this instead of an atexit handler.

Just add a `FileName` function to the CleanableFD, make it virtual and have the mac version return an override with the unlink in its destructor (and the linux one does the close it currently has). This means the normal static destructors deal with closing the FD on linux and removing the file in mac.

xls/common/subprocess.cc
```diff
@@ -149,30 +155,90 @@ class CleanableFd {
...
-        "Failed to write subprocess helper to memfd: ", Strerror(errno)));
+        "Failed to create memfd for subprocess helper: ", Strerror(errno)));
   }
+  XLS_RETURN_IF_ERROR(WriteSubprocessHelperTo(fd));
   if (lseek(fd, 0, SEEK_SET) != 0) {
     return absl::InternalError(absl::StrCat(
         "Failed to seek subprocess helper in memfd: ", Strerror(errno)));
   }
-  return std::move(fd);
+  // Every spawn names this fd through /proc/self/fd, so it has to stay open for
```
allight: We need to make sure that the FD is closed by the static destructors or some other mechanism. Just leaking it is not ok. Especially because we care about the linux build in a way that we (unfortunately) don't for the mac build.

xls/common/subprocess.cc
```diff
@@ -149,30 +155,90 @@ class CleanableFd {
...
-            get_subprocess_helper_embedded().size()) !=
-      get_subprocess_helper_embedded().size()) {
+  return absl::OkStatus();
+}
+
+#if defined(__linux__)
+
+// Runs the helper straight out of an anonymous in-memory file, so that nothing
+// on disk needs to outlive the build that produced it.
+absl::StatusOr<std::string> CreateSubprocessHelper() {
```
allight: Instead maybe make CleanableFD a virtual class with a 

virtual std::string file_name() const = 0;

and have this (in both versions) return absl::StatusOr<std::unique_ptr<CleanableFD>>

allight (changes requested):

xls/common/subprocess.cc
zarubaf: Done.

xls/common/subprocess.cc
zarubaf: Done!

xls/common/subprocess.cc
zarubaf: You are right. The memfd is owned by `MemfdSubprocessHelper` and closed by the inherited `~CleanableFd()` when the function-local static is destroyed.

zarubaf:
Thanks for taking a look. Verified locally on my Mac.

xls/common/subprocess.cc
```diff
@@ -129,50 +136,122 @@ absl::StatusOr<posix_spawn_file_actions_t> CreateChildFileActions(
...
+  // Path to hand to posix_spawn to execute the helper.
+  virtual std::string file_name() const = 0;
+
   operator int() const { return fd_; }
 
+ protected:
+  int fd() const { return fd_; }
+
  private:
   int fd_ = -1;
```
allight, line 162: Just have the FD be only in the linux override.

xls/common/subprocess.cc
```diff
@@ -129,50 +136,122 @@ absl::StatusOr<posix_spawn_file_actions_t> CreateChildFileActions(
...
+                     Strerror(errno)));
+  }
+  auto helper = std::make_unique<TempFileSubprocessHelper>(path.data());
+
+  absl::Status written = WriteSubprocessHelperTo(fd);
+  if (!written.ok()) {
+    close(fd);
+    return written;
+  }
+  if (fchmod(fd, 0700) != 0) {
```
allight, line 244: 0500

No need to have write permission anymore.

xls/common/subprocess.cc
```diff
@@ -129,50 +136,122 @@ absl::StatusOr<posix_spawn_file_actions_t> CreateChildFileActions(
...
+  XLS_RETURN_IF_ERROR(WriteSubprocessHelperTo(*helper));
+  if (lseek(*helper, 0, SEEK_SET) != 0) {
     return absl::InternalError(absl::StrCat(
         "Failed to seek subprocess helper in memfd: ", Strerror(errno)));
   }
-  return std::move(fd);
+  return helper;
 }
 
+#else
```
allight, line 204: #else  // for MACOS and other non-linux

allight (changes requested):
Just a few minor changes. Otherwise LGTM

PR #4881, opened 2026-08-31
[opt] Add masked-xor strength reduction for binary selects

mag-mga:
A binary select of the form `sel(p, [x, xor(x, c)])` is rewritten into `xor(x, sign_ext(p) & c)` when c is fully known, with the selector complemented when the xor arm is the on-false case. The rewrite also applies when an earlier pass (e.g. concat_simp) already split the xor limb into per-slice pieces, as long as the two arms provably differ by a fully known constant (TryFoldXorDelta).

Runs at optimization levels where split-based rewrites are enabled (SplitsEnabled); it is the strength-reduced form of the per-step select in an unrolled CRC/shift loop.


mag-mga:
Hi, @ericastor, please, take a look at this PR. My optimization seems to be close to your `MaybeConvertSelectToMask` inside `select_simp`.

xls/passes/select_simplification_pass_test.cc
```diff
@@ -1187,6 +1187,189 @@ TEST_P(SelectSimplificationPassTest, PrioritySelectWithOnlyNonzeroCaseDefault) {
                      m::SignExt(m::Eq(m::Param("p"), m::Literal(0b00)))));
 }
 
+TEST_P(SelectSimplificationPassTest, SelectWithConstantXorCaseCrcShift) {
```
allight: Please use FunctionBuilder to create tests.

xls/passes/select_simplification_pass.cc
```diff
@@ -1095,6 +1095,293 @@ absl::StatusOr<bool> MaybeConvertSelectToMask(Node* node,
...
+// Depth cap for the xor-delta fold below: together with the `visited` set,
+// recursion stays bounded so that deeply nested (or adversarially shaped) IR
+// cannot exhaust the stack. A fold that would exceed the cap is a conservative
+// miss, never an incorrect answer.
+constexpr int64_t kMaxXorFoldDepth = 1000;
+
+// Flattens `node` into its non-concat pieces, most-significant first: the same
+// order in which `concat` lists its operands and `bits_ops::Concat`
+// reassembles them.
+void FlattenConcats(Node* node, std::vector<Node*>& pieces) {
```
allight: Because both concat_simp and select_simp are in the fixedpoint_simp compound pass this seems like it could cause the optimization to never finish if it triggers (because concat_simp will continuously undo the change).

xls/passes/select_simplification_pass.h
```diff
@@ -172,6 +172,17 @@ namespace xls {
 //     potentially more efficient components, and exposing further optimization
 //     opportunities.
 //
+// 16. **Converting Binary `Select` to Masked XOR**:
```
allight: This seems very unlike the rest of the transforms in this pass in that it really depends on the values in the select cases.

This transform should either be in strength_red or more likely in a totally separate pass itself IMO.

(There's also some questions we've been having internally about changing the way that we organize optimization phases which would probably want this to be late in compilation, but that's only peripherally related to this.)

mag-mga:
@ericastor , ping

allight:
@mag-mga please address comments before asking for further review.

Edit: sorry bad github ui. Comments sent

allight (commented):
Oh whoops sorry hadn't actually sent these

ericastor:
Apologies - I was at a conference all last week, and then today was a US holiday. I'll pick this up for a proper review in the morning.

xls/passes/select_simplification_pass.cc
ericastor: IIUC, this doesn't actually flatten the concats - it reads *through* them so the operation can see what it's working on. It still only applies if this optimization actually takes an action, which I don't think is something `concat_simp` can undo.

xls/passes/select_simplification_pass.h
ericastor: It really is closely related to transform (14) above. I would agree that maybe both of these should be split out into another pass, though.

ericastor:
The code seems solid, and this is an interesting proposal - though I think I agree with @allight that it's starting to look like it'd be worth splitting out the "select -> bitwise operations" transforms into their own pass. Are you able to share a concrete example of something where this makes a difference to optimized IR quality or to PPA?

xls/passes/select_simplification_pass_test.cc
mag-mga: Rewrote to FunctionBuilder

xls/passes/select_simplification_pass.cc
mag-mga: Make it as independent select_to_bitwise_ops pass in the second commit

mag-mga:
> The code seems solid, and this is an interesting proposal - though I think I agree with @allight that it's starting to look like it'd be worth splitting out the "select -> bitwise operations" transforms into their own pass. Are you able to share a concrete example of something where this makes a difference to optimized IR quality or to PPA?

We had a hot CRC function inside many loops (something like that below)
```
top fn __crc16_attach_single_proc__crc16_step(crc: bits[16] id=1, bit: bits[1] id=2) -> bits[16] {
  bit_slice.14: bits[15] = bit_slice(crc, start=0, width=15, id=14, pos=[(0,9,16)])
  literal.13: bits[1] = literal(value=0, id=13, pos=[(0,9,16)])
  bit_slice.4: bits[1] = bit_slice(crc, start=15, width=1, id=4, pos=[(0,8,20)])
  shifted: bits[16] = concat(bit_slice.14, literal.13, id=15, pos=[(0,9,16)])
  CRC16_POLY: bits[16] = literal(value=4129, id=3, pos=[(0,5,19)])
  feedback: bits[1] = xor(bit_slice.4, bit, id=5, pos=[(0,8,20)])
  xor.8: bits[16] = xor(shifted, CRC16_POLY, id=8, pos=[(0,12,4)])
  ret sel.9: bits[16] = sel(feedback, cases=[shifted, xor.8], id=9, pos=[(0,11,2)])
}
```
Previously, it generates
```
module __crc16_attach_single_proc__crc16_step(
  input wire [15:0] crc,
  input wire bit,
  output wire [15:0] out
);
  wire feedback;
  wire [15:0] shifted;
  assign feedback = crc[15] ^ bit;
  assign shifted = {crc[14:0], 1'h0};
  assign out = feedback ? {crc[14:0] ^ 15'h0810, 1'h1} : shifted;
endmodule
```
Now it leads to 
```
module __crc16_attach_single_proc__crc16_step(
  input wire [15:0] crc,
  input wire bit,
  output wire [15:0] out
);
  wire feedback;
  wire [15:0] shifted;
  assign feedback = crc[15] ^ bit;
  assign shifted = {crc[14:0], 1'h0};
  assign out = shifted ^ {16{feedback}} & 16'h1021;
endmodule
```
For our target it is especially better for FPGA LUT and sometimes also better in synthesis (although some synthesis tools can finally synthesize both functions in the same way).

I'm not confident I have enough permission to provide PPA :(

ericastor:
> > The code seems solid, and this is an interesting proposal - though I think I agree with @allight that it's starting to look like it'd be worth splitting out the "select -> bitwise operations" transforms into their own pass. Are you able to share a concrete example of something where this makes a difference to optimized IR quality or to PPA?
> 
> We had a hot CRC function inside many loops (something like that below)
> 
> ```
> top fn __crc16_attach_single_proc__crc16_step(crc: bits[16] id=1, bit: bits[1] id=2) -> bits[16] {
>   bit_slice.14: bits[15] = bit_slice(crc, start=0, width=15, id=14, pos=[(0,9,16)])
>   literal.13: bits[1] = literal(value=0, id=13, pos=[(0,9,16)])
>   bit_slice.4: bits[1] = bit_slice(crc, start=15, width=1, id=4, pos=[(0,8,20)])
>   shifted: bits[16] = concat(bit_slice.14, literal.13, id=15, pos=[(0,9,16)])
>   CRC16_POLY: bits[16] = literal(value=4129, id=3, pos=[(0,5,19)])
>   feedback: bits[1] = xor(bit_slice.4, bit, id=5, pos=[(0,8,20)])
>   xor.8: bits[16] = xor(shifted, CRC16_POLY, id=8, pos=[(0,12,4)])
>   ret sel.9: bits[16] = sel(feedback, cases=[shifted, xor.8], id=9, pos=[(0,11,2)])
> }
> ```
> ...
>
> For our target it is especially better for FPGA LUT and sometimes also better in synthesis (although some synthesis tools can finally synthesize both functions in the same way).
> 
> I'm not confident I have enough permission to provide PPA :(

Oh, no worries there! I meant the motivating example, rather than PPA specifics, so I appreciate your sharing it. I'll review this again, and thanks for reworking it as requested.

xls/passes/optimization_pass_pipeline.txtpb
```diff
@@ -424,6 +422,16 @@ compound_passes: [
       # TODO(allight): Unclear if fixedpoint simp is what we want to use here.
       # This is mostly to clean up extraneous bit-slice/concats etc.
       "fixedpoint_simp",
+      "dce",
+      # Convert binary selects whose arms differ by a fully-known xor term into
```
ericastor, line 426: As I mentioned, I'm not actually sure this is true - so we'll check it, and decide where it should live overall. Thanks!

xls/passes/select_to_bitwise_ops_pass.cc
```diff
@@ -0,0 +1,363 @@
...
+  }
+  // The two arms share no foldable structure and at least one is not fully
+  // known, so the delta cannot be proven constant.
+  return std::nullopt;
+}
+
+// Returns the (on_false, on_true) arm pair of a single-bit binary mux: either
+// two explicit cases, or one explicit case plus a default value. Returns
+// std::nullopt for any other select shape.
+std::optional<std::pair<Node*, Node*>> GetBinaryMuxArms(Select* sel) {
```
ericastor, line 187: As a followup, we might want to switch this to use our GenericSelect framework! But for now, this is sensible.

ericastor (approved):
This is great, thank you!

ericastor:
Oh darn, our CI checks are aborting on the lack of a newline at the end of your new files. Are you able to fix that?

I might just do it on import, though.

mag-mga:
> Oh darn, our CI checks are aborting on the lack of a newline at the end of your new files. Are you able to fix that?
> 
> I might just do it on import, though.

@ericastor ,
I've added newlines as a separate commit

mag-mga:
@ericastor , @allight , please, take a look again, I've addressed comments and fixed linter

mag-mga:
I think we should close this MR because it is already merged to main branch

PR #4899, opened 2026-09-03
Add a duck-typing example.

hzeller:
Even without traits, operations that look the same can be applied to different types.

### DIFF
--- a/xls/examples/BUILD
+++ b/xls/examples/BUILD
@@ -220,6 +220,12 @@ xls_dslx_test(
     dslx_test_args = {"compare": "jit"},
 )
 
+xls_dslx_test(
+    name = "duck_typing_test",
+    srcs = ["duck_typing.x"],
+    dslx_test_args = {"compare": "jit"},
+)
+
 xls_dslx_library(
     name = "fir_filter_dslx",
     srcs = ["fir_filter.x"],
--- a/xls/examples/duck_typing.x
+++ b/xls/examples/duck_typing.x
@@ -1,0 +1,84 @@
+// Copyright 2026 The XLS Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


hzeller:
There is the caveat, that this does not seem to work with the jit yet, this is why that '--compare=jit` is commented out.

It unhelpfully crashes with:

```
F0903 08:11:47.593204 2704354 function_converter.cc:210] Check failed: requires_opt.has_value() 
*** Check failure stack trace: ***
    @     0x5a9df4a4de74  absl::lts_20260526::log_internal::LogMessage::SendToLog()
    @     0x5a9df4a4de22  absl::lts_20260526::log_internal::LogMessage::Flush()
    @     0x5a9df13b0f3f  xls::dslx::GetRequiresImplicitToken()
    @     0x5a9df13e0b18  xls::dslx::FunctionConverter::HandleFunction()
    @     0x5a9df13996eb  xls::dslx::(anonymous namespace)::ConvertCallGraph()
    @     0x5a9df139741e  xls::dslx::ConvertModuleIntoPackage()
    @     0x5a9df139b070  xls::dslx::ConvertModuleToPackage()
    @     0x5a9df137f6a4  xls::dslx::AbstractTestRunner::ParseAndTest()
    @     0x5a9df13524d0  main
    @     0x7c700fa2b285  __libc_start_call_main
    @     0x7c700fa2b338  __libc_start_main@GLIBC_2.2.5
    @     0x5a9df135036a  _start
```

Maybe I am also holding it wrong, but it should never crash, but have a useful error message.

In any case, if the code looks good (and there is a desire for a duck type example), then I guess merging and then having a separate bug (Update: that bug is #4901) pointing to the jit situation that then can be resolved in a second step would be good.

hzeller:
Ok, I found why it crashed (a function declared in impl was never called and thus dead code, but it resulted in a crash of `F0903 08:50:49.563064 2705024 function_converter.cc:210] Check failed: requires_opt.has_value()`

I'll file a separate bug for that (once this example is in, I can easily point to this example and go from there. So will file after this example is submitted).

For now, I made the example work by actually calling each of the functions in the impl (from() and default() was not called for all tests).

hzeller:
Isolated the issue I was seeing and filed https://github.com/google/xls/issues/4901

hzeller:
#4901 is fixed, so the observation in https://github.com/google/xls/pull/4899#issuecomment-5522704014 above does not happen anymore.

Maybe this PR went under the radar ? It might be useful, but I don't know if something like this is to be covered by traits and is in the pipeline (didn't find any reference to `trait`s in `docs_src/`), so until then duck-typing might be one way to do things.

Probably something to review for @richmckeever or @erinzmoore ?

PR #4904, opened 2026-09-03
[DSLX] Implement frontend cover() autoinsertion for branches

ansidu:
Instrument each if/else-if/match arm with a cover op during DSLX -> IR conversion, where the original branch structure is still present. Covers are gated on the fully-nested reachability predicate (outer && inner), producing correct per-arm firing.

Label format `<fn>_<branch>_line_<N>_col_<M>` (function, branch kind, line/col) to make it easier to find corresponding source location even without debug info (that is also supported). Enabled via `--emit_auto_cover=branch` (default off). Warns once per function when skipped due to a missing implicit token.


ansidu:
My first attempt was to add an IR -> IR pass, but that approach doesn't work: in IR if/else/match are flattened into independent select nodes, losing which arms are nested (example 1) and which are just data-dependent (example 2).Without this information, gating conditions for cover nodes cannot be computed correctly.

Example 1:
```
    let res = if c2 {
      if c1 { ... } else { ... }
    } else { ... }
```

Example 2:
```
    let tmp = if c1 { ... } else { ... }
    let res = if c2 { tmp } else { ... }
```

xls/dslx/ir_convert/function_converter.cc
```diff
@@ -97,6 +97,37 @@ namespace {
...
+// contains the enclosing function name, the branch kind, and a line/pos; it
+// deliberately excludes any file path so the label can be used as a Verilog
+// cover property name.
+std::string CoverpointLabel(const FunctionBase* function, const Expr* node,
+                            bool then_arm) {
+  std::string_view function_name = function->name();
+  std::optional<Span> span = node->GetSpan();
+  if (span.has_value()) {
+    return absl::StrFormat(
+        "%s_%s_line_%d_pos_%d", function_name, then_arm ? "then" : "else",
```
dplassgit, line 111: `_col_` would be more readable than `_pos_` since "pos" means "position in the file" rather than "column on this line". Same comment elsewhere in this PR

dplassgit (changes requested):

xls/dslx/ir_convert/function_converter.cc
ansidu: replaced `pos` with `col`

ansidu:
I want to clarify the behavior of cover-related options. Currently, the `--emit_cover` option enables or disables cover property generation. I introduced a new option `--emit_auto_cover=...`. Should  `--emit_cover` also apply to auto-inserted cover nodes? 
In other words, should
 `ir_converter --emit_cover=false --emit_auto_cover=branch`
 disable the generation of auto-inserted cover nodes?

ansidu:
I've updated the patch to properly handle if-else if-else chains. Previously, for an `else if` two cover nodes were emitted: one for `else` and one for the nested `if` (since `else if` is structurally a nested conditional). I found this rather confusing, so the updated patch emits only one cover node for the nested `if`, skipping the node for the outer `else`.

dplassgit:
> I want to clarify the behavior of cover-related options. Currently, the `--emit_cover` option enables or disables cover property generation. I introduced a new option `--emit_auto_cover=...`. Should `--emit_cover` also apply to auto-inserted cover nodes? In other words, should `ir_converter --emit_cover=false --emit_auto_cover=branch` disable the generation of auto-inserted cover nodes?

Yes, I think `--emit_cover=false` should prevent auto-inserted cover nodes.

xls/dslx/ir_convert/function_converter.cc
```diff
@@ -1405,6 +1441,17 @@ absl::Status FunctionConverter::HandleMatch(const Match* node) {
                                            not_any_prev_selected,
                                            this_arm_selected});
           });
+      if (options_.emit_auto_cover.Contains(AutoCoverKind::kBranch)) {
```
dplassgit: As mentioned in the main PR discussion, if the auto_cover is explicitly set to false, this should not run. Or you could do it when the flags are parsed.

dplassgit (changes requested):

xls/dslx/ir_convert/function_converter.cc
ansidu: Done. Added `ShouldEmitAutoCover` method to `ConvertOptions`

dplassgit (approved):

PR #4930, opened 2026-09-06
Fix bad transformation in division simplification

meheff:
The arithmetic simplification pass checks whether `y << k` can overflow before rewriting:

```
   x / (y << k)  →  (x / y) >> k
```

But it converted the unsigned maximum shift using signed ToInt64() which is incorrect as the semantics require interpretation of the value as an unsigned number.

### DIFF
--- a/xls/passes/arith_simplification_pass.cc
+++ b/xls/passes/arith_simplification_pass.cc
@@ -1439,9 +1439,10 @@ absl::StatusOr<bool> MatchArithPatterns(int64_t opt_level, Node* n,
 
       const int64_t leading_zeros =
           query_engine.KnownLeadingZeros(y).value_or(0);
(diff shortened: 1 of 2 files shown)


meheff:
Looks like the cla error is due to an codex push which added a local email address meheff@<somerandommachine.net>.  Feel free to just clone this and submit the fix internally.

ericastor:
@meheff - we definitely shouldn't just clone & submit internally with a CLA violation flagged. Any chance you can get it fixed on your end?

ericastor:
I realized I have not looked at the actual content of this PR. As such, I'm taking this as a bug report instead & will fix it separately.

meheff:
The CLA issue is that codex added meheff@mylocalmachinename.net as a coauthor. My own CLA is still up to date. I reopened this as a new PR https://github.com/google/xls/pull/4949 without this problem.

PR #4942, opened 2026-09-09
Make rebuild golden more Posix compatible

hzeller:
* `/bin/bash` is not a location one can expect bash to be. Use `/usr/bin/env bash` instead.
  * Similarly, `/bin/true` is also not guaranteed to be there; use the built-in `true` of bash instead.


NL02 (approved):

hzeller:
make sure that the copybara review does not accidentally ends up at my old google address; @erinzmoore had observed that in my previous PR

PR #4949, opened 2026-09-09
Fix bad transformation in division simplification

meheff:
The arithmetic simplification pass checks whether y << k can overflow before rewriting:

```
   x / (y << k)  →  (x / y) >> k
```

But it converted the unsigned maximum shift using signed ToInt64() which is incorrect as the semantics require interpretation of the value as an unsigned number.


NL02:
I believe @ericastor addressed the bug and submitted it in this [commit](https://github.com/google/xls/commit/bc70bb70d81f9c9ba5828d86010b67d83e94b2f7).

PR #4951, opened 2026-09-10
Support Darwin subprocess working directories with posix_spawn

ecpeterson:
This PR was primarily authored by Codex.

Darwin builds of `xls/common/subprocess.cc` currently encounter Linux-only `memfd` support and the `/proc/self/fd` helper path. Use `posix_spawnp` directly on Darwin, with a child working-directory file action when requested. The Linux embedded-helper path and shared spawn setup retain their existing variable names and comments.

Apple's macOS 26 SDK introduces `posix_spawn_file_actions_addchdir` and deprecates the `_np` spelling. XLS still targets macOS 10.15. When the SDK declares the standard function, an availability check selects it on macOS 26 or later and retains `_np` for older supported systems. Builds whose minimum target is already macOS 26 use only the standard function; builds using older SDKs retain the older spelling. This avoids requiring a macOS 26 symbol on the earlier supported deployment targets.

The child keeps the existing argument, environment, output-capture, and timeout handling. New tests exercise PATH lookup with a child working directory, verify that the parent directory stays unchanged, and run an executable addressed relative to the child directory. Existing tests cover output capture, exit status, timeouts, and environment additions.

Validation: formatting, absolute-include checks, and `git diff --check` pass. A standalone probe of the API-selection helper compiled with deprecation and unguarded-availability warnings treated as errors for x86_64/macOS 10.15, arm64/macOS 11, and arm64/macOS 26 deployment targets; its native spawn/chdir smoke test passed. The full `//xls/common:subprocess_test` target passed all 14 tests on native Apple Silicon, including the new working-directory cases and the existing timeout/output/environment cases. Linux execution and execution on older macOS versions remain untested.

### DIFF
--- a/xls/common/BUILD
+++ b/xls/common/BUILD
@@ -318,6 +318,7 @@ cc_test(
     deps = [
         ":subprocess",
         ":xls_gunit_main",
+        "//xls/common/file:temp_directory",
         "//xls/common/status:matchers",
         "@abseil-cpp//absl/status",
         "@abseil-cpp//absl/status:status_matchers",
--- a/xls/common/subprocess.cc
+++ b/xls/common/subprocess.cc
@@ -15,7 +15,9 @@
 #include "xls/common/subprocess.h"
 
 #include <fcntl.h>
+#if !defined(__APPLE__)
 #include <linux/memfd.h>
+#endif
 #include <signal.h>  // NOLINT
 #include <spawn.h>
 #include <stdlib.h>  // NOLINT for WIFEXITED, WEXITSTATUS; not in <cstdlib>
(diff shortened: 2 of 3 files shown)


cdleary:
@ecpeterson nice contribution! just to avoid platform specific ifdefs being too mingled in the code (which we try to avoid) I made a commit that stacks on top of yours that factors out a subprocess_for_os layer with an associated bazel configuration flag for which source file to use in #4972

ecpeterson:
@cdleary back at you, nice project :) Prasahnt put me on to it, and it’s been a treat to explore. I hope union types for tagged tuples land soon!

I’ll close this in favor of your version.

PR #4952, opened 2026-09-10
Reserve fixed-latency RAM response capacity and gate accepted requests

ecpeterson:
This PR was primarily authored by Codex.

Make the existing latency-one RAM rewrites reserve enough space for each read response before issuing its request, and issue physical RAM operations only for accepted requests. This addresses two ways that a consumer stall can break the correspondence between the generated proc's channel operations and the RAM's actual operations: admitting a response with no remaining storage, and repeatedly issuing a request that is still waiting for a ready/valid handshake.

### What we were trying to build

The motivating workload is a pipelined actor runtime whose generated executors read actor state and mailbox contents from synchronous RAM. Multiple operations can overlap, while result delivery, routing, or an external sink can stall. We need accepted reads to produce exactly one response in order, with that response retained until the consumer accepts it. Stalling a response consumer must eventually stop new requests without losing work already issued to the RAM.

Using inferred synchronous RAM is important to this design: moving all state back into registers or requiring the application to forbid downstream stalls would avoid the intended architecture. The request/response adapter generated by XLS is the point where ordinary ready/valid channel semantics must be reconciled with the physical memory's fixed response timing.

### How XLS currently adapts the RAM

The relevant `Ram1RWRewrite` and `Ram1R1WRewrite` paths support latency one. They replace channel ports with physical RAM signals, delay the read-enable/valid condition by one register to produce response-valid, and put a one-entry zero-latency buffer on the response path. The buffer's upstream-ready signal is also used as the request channel's ready signal.

That reuse conflates two different questions:

- Can the buffer accept a response arriving **this cycle**?
- Can the adapter commit to a read whose response arrives **next cycle**, regardless of what the consumer does then?

An ordinary ready/valid producer can retain its output when ready is low. A latency-one RAM has already committed to returning a response once its read enable fires. Deasserting request-ready later cannot postpone that response.

### This is the shared RAM implementation used by current codegen

The helper in `xls/codegen/block_conversion.cc` is shared by codegen 1.0 for ordinary channel buffering and `RamRewritePass` for RAM responses. Codegen 1.5, which the CLI currently selects by default, explicitly reuses this same RAM pass in its post-block-conversion pipeline. Its channel-lowering pass skips ordinary I/O flops on RAM channels because the shared RAM pass owns their buffering.

Consequently, this fix applies to the RAM adapter used by the default codegen path; changing to codegen 1.5 would not select a different RAM implementation. The separate zero-latency helper in `xls/codegen_v_1_5/channel_to_port_io_lowering_pass.cc` handles ordinary channel I/O and the skid-buffer composition. That ordinary-channel helper is unchanged by this PR and does not need the RAM-only reservation mode introduced here.

The caller evidence is in [codegen version selection](https://github.com/google/xls/blob/850f8765f72622689f934a1e0841997f96a3deee/xls/tools/codegen.cc), [the v1.5 post-conversion pipeline](https://github.com/google/xls/blob/850f8765f72622689f934a1e0841997f96a3deee/xls/codegen_v_1_5/codegen.cc), and [RAM-channel handling during v1.5 I/O lowering](https://github.com/google/xls/blob/850f8765f72622689f934a1e0841997f96a3deee/xls/codegen_v_1_5/channel_to_port_io_lowering_pass.cc). The 2.0 [unified generator](https://github.com/google/xls/blob/850f8765f72622689f934a1e0841997f96a3deee/xls/codegen/unified_generator.cc) also invokes the common post-conversion pass pipeline; that call graph is not a claim of completed 2.0 RAM regression coverage. The shared RAM pass is the appropriate fix location. The new generated-proc regression runs the full codegen 1.5 pipeline and checks both RAM kinds against a one-cycle external-memory model.

### Failure 1: current capacity is mistaken for a future reservation

Suppose the response buffer is empty, read A has already been issued, and the consumer stops accepting responses just as A returns. Upstream main's `!occupied` readiness is high at the start of that cycle because the buffer is still empty. It can therefore admit read B while A is being saved into the last slot.

The following sequence isolates that reservation problem; assume each request is issued only once:

| Cycle | Event | Consequence with current-cycle readiness |
| --- | --- | --- |
| n − 1 | Issue read A with an empty response buffer. | A must return in cycle n. |
| n | A returns; consumer is stalled; read B is offered. | A fills the buffer, but request-ready still admits B using the old empty state. |
| n + 1 | B returns; consumer remains stalled. | The buffer still holds A. B cannot be backpressured at the physical RAM and has nowhere to be saved. |
| Later | Consumer resumes. | A can be delivered, but B's response was not retained. A proc waiting for that response can remain blocked. |

Adding storage without accounting for outstanding responses would only move the point at which the adapter can overcommit.

A generated-proc regression confirms that this is observable at the application output, not only in internal capacity accounting. With reads of addresses 1 through 6, a write to unrelated address 7, and a 16-cycle output stall, pinned upstream can return `A1, A2, A3, A5, A6`: the `A4` response is lost. An earlier stall can leave the proc unable to finish after returning only `A1`. The complete patch returns all six values in order and issues the expected six reads and one write. Both block interpretation and simulation of the emitted Verilog reproduce this before/after difference; the always-ready controls pass both before and after the fix.

### Failure 2: valid is mistaken for an accepted physical request

The old rewrite derives physical read enables, and the shared 1RW write enable, from request-valid and the operation's enable bit. Once request-ready can go low, a compliant producer may hold the same request valid over several cycles. Driving the RAM from valid alone then reissues that held request every cycle, despite the proc not having completed another channel transfer. Delaying that ungated read condition also creates response-valid indications for those unaccepted reads.

Correct capacity reservation therefore has to be paired with handshake gating. Fixing only the ready calculation would leave the physical RAM free to execute the work that the adapter is telling the producer has not been accepted.

### New response-capacity behavior

The shared helper gains `ZeroLatencyBufferReadyMode`. This is a C++ code-generation choice, selected by the caller when constructing block IR; it adds no runtime mode bits or hardware selector.

| Mode | Meaning of returned ready | Callers |
| --- | --- | --- |
| `kCurrentCycle` | A producer may transfer its value now when the saved slot is empty. | Ordinary channel buffers in codegen 1.0 retain upstream behavior, including their existing ready-path isolation. |
| `kNextCycle` | There will be a saved slot available for a value arriving next cycle. | Both shared latency-one RAM response rewrites. |

Next-cycle mode computes occupancy **after processing this cycle's response and consumer transfer**. It changes the meaning of the ready output wired back to the RAM request channel; the arriving response still uses the same payload buffer and bypass path. No ordinary channel or skid-buffer timing policy changes in this PR.

Ignoring reset for the following equations, let `B` mean the saved slot is occupied, `V` mean a RAM response is arriving now, and `C` mean the consumer is ready:

```text
store_response = V && !B && !C
B_next         = store_response || (B && !C)
request_ready  = !B_next
```

The adapter issues another read only if this cycle leaves the slot empty. That is enough room for the next cycle's response even if the consumer becomes unready before it arrives. If the consumer is ready when it arrives, the response can bypass the empty buffer instead.

In the example above, cycle n now saves A and deasserts request-ready immediately. B remains with the request producer. While A is held, no new read is accepted. In the cycle that the consumer finally removes A, the slot becomes available and B can be accepted; B returns one cycle later and can be delivered or saved safely. The adapter does not need to predict future consumer readiness.

With reset establishing `B = V = 0`, accepted-read gating maintains `B_next && V_next == 0`: a saved response and another arriving response cannot coexist. When saved response A is consumed, request B may be accepted immediately, and B's response arrives on the following cycle. The existing payload-load and saved-valid updates therefore remain sufficient.

This is not a blanket rule to wait a cycle after every response. With an empty buffer and a continuously ready consumer, each arriving response bypasses, `B_next` stays zero, and the adapter can accept a new read every cycle. It reserves storage when needed without adding a RAM-latency stage or unconditionally serializing steady-state requests.

### New physical-request behavior

The rewrite uses the final ready signal to determine which request actually fires:

```text
read_fire             = request_valid && read_enable && request_ready
write_fire_1RW        = request_valid && write_enable && request_ready
response_valid(t + 1) = read_fire(t)
```

Those fire signals drive the physical enables, and only an accepted read creates a response-valid indication. A held request produces no repeated RAM operations while stalled, then issues once when accepted. For 1RW, reads and writes share the request channel, so both physical enables respect its handshake. The independent 1R1W write channel remains on its existing always-ready path; this PR does not redesign its hazard policy or write-completion protocol.

### Area, throughput, and impact on other designs

The change affects each configured latency-one 1RW or 1R1W RAM adapter, including adapters generated by codegen 1.5. It is not conditional on selecting `zerolatency` for ordinary channel I/O: the RAM rewrite inserts its own response buffer. Designs without these RAM configurations do not acquire reservation logic from this PR. The source identifies the affected configurations; it does not provide user-population statistics or representative hardware instance counts.

| Aspect | Expected effect and tradeoff |
| --- | --- |
| Storage area and latency | The existing one-entry payload buffer, saved-valid register, and delayed response-valid register are retained. No additional queue, RAM macro, or register stage is introduced. |
| Control area and timing | Computing next occupancy and gating accepted operations adds combinational control per adapter. The request-ready/physical-enable path now depends on the arriving response's valid and consumer-ready signals. Logic mapping, fanout, and this longer dependency path may affect area and clock frequency; the seeded ASAP7 area/delay estimates below measure that cost on focused controllers. They do not measure FPGA LUTs or placed timing. |
| Steady read throughput | With a ready consumer, the adapter can still accept one read per cycle. This cycle-level capacity does not guarantee unchanged maximum clock frequency or measured system throughput. |
| Response stalls | Requests can be refused a cycle earlier to reserve the final slot. This intentionally reduces admission compared with the overcommitting implementation. More outstanding work during stalls would require additional response capacity and corresponding reservation accounting. |
| Shared 1RW writes | Reads and writes use the same ready signal. A saved read response can therefore delay a write-only request, even though that write needs no read-response slot. The existing shared-ready coupling remains, and earlier reservation can start that backpressure sooner. Per-operation admission would be a separate optimization. |
| Independent 1R1W writes | The existing always-ready write-request path is unchanged; stalled reads do not gain a new connection to that write-ready path. |

Gating also prevents the repeated physical operations previously caused by a request held valid while stalled. That may reduce unnecessary switching, but no power improvement is claimed. The benefit established by the design is transaction correctness with fixed response capacity. Area, timing, and workload throughput must be measured separately; correctness does not make those costs negligible.

### Why this addresses the workload

Together, reservation and handshake gating give the adapter a consistent accounting boundary: every accepted read reserves a place for its inevitable response, and no unaccepted read creates another response outside that accounting. The runtime can allow response-side stalls while retaining the RAM-backed, pipelined implementation. The intended result is preservation of accepted transactions, not merely a higher benchmark rate.

The change is limited to the existing latency-one rewrites. It does not claim support for arbitrary response latency, extra outstanding-request queues, or complete-system timing closure. This is a standalone, single-commit PR against `main`. Its production diff adds next-occupancy reporting to the shared helper and handshake gating to the RAM rewrite; it preserves the existing ordinary-buffer payload, valid, and ready behavior. No other buffer change is required.

### Validation

Validation ran natively on Apple Silicon, using upstream `850f8765f72622689f934a1e0841997f96a3deee` as the baseline. The native builds included the independent Darwin subprocess patch on local validation branches; that host-support change is not part of this PR and does not alter the generated hardware.

- The complete `//xls/codegen:ram_rewrite_pass_test` target passes: 76 tests passed and 9 intentionally skipped. This includes the two direct handshake tests and the two new generated-proc cases, one for each RAM kind. The direct fixtures now run port legalization before integer-valued interpretation, removing empty-tuple mask ports as the normal codegen pipeline does.
- The generated-proc cases use the full default codegen 1.5 pipeline. A compliant command source holds its input until accepted; a memory model returns data exactly one cycle after each physical read; the output stalls for 16 cycles at six different starting points. An always-ready control is included. The checks require all commands to complete, exact physical read/write sequences, correct write data, and every read result exactly once and in order after draining.
- Those same generated-proc cases fail on pinned upstream. In that baseline, the emitted Verilog also fails 10 of 14 equivalent Icarus Verilog scenarios and passes the other 4, including both always-ready controls. All 14 emitted-Verilog scenarios pass with this patch.
- The 81 pre-existing RAM cases pass on pinned upstream: 72 passed and 9 intentionally skipped. Thus that suite alone does not detect this response-loss scenario. The new generated-proc regression supplies the missing coverage.
- The direct response-buffer fixture checks reservation and recovery; the 1R1W fixture checks a held read; the 1RW fixture checks both a held read and a held write. The two direct handshake tests fail on the baseline and pass with this patch.
- A downstream D1 RAM-backed actor-topology smoke check also passes after regenerating RTL with the patched native codegen executable. It starts from the project's saved optimized-IR fixture; DSLX conversion and optimization are not part of this check.
- `ZeroLatencyBufferReservesNextCycleRamResponse` passes on the submitted standalone change. Formatting, absolute-include checks, and `git diff --check` pass.
- A downstream D3 decoder profile with eight decoupled schedulers, two pipeline stages, and II=1 completes at 174 cycles per step with this fix. The measurement covers 24 step intervals after warmup. RTL freshly generated by the submitted standalone implementation is byte-identical to the profiled RTL. This is a cycle-level observation, not a measurement of attainable clock frequency.

Linux execution remains pending. The synthesis comparison below reports mapped cell area and estimated combinational delay; power, placed timing, and device throughput have not been measured. The tests establish transaction preservation during stalls, not unchanged maximum clock frequency, write latency under every workload, or complete-system timing closure.

### Reading these measurements for this PR

For the standalone correction, the 1RW controller changes best mapped area by **-16.02%** and best estimated delay by **+20.58%** relative to main. The 1R1W controller changes those metrics by **+1.67%** and **+0.12%**, respectively. The baseline fails the stall regressions described above; this comparison quantifies the synthesis effects of the correction. Improvements or regressions below the stated noise thresholds should not be treated as established conclusions.

The ordinary-channel cases serve as controls and are unchanged by this correction. RAM controller flop counts remain 37 for 1RW and 38 for 1R1W in the measured examples; the differences come from combinational mapping rather than added response storage.

### Seeded ASAP7 synthesis comparison

Following [XLS's PPA submission guidance](https://github.com/google/xls/blob/main/CONTRIBUTING.md#documenting-ppa-impact), each circuit/variant received 50 runs with `--autoidx=100000*s` and `--hash-seed=s`, for `s=1..50`. Those Yosys-only runs converged to identical mapped netlists within each circuit/variant. We therefore added 50 runs with XLS `--randomize_order_seed=s`, still varying both Yosys seeds. The Verilog-order sweep produces actual mapping variation. The first two tables report the best of all **100 runs per circuit/variant: 1,400 successful syntheses across the two reported variants**. The distribution tables below give best, mean, sample variance, and worst separately for each 50-run family and for the pooled 100 runs. Both families and every seed are retained in the report data.

The baseline is upstream `850f8765f72622689f934a1e0841997f96a3deee`; the patched version is this PR at `05a16983f3e9ffcbf4a59abf39e5029ff6817dd6`. These are the exact source revisions used for the measurements, rather than a moving `main`. The independent Darwin subprocess support is present only in native validation builds and does not change emitted hardware.

The flow uses native Yosys 0.69+10 (`370a53acf`, OSS CAD Suite 2026-09-10) and its ABC, with ASAP7 rev28 RVT slow/slow NLDM libraries pinned to `d88477438935a5a388bd6294f682dc405c93c5d2`. It follows the pinned `rules_hdl` synthesis/ABC pass sequence, translated from Tcl to Yosys commands because this native binary lacks the Tcl frontend. It uses the same ASAP7 excluded-cell patterns and tie cells, with a `BUFx2_ASAP7_75t_R` input driver and 1 fF output loads, no specified clock-period target, no adder-specific mapping, and an added final `stime -p` after ABC sizing. Each accepted run passes `check -assert`, contains no remaining generic `$_*` cells, and records library-mapped area.

The guidelines specify a measurement procedure, not a fixed benchmark suite. The two RAM controllers exercise both edited RAM rewrite paths. Ordinary zero-latency buffers in both codegen implementations and a skid-buffer case serve as unaffected controls. They do not establish effects on every XLS workload or how frequently users select these configurations.

The ordinary-channel tests are pass-through procs with one output buffer and no input buffer; the four-buffer cases connect four emitted modules in series before flattening. RAM tests use the generated-proc regression's controller: 4-bit addresses, 32-bit returned data, a read/write command bit, constant `0x55` write data, and no masks. They use two pipeline stages, latency-one 1RW or 1R1W RAM adaptation, and no ordinary I/O flops. External RAM storage is outside the synthesized controller boundary. The baseline RAM implementations are incorrect under the separately tested output stalls; their synthesis numbers quantify the correction rather than establish functional equivalence.

**Best mapped cell area (µm²; lower is better):**

| Circuit | Baseline | Patched |
| --- | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | 36.129 | 36.129 |
| Four 128-bit zero-latency buffers, codegen 1.0 | 421.202 | 421.202 |
| 32-bit zero-latency buffer, codegen 1.5 | 36.129 | 36.129 |
| Four 128-bit zero-latency buffers, codegen 1.5 | 421.202 | 421.202 |
| Four 128-bit skid buffers, codegen 1.5 | 866.810 | 866.810 |
| 32-bit 1RW RAM controller | 44.425 | 37.310 |
| 32-bit 1R1W RAM controller | 45.300 | 46.058 |

**Best final ABC delay estimate (ps; lower is better):**

| Circuit | Baseline | Patched |
| --- | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | 161.150 | 161.150 |
| Four 128-bit zero-latency buffers, codegen 1.0 | 278.620 | 278.620 |
| 32-bit zero-latency buffer, codegen 1.5 | 161.150 | 161.150 |
| Four 128-bit zero-latency buffers, codegen 1.5 | 278.620 | 278.620 |
| Four 128-bit skid buffers, codegen 1.5 | 236.830 | 236.830 |
| 32-bit 1RW RAM controller | 193.710 | 233.570 |
| 32-bit 1R1W RAM controller | 194.950 | 195.190 |

The area and delay tables are separate minima and may come from different seeds. They do not describe one simultaneously optimal implementation; the full data also records delay at the minimum-area seed. Delay is ABC's final combinational cell-delay estimate under the stated load/driver assumptions, not post-placement timing, a maximum clock frequency, or an FPGA result. Power and external RAM macro area are not measured. Differences below XLS's stated 2% area / 10% delay noise thresholds are not claimed as established improvements.

**D3 boundary:** both baseline and patched full D3 designs fail Yosys's structural `check -assert` with the same two reported logic-loop problems in effect-window/router/FIFO readiness. The exploratory full-decoder mapping was stopped after ABC cut those feedback paths. No full-D3 area/delay result is claimed; the tables cover the seven loop-free focused circuits above. The separate patched RTL throughput profile completes at 174 cycles per step as described above. It does not establish physical timing closure.

#### Distribution across synthesis seeds

Each row describes repeated synthesis of one circuit/variant, with equal weight per run. Mean is the arithmetic mean. Sample variance is `sum((x - mean)^2) / (n - 1)`, in µm⁴ for area and ps² for delay. Best and worst mean the smallest and largest observed values, not guaranteed bounds. These describe sensitivity to the sampled synthesis/order seeds, not silicon process variation, confidence intervals, or a distribution over XLS workloads. Statistics use the unrounded recorded results; displayed values are rounded.

**Yosys seeds plus randomized Verilog order — 50 runs per row.** This family exposes mapping variability that the fixed-order runs did not. The best in these tables is restricted to these 50 runs, so it can differ from the 100-run minima above.

**Mapped cell area:**

| Circuit | Variant | Best (µm²) | Mean (µm²) | Sample variance (µm⁴) | Worst (µm²) |
| --- | --- | ---: | ---: | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | Baseline | 36.144 | 36.466 | 0.034462 | 36.640 |
| 32-bit zero-latency buffer, codegen 1.0 | Patched | 36.144 | 36.466 | 0.034462 | 36.640 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Baseline | 421.202 | 427.151 | 5.88332 | 428.171 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Patched | 421.202 | 427.151 | 5.88332 | 428.171 |
| 32-bit zero-latency buffer, codegen 1.5 | Baseline | 36.144 | 36.471 | 0.0307078 | 36.640 |
| 32-bit zero-latency buffer, codegen 1.5 | Patched | 36.144 | 36.471 | 0.0307078 | 36.640 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Baseline | 421.202 | 427.422 | 4.39094 | 428.171 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Patched | 421.202 | 427.422 | 4.39094 | 428.171 |
| Four 128-bit skid buffers, codegen 1.5 | Baseline | 866.810 | 876.336 | 55.8523 | 888.141 |
| Four 128-bit skid buffers, codegen 1.5 | Patched | 866.810 | 876.336 | 55.8523 | 888.141 |
| 32-bit 1RW RAM controller | Baseline | 44.425 | 44.489 | 0.00412042 | 44.586 |
| 32-bit 1RW RAM controller | Patched | 37.310 | 37.371 | 0.00276783 | 37.514 |
| 32-bit 1R1W RAM controller | Baseline | 45.300 | 45.446 | 0.00227292 | 45.562 |
| 32-bit 1R1W RAM controller | Patched | 46.058 | 46.231 | 0.00132196 | 46.321 |

**Final ABC delay estimate:**

| Circuit | Variant | Best (ps) | Mean (ps) | Sample variance (ps²) | Worst (ps) |
| --- | --- | ---: | ---: | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | Baseline | 161.150 | 162.469 | 1.34638 | 164.410 |
| 32-bit zero-latency buffer, codegen 1.0 | Patched | 161.150 | 162.469 | 1.34638 | 164.410 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Baseline | 278.620 | 282.756 | 23.04 | 293.760 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Patched | 278.620 | 282.756 | 23.04 | 293.760 |
| 32-bit zero-latency buffer, codegen 1.5 | Baseline | 161.150 | 162.531 | 1.30416 | 164.410 |
| 32-bit zero-latency buffer, codegen 1.5 | Patched | 161.150 | 162.531 | 1.30416 | 164.410 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Baseline | 278.620 | 282.275 | 17.3001 | 293.760 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Patched | 278.620 | 282.275 | 17.3001 | 293.760 |
| Four 128-bit skid buffers, codegen 1.5 | Baseline | 236.830 | 238.674 | 1.53277 | 240.810 |
| Four 128-bit skid buffers, codegen 1.5 | Patched | 236.830 | 238.674 | 1.53277 | 240.810 |
| 32-bit 1RW RAM controller | Baseline | 193.710 | 193.710 | 0 | 193.710 |
| 32-bit 1RW RAM controller | Patched | 233.570 | 238.639 | 5.48653 | 244.370 |
| 32-bit 1R1W RAM controller | Baseline | 194.950 | 194.980 | 0.000297796 | 194.990 |
| 32-bit 1R1W RAM controller | Patched | 195.190 | 198.963 | 0.29645 | 199.040 |

<details>
<summary>Fixed Verilog order, varying Yosys seeds: all statistics (50 runs per row)</summary>

All 50 runs produced the same mapped netlist within each circuit/variant. Thus the mean, best, and worst coincide, and both sample variances are exactly zero. This is an observation about this seed family, not a general claim that mapping is deterministic under other perturbations.

| Circuit | Variant | Area: best = mean = worst (µm²) | Area variance (µm⁴) | Delay: best = mean = worst (ps) | Delay variance (ps²) |
| --- | --- | ---: | ---: | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | Baseline | 36.129 | 0 | 163.610 | 0 |
| 32-bit zero-latency buffer, codegen 1.0 | Patched | 36.129 | 0 | 163.610 | 0 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Baseline | 421.202 | 0 | 293.760 | 0 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Patched | 421.202 | 0 | 293.760 | 0 |
| 32-bit zero-latency buffer, codegen 1.5 | Baseline | 36.129 | 0 | 163.610 | 0 |
| 32-bit zero-latency buffer, codegen 1.5 | Patched | 36.129 | 0 | 163.610 | 0 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Baseline | 421.202 | 0 | 293.760 | 0 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Patched | 421.202 | 0 | 293.760 | 0 |
| Four 128-bit skid buffers, codegen 1.5 | Baseline | 886.479 | 0 | 238.270 | 0 |
| Four 128-bit skid buffers, codegen 1.5 | Patched | 886.479 | 0 | 238.270 | 0 |
| 32-bit 1RW RAM controller | Baseline | 44.425 | 0 | 193.710 | 0 |
| 32-bit 1RW RAM controller | Patched | 37.354 | 0 | 238.770 | 0 |
| 32-bit 1R1W RAM controller | Baseline | 45.446 | 0 | 194.990 | 0 |
| 32-bit 1R1W RAM controller | Patched | 46.219 | 0 | 199.040 | 0 |

</details>

<details>
<summary>Pooled statistics for all 100 runs per circuit/variant</summary>

These include both 50-run families and correspond to the same population of recorded runs used for the headline minima. They give each family 50% weight by construction. Because half the runs repeat an identical fixed-order result, pooled means and variances should be read alongside the separate family statistics; they do not estimate the distribution under Verilog-order randomization alone.

**Mapped cell area:**

| Circuit | Variant | Best (µm²) | Mean (µm²) | Sample variance (µm⁴) | Worst (µm²) |
| --- | --- | ---: | ---: | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | Baseline | 36.129 | 36.297 | 0.045652 | 36.640 |
| 32-bit zero-latency buffer, codegen 1.0 | Patched | 36.129 | 36.297 | 0.045652 | 36.640 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Baseline | 421.202 | 424.177 | 11.8514 | 428.171 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Patched | 421.202 | 424.177 | 11.8514 | 428.171 |
| 32-bit zero-latency buffer, codegen 1.5 | Baseline | 36.129 | 36.300 | 0.0446929 | 36.640 |
| 32-bit zero-latency buffer, codegen 1.5 | Patched | 36.129 | 36.300 | 0.0446929 | 36.640 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Baseline | 421.202 | 424.312 | 11.9453 | 428.171 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Patched | 421.202 | 424.312 | 11.9453 | 428.171 |
| Four 128-bit skid buffers, codegen 1.5 | Baseline | 866.810 | 881.408 | 53.6196 | 888.141 |
| Four 128-bit skid buffers, codegen 1.5 | Patched | 866.810 | 881.408 | 53.6196 | 888.141 |
| 32-bit 1RW RAM controller | Baseline | 44.425 | 44.457 | 0.00306924 | 44.586 |
| 32-bit 1RW RAM controller | Patched | 37.310 | 37.363 | 0.00144724 | 37.514 |
| 32-bit 1R1W RAM controller | Baseline | 45.300 | 45.446 | 0.00112507 | 45.562 |
| 32-bit 1R1W RAM controller | Patched | 46.058 | 46.225 | 0.000692183 | 46.321 |

**Final ABC delay estimate:**

| Circuit | Variant | Best (ps) | Mean (ps) | Sample variance (ps²) | Worst (ps) |
| --- | --- | ---: | ---: | ---: | ---: |
| 32-bit zero-latency buffer, codegen 1.0 | Baseline | 161.150 | 163.039 | 0.995378 | 164.410 |
| 32-bit zero-latency buffer, codegen 1.0 | Patched | 161.150 | 163.039 | 0.995378 | 164.410 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Baseline | 278.620 | 288.258 | 41.9836 | 293.760 |
| Four 128-bit zero-latency buffers, codegen 1.0 | Patched | 278.620 | 288.258 | 41.9836 | 293.760 |
| 32-bit zero-latency buffer, codegen 1.5 | Baseline | 161.150 | 163.071 | 0.939277 | 164.410 |
| 32-bit zero-latency buffer, codegen 1.5 | Patched | 161.150 | 163.071 | 0.939277 | 164.410 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Baseline | 278.620 | 288.018 | 41.8698 | 293.760 |
| Four 128-bit zero-latency buffers, codegen 1.5 | Patched | 278.620 | 288.018 | 41.8698 | 293.760 |
| Four 128-bit skid buffers, codegen 1.5 | Baseline | 236.830 | 238.472 | 0.799902 | 240.810 |
| Four 128-bit skid buffers, codegen 1.5 | Patched | 236.830 | 238.472 | 0.799902 | 240.810 |
| 32-bit 1RW RAM controller | Baseline | 193.710 | 193.710 | 0 | 193.710 |
| 32-bit 1RW RAM controller | Patched | 233.570 | 238.704 | 2.7199 | 244.370 |
| 32-bit 1R1W RAM controller | Baseline | 194.950 | 194.985 | 0.000170667 | 194.990 |
| 32-bit 1R1W RAM controller | Patched | 195.190 | 199.001 | 0.148225 | 199.040 |

</details>


allight:
Minor: The first line of the PR message should describe (simply) what the CL is doing, it's the only part actually displayed by default.

Don't have that be "this was written by codex".

It may take a while to review this since its rather large.

PR #4954, opened 2026-09-10
dslx: typecheck semantic sum constructors

dank-openai:
## Summary

- Typecheck unit, tuple, and named semantic sum constructors while preserving their parsed AST nodes.
- Infer constructor parametrics from payloads and context, including imported types and declared defaults.
- Report invalid arity, payload shape, field names, and payload types at the constructor.

## Why This Change

A payload-carrying enum constructor establishes both the variant and the sum type. For example:

```dslx
#![feature(type_inference_v2)]
enum MaybeU32 { None, Some(u32) }
const X = MaybeU32::Some(u32:7);
```

`X` has type `MaybeU32`, with the `Some` variant's `u32` payload. The same typechecking path handles named payloads and generic sums; constructor calls with the wrong payload or an incompatible type produce type errors. Constructor expressions retain their parsed AST nodes: tuple constructors are marked `Invocation` nodes, with unit and named forms represented by `ColonRef` and `StructInstance`.

This is initial constructor support. Constructor names are not first-class values, and semantic-sum matching, lowering, and aggregate payloads remain outside this slice. Supported payload members are bits-like values, enums, and empty semantic sums until aggregate flattening supports more types.

## Tests

The semantic-sum typechecking and type-information serialization suites pass on Ubuntu 24.04. They cover constructor payload substitution, type inference, diagnostics, and source-checked type-info round trips:

```sh
bazel test \
  //xls/dslx/type_system_v2:typecheck_module_v2_sum_test \
  //xls/dslx/type_system:type_info_to_proto_test
```

More thorough fuzzing passes, that fuzz testing will appear in a later PR.


xls/dslx/type_system/type.h
```diff
@@ -846,15 +846,37 @@ class SumTypeVariant {
 // variant.
 class SumType : public Type {
  public:
+  struct UnknownZeroSelection {};
```
richmckeever: Should these options just be an enum?

xls/dslx/type_system_v2/inference_table_converter_impl.cc
```diff
@@ -103,6 +105,28 @@ bool NeedsMetaType(const InferenceTable& table, const AstNode* node) {
           node->parent()->kind() == AstNodeKind::kTypeAlias);
 }
 
+absl::Status ValidatePhase1SumPayloadMemberType(
```
richmckeever, line 108: Move to some *utils file in the directory that already depends on type.h, or create one for sum types if you prefer.

xls/dslx/type_system_v2/inference_table_converter_impl.cc
```diff
@@ -1304,6 +1316,17 @@ class InferenceTableConverterImpl : public InferenceTableConverter,
       }
     }
 
+    if (node->kind() == AstNodeKind::kMatch) {
```
richmckeever, line 1319: This is temporary, right? Could you add a TODO and/or say "not supported yet" in the error?

xls/dslx/type_system_v2/inference_table_converter_impl.cc
```diff
@@ -1315,6 +1338,21 @@ class InferenceTableConverterImpl : public InferenceTableConverter,
       ti->SetItem(node, **type);
     }
 
+    if (node->kind() == AstNodeKind::kBinop) {
```
richmckeever, line 1341: I assume we need this because ==/!= on sum types is going to generate asserts? Should we have a helper function like IsBinopWithGeneratedAsserts to identify the relevant nodes, in order to get some complexity out of this function and make the scenario more self-documenting?

xls/dslx/type_system_v2/sum_instance_canonicalizer.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace xls::dslx {
+
+// Clones `module` with resolved sum-constructor expressions rewritten to the
+// dedicated `SumInstance` expression form, including nested payloads, in one
+// clone traversal that preserves lexical definition identity. Returns `nullopt`
+// when no constructor expressions need canonicalization. Imports must already
+// be available in `import_data`. Normalize parsed modules before preparation or
+// type inference publishes their nodes.
+absl::StatusOr<std::optional<std::unique_ptr<Module>>> CanonicalizeSumInstances(
```
richmckeever, line 33: A good amount of the complexity in the PR seems to revolve around this notion of replacing invocations with different nodes via cloning.

The real long-term direction in the frontend is to do less (no?) cloning in favor of editing/elaborating the tree as we go.

Obviously we need a more homogeneous structure, more like in IR, to apply this principle wholesale. In the meantime, we've been enabling relatively surgical edits.

So I think the question here is: have you tried just having a flag or optional SumDef pointer or something in Invocation, which "canonicalization" just edits without changing the nature of the node? Could we then avoid messing around with the cloner etc, with an overall gain in simplicity compared to this version? Then long term, we'd use a ReplaceWith() style primitive instead of keeping a modified Invocation node in place.

richmckeever (changes requested):

xls/dslx/type_system_v2/sum_instance_canonicalizer.h
dank-openai: I'll try exactly your suggestion. 

For what it's worth, before making this PR, I did spend a considerable amount of time trying to make a more general cloning function (which is apparently the opposite of what you wanted here :) ), but there were too many edge cases to handle it all correctly.

xls/dslx/type_system/type.h
dank-openai: Combined the two empty marker types into `ZeroSelectionState`, with `kUnspecified` and `kNoZeroVariant`. The selected `SumVariant` reference remains the other `std::variant` alternative. The existing type, zero-value, and sum-typechecking tests pass.

xls/dslx/type_system_v2/inference_table_converter_impl.cc
dank-openai: Moved `ValidatePhase1SumPayloadMemberType` to the existing `inference_table_utils.h/.cc`, which already depend on `type.h`. The validator logic, diagnostic, and TODO are unchanged. All 112 cases in `typecheck_module_v2_sum_test` pass.

xls/dslx/type_system_v2/inference_table_converter_impl.cc
dank-openai: Yes, this is temporary. This PR adds semantic-sum construction; matching will follow in a separate PR. I've changed the error to "not supported yet" and added a `TODO(dank-openai)` at the rejection.

xls/dslx/type_system/type.h
dank-openai: `SumType::ZeroSelection` is now `std::variant<SelectedZeroVariant, NoZeroVariant>`, with both cases declared inside `SumType`. Callers must explicitly supply `SelectedZeroVariant{std::cref(variant)}` or `NoZeroVariant{}`; `kUnspecified` is removed. `NoZeroVariant` is the only payload-free case, so `ZeroSelectionState` is unnecessary.

Diagnostic formatting now reads the dumped type description directly, so it no longer constructs an incomplete `SumType`. The serialized fields are unchanged. All nine affected test targets pass.

xls/dslx/type_system_v2/inference_table_converter_impl.cc
dank-openai: Extracted `IsComparisonRequiringImplicitToken` into the anonymous namespace and documented the operand-validity assertions that motivate the token. Token marking is unchanged; this PR records the requirement, with the corresponding sum-comparison lowering outside its scope.

xls/dslx/type_system_v2/sum_instance_canonicalizer.h
dank-openai: Implemented your suggestion: constructor expressions stay in the original parsed tree. For example, `E::V(x)` remains the same `Invocation`, marked `kSumConstructor`; unit and named forms retain their `ColonRef` and `StructInstance` nodes.

Removed the constructor canonicalizer and its module-cloning step. A borrowed `SumConstructorView` gives typechecking and parametric inference access to the original payload expressions. The conversion-order walker still visits payload arguments, then skips function resolution for marked constructors.

Added a regression that checks that typechecking preserves the original module, declarations, constructor expressions, and nominal sum definition by pointer identity. It failed on the previous implementation and passed with the refactor. The affected Google test matrix passed all 33 targets before the final comment and helper-visibility cleanup. After that cleanup, formatting, whitespace, and source checks passed; tests and builds were not rerun.

xls/dslx/frontend/ast.h
```diff
@@ -596,7 +603,8 @@ class TypeRefTypeAnnotation : public TypeAnnotation {
   TypeRefTypeAnnotation(
       Module* owner, Span span, TypeRef* type_ref,
       std::vector<ExprOrType> parametrics,
-      std::optional<const StructInstanceBase*> instantiator = std::nullopt);
+      std::optional<const StructInstanceBase*> instantiator = std::nullopt,
```
richmckeever: It would be good to rename this to `struct_instantiator`.

xls/dslx/type_system_v2/populate_table_visitor.cc
```diff
@@ -275,6 +291,21 @@ class PopulateInferenceTableVisitor : public PopulateTableVisitor,
         return table_.SetTypeAnnotation(node, type_ref_annotation);
       }
 
+      XLS_ASSIGN_OR_RETURN(std::optional<SumRef> sum_type_ref,
+                           GetSumRef(node, import_data_));
+      if (sum_type_ref.has_value()) {
+        TypeAnnotation* type_ref_annotation =
```
richmckeever: I think what this is doing is more static resolution than we have for structs/enums, and perhaps it would be inadequate for something like

```
fn foo<T: type>() -> T {
  T::SOME_MEMBER
} 
```

Structs and enums don't have a counterpart to SumAnnotation; we keep the TRTA in play for longer. I think we've found lately that, if anything, we need to move in the direction of deferring more (especially in populate_table_visitor.cc) and doing more with the ParametricContext known.

xls/dslx/type_system_v2/import_utils.cc
```diff
@@ -239,14 +321,190 @@ GetTypeVariableTypeAnnotationForSubject(const ColonRef* ref,
...
+    return (*instance)->GetUnorderedMembers();
+  } else if (const auto* instance =
+                 std::get_if<const SumInstance*>(&expression_)) {
+    return (*instance)->struct_payload_field_args();
+  } else {
+    return {};
+  }
+}
+
+absl::Status ClassifySumConstructors(AstNode* root,
```
richmckeever: Two points about this.

For an edge case like:

```
fn f<T: type>() -> T {
  T::Member(stuff)
}
```

I think you need to be in ConvertInvocation or its helpers, in order to correctly decide what T::Member() is "invoking", whether it's a sum type, and which one. We could make this error initially, since generics are kind of a WIP. But I think we should consider this (and the struct-format counterpart) in deciding the phasing of the logic.

Maybe we actually want this classification to be in ResolveColonRefTarget in inference_table_converter_impl.cc? Which admittedly would mean my suggestion of adding members in Invocation may have implied too static of an analysis. Hopefully putting it in ResolveColonRefTarget (or similar) would lead to the simplest and most powerful version of the PR yet, but if it's a large can of worms, we could fall back on just not supporting examples like the above until later.

Second point: if we keep this logic around, it's more standard to use an AstNodeRecursiveVisitor, or if invoked from an existing one of those (as in the case of populate_table_visitor.cc), to work it into the handler functions (HandleInvocation, HandleStructInstance, etc.) already there. That way the flow is clearer and there aren't forced/hidden "extra descents".

richmckeever (changes requested):

xls/dslx/frontend/ast.h
dank-openai: Renamed this to `struct_instantiator` throughout the constructor parameter, member, accessor and direct callers, including the corresponding field in the type-reference resolver.

xls/dslx/type_system_v2/import_utils.cc
dank-openai: For the generic-construction point, I believe supporting construction through a generic type would be a large change that does not belong in this PR. I made `T::Member(...)` and its struct-format counterpart produce an explicit "not supported yet" error at the original expression instead. I added coverage for both forms, the named splat form, and aliases; the sum and generics typechecking suites passed. Your separate point about traversal in this PR is still open.

dank-openai: Good point. I removed the five extra constructor walks from inference. I kept one module-wide pass using `AstNodeRecursiveVisitor` because the warning for useless expressions needs to distinguish a discarded constructor such as `E::V(u32:0);` from a potentially effectful function call. During inference, the existing invocation and brace-construction handlers now classify just the node they're visiting, including when the earlier pass hasn't seen it.

I added tests for a constructor inside qualified type syntax and for a constructor deliberately left unclassified before inference, to check that it isn't treated as a function dependency. All seven targeted Bazel test suites passed on an Ubuntu devbox.

xls/dslx/frontend/ast.h
```diff
@@ -164,6 +164,13 @@ XLS_DSLX_AST_NODE_EACH(FORWARD_DECL)
 class StructDefBase;
 class StructInstanceBase;
 
+// Borrowed constructor syntax used as evidence for sum parametric inference.
```
richmckeever: Is this comment a relic from an earlier rev? I don't get how it applies to the variant declaration here.

xls/dslx/type_system_v2/flatten_in_type_order.cc
```diff
@@ -76,15 +76,71 @@ class Flattener : public AstNodeVisitorWithDefault {
...
-    XLS_RETURN_IF_ERROR(node->struct_ref()->Accept(this));
-    for (const auto& [_, member] : node->GetUnorderedMembers()) {
-      XLS_RETURN_IF_ERROR(member->Accept(this));
+    XLS_ASSIGN_OR_RETURN(
+        std::optional<SumConstructorRef> constructor,
+        ResolveSumConstructor(node->struct_ref(), import_data_));
+    if (constructor.has_value()) {
+      return HandleSumConstructor(node, *constructor);
+    } else {
+      XLS_RETURN_IF_ERROR(node->struct_ref()->Accept(this));
```
richmckeever: Unnest the else branch.

xls/dslx/type_system_v2/flatten_in_type_order.cc
```diff
@@ -76,15 +76,71 @@ class Flattener : public AstNodeVisitorWithDefault {
...
+                         ResolveSumConstructor(node, import_data_));
+    if (constructor.has_value() && constructor->variant->is_unit()) {
+      return HandleSumConstructor(node, *constructor);
+    } else {
+      return DefaultHandler(node);
+    }
+  }
+
+  absl::Status HandleSumInstance(const SumInstance* node) override {
+    XLS_ASSIGN_OR_RETURN(
```
richmckeever: This handler seems redundant, i.e. the natural descent into node->constructor_ref() would cause HandleColonRef to take care of it.

xls/dslx/type_system_v2/import_utils.cc
```diff
@@ -239,14 +320,236 @@ GetTypeVariableTypeAnnotationForSubject(const ColonRef* ref,
...
+  return TypeInferenceErrorStatus(
+      ref.span(), nullptr,
+      absl::Substitute("Sum constructor `$0` through a generic type is not "
+                       "supported yet.",
+                       ref.ToString()),
+      file_table);
+}
+
+const Expr* SumConstructorView::expression() const {
+  return std::visit([](const auto* node) -> const Expr* { return node; },
```
richmckeever: return ToAstNode(expression_);

xls/dslx/type_system_v2/inference_table_converter_impl.cc
```diff
@@ -245,9 +265,11 @@ class InferenceTableConverterImpl : public InferenceTableConverter,
                   node->parent()->kind() == AstNodeKind::kImpl))));
     for (const AstNode* node : nodes) {
       VLOG(5) << "Next node: " << node->ToString();
-      if (node->kind() == AstNodeKind::kInvocation) {
+      if (const auto* invocation = dynamic_cast<const Invocation*>(node);
```
richmckeever: We've basically banned dynamic_cast in here, in favor of kind() checks + down_cast, because it was found to have a major impact on performance.

xls/dslx/frontend/ast.h
```diff
@@ -164,6 +164,13 @@ XLS_DSLX_AST_NODE_EACH(FORWARD_DECL)
...
+// The owning module keeps these expressions alive; inferred parametrics belong
+// to the inference context rather than to these shared syntax nodes.
+using SumConstructorExpr =
+    std::variant<const ColonRef*, const Invocation*, const StructInstance*,
+                 const SumInstance*>;
```
richmckeever: Do we even still need SumInstance to exist as a class? I think the idea was to basically replace it with the other things in this variant.

PR #4961, opened 2026-09-11
Add LoopIdiomPass

cluzze:
Adds a new pass - loop-idiom. For now just recognizes zero-filled array shift (DSLX `data << shift` expressed as a `for` loop with `result[i] = if i < shift { false } else { data[i-shift] }`). The counted_for body lowers to an `array_update` of `sel(ult(i, shift), [array_index(data, [i - shift]), 0])`; rewriting that to a packed `shll` + per-bit `bit_slice` shape lets codegen emit a native shift instead of per-element index muxes.
I measured PPA on this simple example:
```
pub fn main(data: bool[8], shift: u3) -> bool[8] {
    for (i, result): (u32, bool[8]) in u32:0..u32:8 {
        let value = if i < (shift as u32) {
            false
        } else {
            data[i - (shift as u32)]
        };
        update(result, i, value)
    }(bool[8]:[false, ...])
}
```
Yosys synthesis targeting ASAP7:
| metric | before | after | delta |
|---|---|---|---|
| cell area (µm²) | 13.8947 | 4.0970 | −70.5% |
| longest topological path | 10 | 4 | −60% |
| total cells | 136 | 41 | −69.9% |

### DIFF
--- a/docs_src/passes_list.md
+++ b/docs_src/passes_list.md
@@ -3345,6 +3345,25 @@ ret invoke.1
 
 
 
+## loop_idiom - Recognize loop idioms {#loop_idiom}
+
+
+Recognizes common loop idioms in `counted_for` loops and rewrites them into
+more direct IR forms BEFORE the loop is unrolled.
+
+First target: a zero-filled array shift (DSLX `data << shift` expressed as a
+`for` loop with `result[i] = if i < shift { false } else { data[i-shift] }`).
+The counted_for body lowers to an `array_update` of
+`sel(ult(i, shift), [array_index(data, [i - shift]), 0])`; rewriting that to
+a packed `shll` + per-bit `bit_slice` shape lets codegen emit a native shift
+instead of per-element index muxes.
+
+
+[Header](http://github.com/google/xls/tree/main/xls/passes/loop_idiom_pass.h)
+
+
+
+
 
 ## lut_conversion - LUT Conversion {#lut_conversion}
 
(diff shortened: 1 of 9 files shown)


allight:
While this seems like an interesting idea the transform here seems incredibly fragile, relying on the code being shaped almost exactly like this example before any other transforms are made to it (this pass is literally made to be the first thing done to the input).

For example I think the following will not be recognized by the pass

```rust
fn offset(i: u32, shift: u32) -> (bool, u32) {
  (i < shift, i - shift)
}
pub fn main(data: bool[8], shift: u3) -> bool[8] {
    for (i, result): (u32, bool[8]) in u32:0..u32:8 {
        let oob, src_idx = offset(i, shift as u32);
        let value = if oob {
            false
        } else {
            data[src_idx]
        };
        update(result, i, value)
    }(bool[8]:[false, ...])
}
```

This also seems likely to cause other optimizations to be inhibited by hiding the provenance of the array data from optimizations such as our dataflow-simp, BDD, etc.

The fundamental observation that we should recognize and emit array-shift-like operations more efficiently is good.

We already have tools to express this using the existing array-slice operation (which fills out-of-bounds slice indices with copies of the last value of the source array).

Some suggestions for improving this

1) Like all of our other analyses and optimizations this should be mainly done post-inlining. The lack of induction-variable analysis is unfortunate but being able to see this transform regardless of the surrounding context more than makes up for it.

2) The pass should probably transform the shifter code into something more like

```lisp
(array-slice (array-concat X (array (zero! [element-type X])) Y)
```

Some easy reverses let you encode shift in either direction.

This ensures that all of our other dataflow analyses can easily see through the array.

3) Maybe: Update codegen to emit ArraySlices using shifts 

Of these I'm unsure if 3 is even something we want to do TBH. Though it would be easy enough to make it do this only in cases where the last element is a trivial zero I guess.

PR #4978, opened 2026-09-15
Avoid signed remainder overflow in the JIT

ecpeterson:
This change was primarily authored by Codex. It prevents the XLS JIT from issuing LLVM signed remainder with the signed minimum and a divisor of -1, a valid XLS operation whose result is zero.

`EmitMod` already substitutes a divisor of +1 for zero, because selecting away the result of an unsafe remainder instruction does not prevent that instruction from executing. It does not cover the other LLVM exceptional case: `MIN_INT % -1`. LLVM explicitly gives that operation undefined behavior, even though its mathematical remainder is zero; x86 division instructions can trap while computing the associated overflowing quotient. See [LLVM's srem semantics](https://www.llvm.org/docs/LangRef.html#srem-instruction).

Extend the divisor substitution to signed -1. Every signed dividend has remainder zero modulo -1, so computing modulo +1 preserves XLS's result and avoids the unsafe instruction. Unsigned divisors and the existing modulo-zero policy retain their behavior. This changes JIT lowering only; it does not change synthesized RTL.

Add a shared evaluator regression covering signed minimum, maximum, -1, and zero with divisor -1 at widths 1, 2, 7, 8, 9, 16, 32, 64, 65, and 128. The shared suite exercises the interpreter and the JIT's existing optimized/unoptimized configurations.

Validation:

- A focused C++ probe compiled `EmitMod` extracted verbatim from the before/after source against locally installed LLVM 21.1.8. It generated callable signed/unsigned remainder functions and checked all 8-bit operand pairs plus boundary pairs at 16, 32, and 64 bits: 131,456 checks per executable. On x86-64 under Rosetta, the original code exits with SIGFPE and the patched code passes. Both versions pass on native Apple Silicon. This directly checks the changed emitter, but does not replace XLS's full JIT pipeline regression.
- `git diff --check` and clang-format checks for the changed code pass.
- The native Bazel command below was attempted with dependency downloads disabled. It stopped before compilation because current main requires an LLVM archive absent from the preserved build cache. The added C++ evaluator regression has therefore **not yet been built or run**.

```sh
bazel test --dynamic_mode=off --jobs=2 --test_output=errors \
  --test_arg='--gtest_filter=*InterpretSMod*:*InterpretUMod*' \
  //xls/jit:function_jit_test //xls/interpreter:ir_interpreter_test
```

The motivating erl_hls regression crashed in its pinned Linux XLS JIT while comparing an 8-bit division/remainder corpus. erl_hls currently substitutes +1 for a signed -1 remainder divisor at DSLX level; that workaround should remain until its pinned XLS release includes the fix.


dplassgit:
@ericastor would this be better in an opt pass?

ericastor:
Why would this be better in an opt pass? Our division handling already supports this, it's well-defined behavior; it's just an issue in the JIT implementatioon.

ericastor (approved):
Nice fix, thank you!

xls/interpreter/ir_evaluator_test_base.cc
```diff
@@ -1301,6 +1301,25 @@ TEST_P(IrEvaluatorTestBase, InterpretSMod) {
...
+    SCOPED_TRACE(width);
+    Package package("signed_remainder");
+    FunctionBuilder fb("remainder", &package);
+    BValue lhs = fb.Param("lhs", package.GetBitsType(width));
+    BValue rhs = fb.Param("rhs", package.GetBitsType(width));
+    fb.SMod(lhs, rhs);
+    XLS_ASSERT_OK_AND_ASSIGN(Function * function, fb.Build());
+    for (const Bits& dividend : {Bits::MinSigned(width), Bits::MaxSigned(width),
+                                 Bits::AllOnes(width), Bits(width)}) {
+      SCOPED_TRACE(dividend.ToString());
```
ericastor: Shouldn't this be ToDebugString? There is no ToString function, so this is currently failing to compile.

ericastor (changes requested):

xls/interpreter/ir_evaluator_test_base.cc
ecpeterson: Oops, correct, yes, it should be `ToDebugString`. Corrected and pushed, CI seems healthier.

ericastor (approved):

PR #4993, opened 2026-09-16
Don't build a shared library for embedded data

### LINKED ISSUE #4992: `xls_cc_embed_data`: shared-library link fails on macOS — ELF-only `-soname` passed to the Mach-O linker
## Summary

`xls_cc_embed_data` produces both a static and a shared library. Linking the shared variant fails on macOS, because the toolchain passes the ELF-only `-soname` flag to `ld64.lld`:

```
ERROR: xls/common/BUILD:286:18: Linking xls/common/libsubprocess_helper_embedded.dylib failed:
ld64.lld: error: unknown argument '-soname=libxls_Scommon_Slibsubprocess_Uhelper_Uembedded.dylib'
clang++: error: linker command failed with exit code 1
```

Mach-O uses `-install_name` rather than `-soname`.

## Impact

Any **test** target that depends on embedded data cannot be built on macOS, because tests link their dependencies dynamically by default. Concretely, `//xls/common:subprocess_test` cannot be built at all.

Tool binaries are unaffected — they link the static object — so this is easy to miss: `bazel build` of the main tools succeeds while `bazel test` of an affected target fails.

## Reproduction

On macOS arm64:

```
bazel test -c opt //xls/common:subprocess_test
```

## Suggested fix

Embedded data is only ever linked statically into its consumers, so the shared library is an unused artifact. Passing `disallow_dynamic_library = True` to the `cc_common.create_linking_context_from_compilation_outputs()` call in `xls/build_rules/xls_cc_embed_data_rules.bzl` avoids producing it — which sidesteps the broken link on macOS and drops a dead artifact on every platform.

Verified locally on macOS arm64: `//xls/build_rules/tests/...` passes 31/31 (including `embed_data_test`), and `//xls/common:subprocess_test` then builds and passes.

If you would rather fix this in the toolchain configuration so the shared variant links correctly on Mach-O, that works too — I took the minimal route. Happy to send a PR either way.

## Related

#1434 (tracking issue: OS X aarch64 build/test status), and #4991 (the `subprocess.cc`
macOS issue) -- both are needed before
`//xls/common:subprocess_test` can be built and run on macOS.

Also #1363 (macosx nightly builds are failing), which reports the symptom; this issue is one of the concrete causes.

## Environment

- macOS 26.6.2, arm64 (Apple Silicon)
- Bazel 8.7.0, XLS at `fc572666a`

marketchedai:
Fixes #4992.

`xls_cc_embed_data` produces both a static and a shared library. The shared
variant fails to link on macOS, because the toolchain passes the ELF-only
`-soname` flag to the Mach-O linker:

```
ld64.lld: error: unknown argument '-soname=libxls_Sbuild_Urules_Stests_Slibarbitrary_Udata_Ufile.dylib'
```

This blocks any test target that depends on embedded data, because `cc_test`
links its dependencies dynamically by default. On current `main` that includes
the rule's own test — `//xls/build_rules/tests:embed_data_test` cannot be built
on macOS at all.

Embedded data is only ever linked statically into its consumers, so the shared
library is an unused artifact: nothing in the tree depends on it. Passing
`disallow_dynamic_library = True` skips producing it, which avoids the broken
link on macOS and drops a dead artifact on every platform.

If you would prefer this fixed in the toolchain configuration so the shared
variant links correctly on Mach-O, I am happy to redo it that way; I took the
minimal route.

## Testing

**Tested on macOS (arm64, macOS 26.6.2) only, with Bazel 8.7.0, rebased onto
924b9fd07. I have not built this on Linux — relying on CI for that.** The
change is platform-independent Starlark; its effect on Linux is that the `.so`
is no longer produced, and consumers link the static library as they already
did.

- `//xls/build_rules/tests/...` passes 31/31 with this change, and
  `embed_data_test` **fails to build without it** on macOS.
- `//xls/common/...` passes 37/37.
- `interpreter_main`, `ir_converter_main`, `opt_main` and `codegen_main` build,
  and a DSLX -> IR -> opt -> Verilog run produces correct output.
- `//xls/common:subprocess_helper_embedded` still builds under
  `--dynamic_mode=fully`.

Only two targets use this rule (`//xls/common:subprocess_helper_embedded` and
the test's `arbitrary_data_file`), and `rdeps` shows no consumer of the shared
variant.

🤖 Generation assisted with [Claude Code](https://claude.com/claude-code)


marketchedai:
Rebased onto `924b9fd07` and corrected the justification.

When I opened this, I cited `//xls/common:subprocess_test` as the blocked target. After #4972 merged, that is no longer accurate: macOS no longer links the embedded helper into `subprocess_test`, so that test now builds and passes on `main` without this change.

The rule itself is still broken on macOS, though — the failing target is now the rule's own test, `//xls/build_rules/tests:embed_data_test`, which fails to build on unpatched `main` with the `-soname` error above. I have updated the description and commit message accordingly.

allight:
The Google cla must be agreed to before we can review or merge this pr.

allight:
@marketchedai The Google cla **must** be agreed to before we can review or merge this pr.

allight:
Again you must agree to the CLA for PRs to be considered: https://github.com/google/xls/pull/4993/checks?check_run_id=107370647803

PR #4994, opened 2026-09-16
Fix subprocess helper exec on macOS

### LINKED ISSUE #4991: `xls/common/subprocess.cc` does not compile on macOS: uses `<linux/memfd.h>`, `memfd_create()` and `/proc/self/fd`
## Summary

`xls/common/subprocess.cc` unconditionally depends on three Linux-only facilities, so it cannot compile on macOS:

```
xls/common/subprocess.cc:18:10: fatal error: 'linux/memfd.h' file not found
   18 | #include <linux/memfd.h>
      |          ^~~~~~~~~~~~~~~
```

Beyond the header, the same code path also uses:

- `memfd_create()` / `MFD_CLOEXEC` — a Linux syscall with no macOS equivalent
- the `/proc/self/fd/<fd>` path handed to `posix_spawnp()` — macOS has no `/proc` filesystem

So this is not just a missing include guard; the mechanism for exec'ing the embedded `subprocess_helper` out of an anonymous in-memory file descriptor does not exist on macOS.

The file still carries an `#if defined(__APPLE__)` block for `environ`, which suggests macOS support was intended. The Linux-only path came in with 2c355ce53 ("Embed subprocess_helper in subprocess.cc", 2026-04-30).

## Impact

Blocks `//xls/tools:codegen_main` on macOS, plus anything else depending on `//xls/common:subprocess`. Notably it does **not** block `interpreter_main`, `ir_converter_main` or `opt_main` — none of those depend on `subprocess` — but since Bazel aborts on the first error, a plain `bazel build` of all four appears to fail entirely.

This is one of the reasons `Nightly MacOS` is red (see also the llvm-project checksum issue).

## Reproduction

On macOS arm64, after working around the llvm-project checksum problem:

```
bazel build -c opt -- //xls/tools:codegen_main
```

## Suggested fix

Introduce a `GetSubprocessHelperPath()` that encapsulates producing an exec'able path for the embedded helper:

- **Linux**: unchanged — write the helper to a memfd, return `/proc/self/fd/<fd>`, hold the descriptor open for the lifetime of the process.
- **Apple**: materialize the helper into a temporary file with `mkstemp()` and mark it executable with `fchmod(0700)`, unlinking it via `std::atexit()` (and on the error paths).

The temporary file is unavoidable on macOS: unlike a Linux memfd it has a name and is not reclaimed automatically, it cannot be unlinked up front because exec resolves it by path, exec'ing an unlinked file through `/dev/fd/<fd>` is rejected with `EACCES`, and there is no `fexecve()`. A process killed by a signal will still leave one behind.

I have this working locally; `//xls/common:subprocess_test` (12 tests, covering non-zero exits, crashing children, large stdout/stderr, and environment variable propagation) passes on macOS arm64. Happy to send it as a PR.

Note that running that test on macOS also requires the `xls_cc_embed_data` shared-library fix filed separately.

## Related

#1434 (tracking issue: OS X aarch64 build/test status).

One note in that issue appears to be out of date: it states that the hermetic
LLVM toolchain does not work on OS X. I did not need to disable it -- building
with the hermetic toolchain on macOS arm64 worked, once the problem above and
the `xls_cc_embed_data` shared-library problem were addressed.

Also #1363 (macosx nightly builds are failing), which reports the symptom; this issue is one of the concrete causes.

## Environment

- macOS 26.6.2, arm64 (Apple Silicon)
- Bazel 8.7.0, XLS at `fc572666a`

marketchedai:
Fixes #4991.

`xls/common/subprocess.cc` cannot compile on macOS: it unconditionally uses
`<linux/memfd.h>`, `memfd_create()`/`MFD_CLOEXEC`, and the `/proc/self/fd/<fd>`
path passed to `posix_spawnp()`. macOS has none of these.

This introduces `GetSubprocessHelperPath()`, which encapsulates producing an
exec'able path for the embedded helper:

- **Linux — behavior unchanged.** Still writes the helper to a memfd and
  returns `/proc/self/fd/<fd>`, holding the descriptor open for the lifetime
  of the process.
- **Apple.** Materializes the helper into a temporary file via `mkstemp()`,
  marks it executable with `fchmod(0700)`, and unlinks it through
  `std::atexit()` (the error paths unlink it too).

A named file is unavoidable on macOS. Unlike a memfd it is not reclaimed
automatically, and it cannot be unlinked up front because exec resolves it by
path. I checked the alternatives rather than assuming: exec'ing an unlinked
file through `/dev/fd/<fd>` is rejected with `EACCES`, and there is no
`fexecve()`. A process killed by a signal will still leave one behind.

The call site in `ExecInChildProcess()` is deliberately not platform-guarded —
the whole point is that the platform difference lives in one function.
`subprocess_helper` is now a reference to a function-local `static const`,
which has thread-safe initialization, is never mutated, and outlives the
`c_str()` pointers handed to `posix_spawnp()`.

## Testing

**Tested on macOS (arm64, macOS 26.6.2) only. The Linux path in this change
has not been compiled anywhere — I do not have a Linux machine available, and
I could not cross-check locally: undefining `__APPLE__` to force the other
branch breaks macOS's own libc++ headers, which key off that macro. I am
relying on CI for Linux coverage, and would appreciate a careful look at the
`#else` branch.**

On macOS arm64, with Bazel 8.7.0:

- `//xls/common:subprocess_test` passes (12 tests, covering non-zero exits,
  crashing children, large stdout/stderr, and environment variable
  propagation). Note this requires #4992 to be fixed first, otherwise the test
  cannot be built on macOS at all.
- `//xls/common/...` passes 36/36.
- `interpreter_main`, `ir_converter_main`, `opt_main` and `codegen_main` all
  build, and a DSLX -> IR -> opt -> Verilog run produces correct output.
- Confirmed no temporary files accumulate: repeated runs leave none behind.

🤖 Generation assisted with [Claude Code](https://claude.com/claude-code)


allight:
We are already reviewing https://github.com/google/xls/pull/4972 for this.

marketchedai:
Thanks @allight — agreed, #4972 supersedes this, and it is the better fix: a
proper per-OS split selected by Bazel rather than more `#ifdef`s inside one
file. Closing in favour of it.

For anyone finding this later: #4972 merged as 225703468 and removes the
`memfd_create()` / `/proc/self/fd` dependency from `subprocess.cc` entirely.

Note that #4993 is a separate problem and is still outstanding on current
`main`: `xls_cc_embed_data` still builds a shared library, which fails to link
on macOS because the toolchain passes the ELF-only `-soname` to the Mach-O
linker. That still blocks building `//xls/common:subprocess_test` on macOS.

PR #5007, opened 2026-09-18
Allow configuring solver threads for QuickCheck proofs

### LINKED ISSUE #5006: [enhancement] Allow configuring Z3 threads for QuickCheck proofs
## Summary

`prove_quickcheck_main` currently runs Z3 proofs with a single solver thread.

XLS already has support for configuring the number of Z3 threads in `CreateSolver(ctx, num_threads)`, where both `threads` and `smt.threads` are set. However, the proof path currently hardcodes:

```cpp
CreateSolver(ctx, /*num_threads=*/1);
```

It would be useful to expose this setting to `prove_quickcheck_main`.

## Motivation

Some QuickCheck properties produce sufficiently large SMT problems that Z3 can benefit from using multiple solver threads.

This is especially useful when `prove_quickcheck_main` is used for exhaustive formal proof rather than random QuickCheck execution.

The proposed option would affect only Z3's internal parallelism. It would not run multiple QuickCheck properties or test cases concurrently.

## Proposed interface

Add an option such as:

```text
--z3_num_threads=N
```

with the current behavior preserved as the default:

```text
--z3_num_threads=1
```

For example:

```text
prove_quickcheck_main --z3_num_threads=8 --test_filter=my_property module.x
```

Invalid values such as zero or negative thread counts should be rejected.

## Implementation

The setting can be propagated through `ParseAndProveOptions` to the Z3 proof path and then used when creating the solver.

Since `CreateSolver()` already accepts `num_threads`, no new Z3 configuration mechanism is required.

I have a patch implementing this, including tests for the default/single-threaded behavior, multi-threaded configuration, invalid thread counts, successful proofs, and counterexamples.

The current workaround is to modify the Z3 proof path locally and replace the hardcoded `CreateSolver(ctx, /*num_threads=*/1)` with the desired number of threads. This requires maintaining a local patch to XLS and rebuilding the relevant binaries. There is currently no command-line option or `ParseAndProve` configuration that allows users of `prove_quickcheck_main` to control Z3 solver parallelism without changing the source code.

Expose the existing Z3 thread configuration through the QuickCheck proof interface, preferably as an option such as `--z3_num_threads=N` in `prove_quickcheck_main`, with `1` as the default to preserve current behavior. The option should be propagated through `ParseAndProveOptions` to the Z3 solver creation path, where XLS already supports configuring both `threads` and `smt.threads`. Invalid values such as zero or negative thread counts should be rejected. No changes to QuickCheck semantics or parallel execution of test cases are required.

tilir:
This change allows the number of Z3 solver threads used by QuickCheck proofs to be configured via:

```text
--z3_num_threads=N
```

The default remains `1`, preserving the existing behavior.

The option is propagated through `ParseAndProveOptions` to the Z3 proof path and ultimately to the existing `CreateSolver(ctx, num_threads)` API. Zero and negative values are rejected.

This controls Z3's internal parallelism only; it does not run QuickCheck properties or test cases concurrently.

The change also adds tests covering single- and multi-threaded configurations, invalid thread counts, successful proofs, and counterexamples, and updates the `prove_quickcheck_main` documentation accordingly.

Fixes #5006


dplassgit (approved):

ericastor:
Before we move forward with this, I'd like to suggest that we remove "Z3" from the flag name; we might want to add support for other solvers soon.

tilir:
> Before we move forward with this, I'd like to suggest that we remove "Z3" from the flag name; we might want to add support for other solvers soon.

Makes sense. I'll rename it to `--solver_num_threads` and keep the Z3-specific mapping inside the Z3 backend.

My only concern with making the option solver-independent is that not every solver necessarily supports the same threading model. I think the clean semantics would be "maximum number of threads available to an individual solver invocation"; backends that support it can map it to their corresponding option, while unsupported backends should report an error rather than silently ignore a value greater than 1. 

Would you be OK with this approach?

ericastor:
> > Before we move forward with this, I'd like to suggest that we remove "Z3" from the flag name; we might want to add support for other solvers soon.
> 
> Makes sense. I'll rename it to `--solver_num_threads` and keep the Z3-specific mapping inside the Z3 backend.
> 
> My only concern with making the option solver-independent is that not every solver necessarily supports the same threading model. I think the clean semantics would be "maximum number of threads available to an individual solver invocation"; backends that support it can map it to their corresponding option, while unsupported backends should report an error rather than silently ignore a value greater than 1.
> 
> Would you be OK with this approach?

Sounds great to me!

tilir:
> > > Before we move forward with this, I'd like to suggest that we remove "Z3" from the flag name; we might want to add support for other solvers soon.
> > 
> > 
> > Makes sense. I'll rename it to `--solver_num_threads` and keep the Z3-specific mapping inside the Z3 backend.
> > My only concern with making the option solver-independent is that not every solver necessarily supports the same threading model. I think the clean semantics would be "maximum number of threads available to an individual solver invocation"; backends that support it can map it to their corresponding option, while unsupported backends should report an error rather than silently ignore a value greater than 1.
> > Would you be OK with this approach?
> 
> Sounds great to me!

Done, please take a look.

xls/dslx/run_routines/run_routines_test.cc
```diff
@@ -865,6 +865,7 @@ fn quickcheck_that_fails(x: u1) -> bool {
 }
 )";
   ParseAndProveOptions options;
+  options.solver_num_threads = 2;
```
ericastor: This seems like an unexpected change?

tilir: Sure, leftover from some local experiments. Thanks for catching that. Fixed.

ericastor:
... darn. We had a broken build for the weekend, and this was apparently rebased onto it. With my apologies for the inconvenience, can you please rebase to head?

tilir:
> ... darn. We had a broken build for the weekend, and this was apparently rebased onto it. With my apologies for the inconvenience, can you please rebase to head?

Done

ericastor (approved):
Thanks! This is a great change.

PR #5028, opened 2026-09-21
Fix rules_bison and rules_cc_autoconf build failures and aot-compiler linker error with newer LLVM

allight:
With the upgraded LLVM/libc++ toolchain, stricter transitive header inclusion causes C++ compilation failures in `@rules_bison` and `@rules_cc_autoconf`:
- Add `//dependency_support/rules_bison:include_cstdlib.patch` via `single_version_override` so `src/bazel_runfiles.cc` explicitly includes `<cstdlib>` for `getenv()`.
- Add `bazel_dep(name = "rules_cc_autoconf", version = "0.24.0")` to pick up the upstream `<iterator>` header include in `debug_logger.h` for `std::back_inserter`.

In newer LLVM versions, `llvm::SplitModule` calls `GlobalValue::externalize()`, which invokes `GlobalValue::nameUnnamed()` and names unnamed globals `__llvm_unnamed` (e.g. `__llvm_unnamed`, `__llvm_unnamed.1`) instead of `__llvmsplit_unnamed`.

Update `aot_compiler_generate_segments_main` to salt symbols containing `__llvm_unnamed` (in addition to `__llvmsplit_unnamed`) across `module->global_values()`, preventing duplicate symbol errors at link time when multiple sharded AOT targets (`jobs > 1`) are linked into the same binary.

Generated with Google Antigravity


allight:
@richmckeever

ericastor (approved):

